PR #15, opened 2022-01-24
improve nydus snapshotter container image

bergwolf:
A few improvements for the nydus-snapshotter image.

1. support customized containerd root path
2. use shared daemon mode
3. handle base image CA issue
4. do not install recommended packages
5. clean up apt cache

Signed-off-by: Peng Tao <bergwolf@hyper.sh>


eryugey (approved):

changweige (approved):
Thanks

PR #17, opened 2022-01-26
fix image builder ssl issue and trigger build on pull_request and release

bergwolf:
* `nydus-snapshotter:latest` is built and pushed on each push to main
* `nydus-snapshotter:$release_tag` is built and pushed on each release tag
* `nydus-snapshotter:local` is built on each pull request but not pushed, for testing purpose

Tested at https://github.com/bergwolf/nydus-snapshotter/actions/workflows/image.yml

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,9 +1,14 @@
 name: nydus-snapshotter image
 
 on:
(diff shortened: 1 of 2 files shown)


changweige (approved):
LGTM

Issue #8, opened 2022-01-18
nydus-snapshotter should look for nydusd/nydus-image in $PATH


bergwolf:
Instead of hardcoding it or just relying on a fixed parameter.

liubin:
Can I take this issue if it is not assigned yet?

changweige:
Sure. Welcome to contribute.

PR #18, opened 2022-01-26
cmd: remove nydusd-path and nydusimg-path startup options

liubin:
Remove nydusd-path and nydusimg-path from startup options,
instead lookup them in PATH.

Fixes: #8


eryugey (approved):

changweige:
Could we just make the two options as optional?
In particular, if they are not provided, a trial to find them from $PATH is performed.

liubin:
> Could we just make the two options as optional? In particular, if they are not provided, a trial to find them from $PATH is performed.

I changed the default value to an empty string and will look up in PATH only when they are not specified.

changweige (approved):
LGTM. Thanks

Issue #19, opened 2022-01-27
thread-num for nydusd should be configurable?


liubin:
Now it's set to a fixed value of `10`:

https://github.com/containerd/nydus-snapshotter/blob/672434e7b31924f286932de4ea1941f66fb03b59/pkg/process/manager.go#L133-L138

changweige:
Very much agree

PR #20, opened 2022-01-27
make thread-num for nydusd should be configurable

liubin:
Use a user configurable for thread-num than a fixed value.

Fixes: #19

### DIFF
--- a/cmd/containerd-nydus-grpc/pkg/command/flags.go
+++ b/cmd/containerd-nydus-grpc/pkg/command/flags.go
@@ -49,6 +49,7 @@ type Args struct {
 	DisableCacheManager  bool
 	LogToStdout          bool
 	EnableNydusOverlayFS bool
+	NydusdThreadNum      int
 }
 
 type Flags struct {
@@ -182,6 +183,11 @@ func buildFlags(args *Args) []cli.Flag {
 			Usage:       "whether to disable nydus-overlayfs to mount",
 			Destination: &args.EnableNydusOverlayFS,
 		},
+		&cli.IntFlag{
+			Name:        "nydusd-thread-num",
+			Usage:       "Nydusd daemon thread-num, default will be set to the number of CPUs",
+			Destination: &args.NydusdThreadNum,
+		},
 	}
 }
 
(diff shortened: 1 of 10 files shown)


snapshot/snapshot.go
```diff
@@ -113,6 +113,8 @@ func NewSnapshotter(ctx context.Context, cfg *config.Config) (snapshots.Snapshot
 		nydus.WithLogLevel(cfg.LogLevel),
 		nydus.WithLogDir(cfg.LogDir),
 		nydus.WithLogToStdout(cfg.LogToStdout),
+		nydus.WithLogToStdout(cfg.LogToStdout),
```
changweige, line 116: Duplicated code line?

cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -182,6 +183,11 @@ func buildFlags(args *Args) []cli.Flag {
 			Usage:       "whether to disable nydus-overlayfs to mount",
 			Destination: &args.EnableNydusOverlayFS,
 		},
+		&cli.IntFlag{
+			Name:        "nydusd-thread-num",
+			Usage:       "Nydusd daemon thread-num, default will be set to the number of CPUs",
```
changweige, line 188: I am not very sure if make the default threads number to the number of CPUs is a suitable policy.
Consider that a server equipped with many CPU sockets and cores, dozens of threads will be spawned. But most of them are idle.

liubin: It seems nydusd is using a default of `1`, so if we don't want to set a default value from snapsohtter, we can just don't pass this argument to nydusd if the user doesn't set it.

changweige: Sounds reasonable

liubin: fixed.

snapshot/snapshot.go
liubin: Deleted, my keyboard is broken...

changweige (approved):
Thanks

PR #22, opened 2022-02-09
README.md: add disable_snapshot_annotations=false in cri env

luodw:
As 'containerd config default' produce default configuration
which include disable_snapshot_annotations=true, but nydus-snapshotter
depend on snapshot annotation, so here we should set explicitly.

Signed-off-by: luodaowen.backend <luodaowen.backend@bytedance.com>


changweige (approved):
LGTM

PR #24, opened 2022-02-11
Provide make install to facilitate configuring nydus-snapshotter 

changweige:
Users just perform `make install` and download `nydusd` and `nydus-image` before they can use nydus-snapshotter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,14 @@ clear:
 	rm -f bin/*
 	rm -rf _out
 
+
+.PHONY: install
+install: static-release
+	sudo install -D -m 755 bin/containerd-nydus-grpc /usr/local/bin/containerd-nydus-grpc
+	sudo install -D -m 755 misc/snapshotter/nydusd-config.json /etc/nydus/config.json
+	sudo install -D -m 644 misc/snapshotter/nydus-snapshotter.service /etc/systemd/system/nydus-snapshotter.service
+	sudo systemctl enable /etc/systemd/system/nydus-snapshotter.service
+
 .PHONY: vet
 vet:
 	go vet $(PACKAGES)
--- a/README.md
+++ b/README.md
@@ -49,8 +49,14 @@ Please follow instructions to [configure nydus](./docs/configure_nydus.md) confi
 
 Nydus-snapshotter is implemented as a [proxy plugin](https://github.com/containerd/containerd/blob/04985039cede6aafbb7dfb3206c9c4d04e2f924d/PLUGINS.md#proxy-plugins) (`containerd-nydus-grpc`) for containerd.
 
-A example of starting nydus-snapshotter:
+Assume your server systemd based, install nydus-snapshotter:
+Note: `nydusd` and `nydus-image` should be found from $PATH.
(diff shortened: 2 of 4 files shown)


eryugey (commented):
Add a "prefix: " to commit summary? e.g.

cmd: shorten default socket path

eryugey (commented):
Same to this patch, add "build" or "Makefile" prefix?

build: add install target to Makefile

eryugey (commented):
And we could use "misc: " as commit summary prefix in this patch

Makefile
```diff
@@ -21,6 +21,13 @@ clear:
 	rm -f bin/*
 	rm -rf _out
 
+.PHONY: install
+install: static-release
+	sudo install -D -m 755 bin/containerd-nydus-grpc /usr/local/bin/containerd-nydus-grpc
+	sudo mkdir -p /etc/nydus
+	sudo cp misc/snapshotter/nydusd-config.json /etc/nydus/config.json
+	sudo systemctl enable ${PWD}/misc/snapshotter/nydus-snapshotter.service
```
eryugey, line 29: Do we want to install this systemd service file to system? e.g. /lib/systemd/system/nydus-snapshotter.service ?

README.md
```diff
@@ -49,8 +49,13 @@ Please follow instructions to [configure nydus](./docs/configure_nydus.md) confi
 
 Nydus-snapshotter is implemented as a [proxy plugin](https://github.com/containerd/containerd/blob/04985039cede6aafbb7dfb3206c9c4d04e2f924d/PLUGINS.md#proxy-plugins) (`containerd-nydus-grpc`) for containerd.
 
-A example of starting nydus-snapshotter:
+Assume your server systemd based, install nydus-snapshotter:
+```bash
+make install
+systemctl restart containerd
```
imeoer, line 55: Should mention ensuring nydusd/nydus-image have been installed to $PATH?

changweige:
> And we could use "misc: " as commit summary prefix in this patch

Sure. All commits are prefixed now

README.md
changweige: Yes. we can note it here.

Makefile
changweige: Sorry for previously misunderstanding. It is now installed.

Makefile
```diff
@@ -21,6 +21,14 @@ clear:
 	rm -f bin/*
 	rm -rf _out
 
+
+.PHONY: install
+install: static-release
+	sudo install -D -m 755 bin/containerd-nydus-grpc /usr/local/bin/containerd-nydus-grpc
+	sudo install -D -m 755 misc/snapshotter/nydusd-config.json /etc/nydus/config.json
+	sudo install -D -m 755 misc/snapshotter/nydus-snapshotter.service /etc/systemd/system/nydus-snapshotter.service
```
eryugey, line 29: I don't think we need 755 perm for nydus-snapshotter.service file. 644 should be fine?

changweige: Good point!

eryugey (approved):

Issue #26, opened 2022-02-23
nydus sdk lack of error context


eryugey:
Most errors returned in pkg/nydussdk/client.go don't have a context, so it's hard to determine where the error is returned.

PR #27, opened 2022-02-23
pkg/nydussdk: add more context on returning errors

eryugey:
So that we know where does that error come from, and make debug easier.

Fixes: #26
Signed-off-by: Eryu Guan <eguan@linux.alibaba.com>


imeoer (commented):
LGTM, should also add an error wrap to the `waitUntilSocketReady` call?

PR #28, opened 2022-03-04
mod: update containerd, docker and distribution dependency

bergwolf:
To bring in security fixes.


bergwolf:
```
verifying github.com/docker/distribution@v2.8.0+incompatible: checksum mismatch
	downloaded: h1:u9vuu6qqG7nN9a735Noed0ahoUm30iipVRlhgh72N0M=
	go.sum:     h1:l9EaZDICImO1ngI+uTifW+ZYvvz7fKISBAKpg+MbWbY=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
```
We'd have to wait for the replace `docker/distribution` version to pop up in various go mod proxy servers.

imeoer (approved):

PR #29, opened 2022-03-04
action: add simple e2e ci

bergwolf:


bergwolf:
https://github.com/bergwolf/nydus-snapshotter/runs/5418544506?check_suite_focus=true

misc/example/10-containerd-net.conflist
```diff
@@ -0,0 +1,24 @@
+{
+  "cniVersion": "0.3.1",
```
mikebrow, line 2: 1.0.0

bergwolf: Good point, thanks!

misc/example/README.md
```diff
@@ -0,0 +1,52 @@
+## Example to setup and test nydus-snapshoter
```
imeoer, line 1: Linking this doc on the main page should be better.

imeoer (approved):
LGTM!

Issue #31, opened 2022-03-22
failed to mount daemon AsQl4TceS5aaKIsu33yd6g: failed to shared mount: http response: 503, error code: , error message:


eryugey:
Sometimes we saw mount failure as

failed to mount daemon AsQl4TceS5aaKIsu33yd6g: failed to shared mount: http response: 503, error code: , error message:

No error code no error message, just 503 status code.

PR #32, opened 2022-03-22
daemon: use one nydus client per Daemon

eryugey:
Currently we create new nydusd client in CheckStatus(), SharedMount()
and SharedUmount(), but set max connection in underlying transport to
10. So we sometimes see 503 response when the number of connections
exceed 10.

So let's use one nydus client per Daemon to avoid such 503 error.

Fixes: #31
Signed-off-by: Eryu Guan <eguan@linux.alibaba.com>

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"sync"
 
 	"github.com/containerd/nydus-snapshotter/config"
 	"github.com/containerd/nydus-snapshotter/pkg/nydussdk"
@@ -41,6 +42,10 @@ type Daemon struct {
 	RootMountPoint   *string
 	CustomMountPoint *string
 	nydusdThreadNum  int
+
+	// Client will be rebuilt on Reconnect, skip marshal/unmarshal
+	Client nydussdk.Interface `json:"-"`
+	Once   *sync.Once         `json:"-"`
 }
 
 func (d *Daemon) SharedMountPoint() string {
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -92,31 +93,28 @@ func (d *Daemon) LogFile() string {
 }
 
 func (d *Daemon) CheckStatus() (model.DaemonInfo, error) {
-	client, err := nydussdk.NewNydusClient(d.APISock())
-	if err != nil {
-		return model.DaemonInfo{}, errors.Wrap(err, "failed to check status, client has not been initialized")
+	if d.Client == nil {
+		return model.DaemonInfo{}, errors.New("failed to check status, client has not been initialized")
```
imeoer, line 97: Use a func like `err := d.ensureClient(action string)` to simplify this code?

pkg/daemon/daemon.go
```diff
@@ -41,6 +41,7 @@ type Daemon struct {
 	RootMountPoint   *string
 	CustomMountPoint *string
 	nydusdThreadNum  int
+	Client           nydussdk.Interface
```
imeoer, line 44: Should this be serialized into the meta db?

eryugey: You're right, Client can't be serialized. Skip it in new version. Thanks!

pkg/daemon/daemon.go
```diff
@@ -131,10 +140,36 @@ func (d *Daemon) IsPrefetchDaemon() bool {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create new nydus client")
+	}
+	d.Client = client
+	return nil
+}
+
+func (d *Daemon) ensureClient(action string) error {
+	var err error
+	if d.Client == nil {
```
imeoer, line 154: This check should be put in `sync.Once.Do`, maybe some races occur in concurrency?

eryugey: Done!

imeoer (approved):
LGTM!

eryugey:
> 

Thanks for all the reviews!

PR #34, opened 2022-03-25
Add converter package for tar build

imeoer:
Introduce a new nydus image build mode named tar build.

In order to correspond to the OCI image layer, we need a new build mode, that is
to convert the OCI-formatted tar stream directly into a nydus-formatted tar stream.

The nydus tar stream contains blob and bootstrap files, and the workflow is divided
into two steps, Convert and Merge.

Convert: converts each layer of the OCI tar stream to a nydus tar stream.
Merge: reads the bootstrap files from each layer of the nydus tar stream and merges
them into a final bootstrap.

This package will be imported by acceleration-service and buildkit to support the
conversion or build of nydus image.

Depends on [merge](https://github.com/dragonflyoss/image-service/pull/351) sub-command in builder(nydus-image).


imeoer:
Waiting for https://github.com/dragonflyoss/image-service/pull/351 to be merged, then we can continue to resolve CI issue.

pkg/converter/tool/builder.go
```diff
@@ -0,0 +1,118 @@
...
+		"directory",
+		"--whiteout-spec",
+		"none",
+		"--fs-version",
+		option.RafsVersion,
+		"--blob-offset",
+		// Add blob offset for chunk info with size_of(tar_header) * 2.
+		"1024",
+	}
+	if option.RafsVersion != "" {
```
bergwolf, line 60: args cannot work if `RafsVersion` is empty. Do you mean to disable check for rafs v6? Also IMO it is better to expose the checker option to users/callers instead of hardcoding it here.

pkg/converter/tool/nydusd.go
```diff
@@ -0,0 +1,216 @@
+/*
+ * Copyright (c) 2022. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
bergwolf, line 6: Should this live in `pkg/nydussdk` instead?

pkg/converter/tool/nydusd.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	},
+	"mode": "cached",
+	"iostats_files": false,
+	"fs_prefetch": {
+		"enable": {{.EnablePrefetch}},
+		"threads_count": 10,
+		"merging_size": 131072
+	},
+	"digest_validate": true,
```
bergwolf, line 67: If it is to be used during build process, I'd suggest that we disable digest validation since it affects performance a lot. For tests, we can just enable it for sure. Or are you planning to add an option to set it later on?

bergwolf (commented):
Generally lgtm. A few comments inline.

pkg/converter/tool/builder.go
imeoer: Fixed, added `RafsVersion` option to `ConvertOption`.

pkg/converter/tool/nydusd.go
imeoer: Only use it in converter tests, moved to `tests/nydusd.go`.

bergwolf (approved):

PR #41, opened 2022-04-19
1. fix the bug that daemon status is not fully serialized; 2. add cleanup_on_close config option; 3. add StatusCode in handleMountError

power-more:
1.  When writing daemon status to nydus.db, json serialization will only
    recognize member variables starting with uppercase letters, and
    lowercase ones will be ignored. We change these previously ignored ones
    to uppercase.

2. Previously, the Close() interface of nydus snapshotter performed Cleanup()
    on the filesystem. This behavior is not universal. In fact, on the
    contrary, in most cases, the default behavior of Close() should not affect
    the existing filesystem so that the filesystem can still work normally
    after Reconnect(). While in some special scenarios, the filesystem needs to
    be cleaned up with the close of the snapshotter.
    
    So we add introduce a new config option called 'cleanup_on_close' to
    control whether to clean up the filesystem under the nydus snapshotter when
    the snapshotter is closed, and clean up the nydusd daemon and snapshot at
    the same time.

3. Help to know what StatusCode is returned from nydusd.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -40,7 +40,7 @@ type Daemon struct {
 	Pid              int
 	ImageID          string
 	DaemonMode       string
-	apiSock          *string
+	APISock          *string
 	RootMountPoint   *string
 	CustomMountPoint *string
 	nydusdThreadNum  int
@@ -87,9 +87,9 @@ func (d *Daemon) NydusdThreadNum() string {
 	return ""
 }
 
-func (d *Daemon) APISock() string {
-	if d.apiSock != nil {
-		return *d.apiSock
+func (d *Daemon) GetAPISock() string {
+	if d.APISock != nil {
+		return *d.APISock
 	}
 	return filepath.Join(d.SocketDir, APISocketFileName)
 }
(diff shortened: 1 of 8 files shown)


eryugey:
Why all commits summary starts with "action: "?

pkg/daemon/daemon.go
```diff
@@ -40,7 +40,7 @@ type Daemon struct {
 	Pid              int
 	ImageID          string
 	DaemonMode       string
-	apiSock          *string
+	APISock          *string
```
imeoer, line 43: We'd better add a hint for the field name in JSON format to avoid the case-sensitivity issue like this:

```
APISock *string `json:"APISock"`
```

power-more: The json tag of the member variable is used to ensure persistence compatibility. When the variable name is modified, it can still maintain the consistency of the nydus db storage field name (of course, the first letter of the variable name should still be capitalized), so I think it is not just APISock need to add json tag, all fields can be considered to be added, so I will submit a separate patch to do this in the future.

imeoer (approved):

PR #42, opened 2022-04-20
process: fix shared umount on daemon destroy

imeoer:
For shared daemon mode, we should first umount for virtual daemon,
and then cleanup daemon resource (config, log) and the daemon info
in store, otherwise we can't umount once the config be removed,
because in the erofs scenario, we need to rely on the domain id
recorded in the config to request the daemon to unbind the blob
resources.


eryugey (approved):
LGTM

PR #43, opened 2022-04-21
converter: fix incomplete unpack

imeoer:
The current `Convert` implementation has the following interface:

```
Convert(ctx context.Context, dest io.Writer, opt ConvertOption) (src io.WriteCloser, err error)
```

The caller writes the OCI tar stream to src and Convert converts
src to a nydus tar stream and writes it to dest.

The current implementation must wait for the OCI tar stream to be
unpacked into a directory before calling the builder (nydus-image)
to start the build, which takes advantage of `pr, pw := io.Pipe()`.

Before this commit, the `unpackOciTar` method did not fully read
the OCI tarball when `pw.Close()` was called, which means that the
build is started before the unpacking is complete, which may result
in some files being missing from the nydus image.

Fixup: use a channel to wait for the unpack to complete.

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -221,17 +221,24 @@ func Convert(ctx context.Context, dest io.Writer, opt ConvertOption) (io.WriteCl
 
 	pr, pw := io.Pipe()
 
+	unpackDone := make(chan bool, 1)
 	go func() {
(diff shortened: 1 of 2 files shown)


eryugey (approved):
LGTM!

PR #44, opened 2022-04-29
converter: improve pack performance by fifo

imeoer:
In the previous convert implementation, the builder (nydus-image) would
dump the blob and bootstrap, and then the converter would pack the two
files into a tar format, which has a performance loss, this patch solves this
problem by introducing fifo file.

The converter creates a fifo file, the builder will write the blob and
bootstrap data to the fifo as the writer side, then the converter read
from the fifo and write directly to content store for faster conversion.

Related PR: https://github.com/dragonflyoss/image-service/pull/415

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -14,9 +14,12 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	"github.com/containerd/containerd/archive"
 	"github.com/containerd/containerd/archive/compression"
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/fifo"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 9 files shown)


imeoer:
Wait for this PR https://github.com/dragonflyoss/image-service/pull/415 to be merged then pass the CI testing.

eryugey (approved):

pkg/converter/convert.go
```diff
@@ -240,23 +200,30 @@ func Convert(ctx context.Context, dest io.Writer, opt ConvertOption) (io.WriteCl
...
-			SourcePath:       sourceDir,
-			ChunkDictPath:    opt.ChunkDictPath,
-			PrefetchPatterns: opt.PrefetchPatterns,
-		}); err != nil {
-			return errors.Wrapf(err, "build source %s", sourceDir)
+		blobFifo, err := fifo.OpenFifo(ctx, blobPath, syscall.O_CREAT|syscall.O_RDONLY|syscall.O_NONBLOCK, 0644)
+		if err != nil {
+			return errors.Wrapf(err, "create fifo file")
 		}
+		defer blobFifo.Close()
```
bergwolf, line 208: blobFifo is closed upon convert error.

bergwolf (commented):
Otherwise lgtm!

PR #45, opened 2022-05-09
Remove duplicated chunk info from bootstrap v6 format

zyfjeff:
[Ref: avoid the extra duplicated chunk info table in bootstrap](https://github.com/dragonflyoss/image-service/issues/414)
When the snapshotter pulls the bootstrap file in v6 format, it will analyze the superblock to find the offset of chunk info, and then discard the chunk info part, which can not only avoid downloading chunk info but also download the complete bootstrap when chunk info is needed to support the chunk dict.

**Perf data:**

| Image Size  |  Bootstrap size |  Raw Bootstrap size | Pull Time(remove chunk info)  |  Pull Time | Percent  |
|---|---|---|---|---| --- |
|   1.67G |  36M  |  28M |  1701ms  |  2050ms | 17% | 
|  79M |  1.8M  | 1.3M    |   338ms  |  384ms | 12% |
|   403M |  8.6M | 6.1M  |   424ms  |  553ms | 23% |

Signed-off-by: zyfjeff <zyfjeff@linux.alibaba.com>


pkg/filesystem/nydus/fs.go
```diff
@@ -91,12 +127,118 @@ func (fs *filesystem) newSharedDaemon() (*daemon.Daemon, error) {
...
+	_, err = finalBootstarp.Write(buf)
+
+	if err != nil {
+		return err
+	}
+
+	if isRafsV6(buf) {
+		size := getBootstrapRealSizeInV6(buf)
+		if size < MaxSuperBlockSize {
+			return fmt.Errorf("invalid chunk info offset %d", size)
```
liubogithub, line 191: I get that the start position of chunk info is the size of bootstrap, but can we put "invalid bootstrap size" for the error message instead?

liubogithub (approved):
Looks good with a few nit comments.

liubogithub:
IIUC, in order to add an extra "blob" annotation to the bootstrap layer, it also needs changes on nydusify and accelctl, is there a reference link for those changes?

pkg/filesystem/nydus/fs.go
zyfjeff: fixed

pkg/filesystem/nydus/fs.go
```diff
@@ -25,18 +32,46 @@ import (
 	"github.com/containerd/nydus-snapshotter/pkg/label"
 	"github.com/containerd/nydus-snapshotter/pkg/process"
 	"github.com/containerd/nydus-snapshotter/pkg/signature"
+	"github.com/containerd/nydus-snapshotter/pkg/utils/registry"
 )
 
 // TODO: Move snapshotter needed all image annotations to nydus-snapshotter.
 const LayerAnnotationNydusBlobIDs = "containerd.io/snapshot/nydus-blob-ids"
 
+// RafsV6 layout: 1k + SuperBlock(128) + SuperBlockExtener(256)
```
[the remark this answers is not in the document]
zyfjeff: fixed

zyfjeff:
> IIUC, in order to add an extra "blob" annotation to the bootstrap layer, it also needs changes on the side of build-tools, such as nydusify and accelctl, so is there a reference link for those changes?

I am working on it. currently testing the performance after removing chunk info

pkg/filesystem/nydus/resolver.go
```diff
@@ -0,0 +1,80 @@
...
+	resolver := Resolver{
+		transport: http.DefaultTransport,
+	}
+	return &resolver
+}
+
+func (r *Resolver) Resolve(ref, digest string, labels map[string]string) (io.ReadCloser, error) {
+	named, err := docker.ParseDockerRef(ref)
+	if err != nil {
+		return nil, err
```
imeoer, line 36: We'd better use `errors.Wrap` wrap all `err` as far as possible.

pkg/filesystem/nydus/resolver.go
```diff
@@ -0,0 +1,80 @@
...
+	host := docker.Domain(named)
+	sref := fmt.Sprintf("%s/%s", host, docker.Path(named))
+	nref, err := name.ParseReference(sref)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse ref %q (%q)", sref, digest)
+	}
+	keychain := auth.GetRegistryKeyChain(host, labels)
+
+	var tr http.RoundTripper
+	if nref.Context().Registry.Scheme() == "https" {
```
imeoer, line 47: Why do we enable auth only for `https`? the `http` registry server may also require auth.

pkg/utils/registry/registry.go
```diff
@@ -39,3 +47,40 @@ func ParseImage(imageID string) (Image, error) {
...
+		rRef = ref
+	}
+	if layerDigest, ok := labels[label.CRIDigest]; ok {
+		rDigest = layerDigest
+	}
+	return
+}
+
+func AuthnTransport(ref name.Reference, tr http.RoundTripper, keychain authn.Keychain) (http.RoundTripper, error) {
+	if keychain == nil || (reflect.ValueOf(keychain).Kind() == reflect.Ptr && reflect.ValueOf(keychain).IsNil()) {
```
imeoer, line 62: Why we need to check `nil` interface?

pkg/filesystem/nydus/resolver.go
zyfjeff: I misunderstood, my local localhost registry does not have auth, so there will be an error because I can't get the keychain, so here I only create a RoundTripper with auth for HTTPS

pkg/utils/registry/registry.go
zyfjeff: When no auth exists, a null pointer will be passed in from the outside, which will cause a non-nil interface to exist.

pkg/filesystem/nydus/fs.go
```diff
@@ -91,12 +127,118 @@ func (fs *filesystem) newSharedDaemon() (*daemon.Daemon, error) {
...
+		_, err := io.CopyN(finalBootstarp, tr, int64(size-MaxSuperBlockSize))
+		return err
+	}
+
+	// Copy remain data to bootstrap file
+	_, err = io.Copy(finalBootstarp, tr)
+	return err
+}
+
+func (fs *filesystem) PrepareLayer(ctx context.Context, s storage.Snapshot, labels map[string]string) error {
```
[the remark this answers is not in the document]
zyfjeff: fixed

pkg/filesystem/nydus/resolver.go
zyfjeff: fixed

pkg/filesystem/nydus/fs.go
```diff
@@ -91,12 +127,118 @@ func (fs *filesystem) newSharedDaemon() (*daemon.Daemon, error) {
...
+	err = os.Mkdir(filepath.Dir(workdir), 0755)
+	if err != nil {
+		return errors.Wrap(err, "failed to create bootstrap dir")
+	}
+	nydusBootstrap, err := os.OpenFile(workdir, os.O_CREATE|os.O_RDWR, 0755)
+	if err != nil {
+		return errors.Wrap(err, "failed to create bootstrap file")
+	}
+
+	legacyNydusBootstrap, err := os.OpenFile(legacy, os.O_CREATE|os.O_RDWR, 0755)
```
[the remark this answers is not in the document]
zyfjeff: Yup, Will there be a problem?

pkg/filesystem/nydus/fs.go
```diff
@@ -91,12 +127,118 @@ func (fs *filesystem) newSharedDaemon() (*daemon.Daemon, error) {
...
+	if ref == "" || layerDigest == "" {
+		return fmt.Errorf("can not find ref and digest from label %+v", labels)
+	}
+
+	readerCloser, err := fs.resolver.Resolve(ref, layerDigest, labels)
+	if err != nil {
+		return errors.Wrapf(err, "failed to resolve from ref %s, digest %s", ref, layerDigest)
+	}
+	defer func() {
+		readerCloser.Close()
```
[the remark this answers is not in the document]
zyfjeff: fixed

pkg/filesystem/nydus/fs.go
```diff
@@ -91,12 +127,118 @@ func (fs *filesystem) newSharedDaemon() (*daemon.Daemon, error) {
...
-	panic("implement me")
+func writeBootstrapToFile(reader io.Reader, bootstrap *os.File, LegacyBootstrap *os.File) error {
+	start := time.Now()
+	defer func() {
+		duration := time.Since(start)
+		log.L.Infof("read bootstrap duration %d", duration.Milliseconds())
+	}()
+	rd, err := gzip.NewReader(reader)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
zyfjeff: fixed

pkg/filesystem/nydus/fs.go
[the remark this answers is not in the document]
imeoer: No, just an empty file is left.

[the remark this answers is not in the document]
zyfjeff: If it is legacy format, an empty bootstrap file will be created here, and the empty bootstrap file will be load first, so I must delete the empty file here

pkg/filesystem/nydus/fs.go
```diff
@@ -91,12 +127,139 @@ func (fs *filesystem) newSharedDaemon() (*daemon.Daemon, error) {
...
+				file: nydusBootstrap,
+				path: workdir,
+			},
+			{
+				file: legacyNydusBootstrap,
+				path: legacy,
+			},
+		}
+		for _, in := range closeEmptyFile {
+			size, err := in.file.Seek(0, 2)
```
imeoer, line 251: We can use the `io.SeekEnd` constant.

zyfjeff: fixed

PR #46, opened 2022-05-11
makefile: ensure cross-compilation compatibility

imeoer:
We need to ensure the snapshotter can be cross-compiled across all
platforms, and although we currently only support linux. We must
make sure that external projects (such as buildkit) are not affected
by cross-compiling when they import package such as `converter`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ SUDO = $(shell which sudo)
 GO_EXECUTABLE_PATH ?= $(shell which go)
 NYDUS_BUILDER ?= /usr/bin/nydus-image
 NYDUS_NYDUSD ?= /usr/bin/nydusd-fusedev
+GOOS ?= linux
+GOARCH ?= amd64
 #GOPROXY ?= https://goproxy.io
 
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -41,8 +40,13 @@ vet:
 check: vet
 	golangci-lint run
 
+cross-build-test:
+	GOOS=linux GOARCH=amd64 go build -v -o /dev/null ./cmd/containerd-nydus-grpc
+	GOOS=windows GOARCH=amd64 go build -v -o /dev/null ./cmd/containerd-nydus-grpc
+	GOOS=darwin GOARCH=amd64 go build -v -o /dev/null ./cmd/containerd-nydus-grpc
```
bergwolf, line 46: How about using github action to do a matrix build for GOOS{linux, windows, darwin} and GOARCH{amd64, arm64}?

eryugey (approved):
LGTM

Makefile
imeoer: Thanks for the comment, fixed.

PR #52, opened 2022-05-23
daemon: ignore checks for nydus binary when DaemonMode=none

power-more:
when DaemonMode=none, nydusd and nydus-image are not required.


config/config.go
```diff
@@ -86,6 +86,11 @@ func (c *Config) FillupWithDefaults() error {
 }
 
 func (c *Config) SetupNydusBinaryPaths() error {
+	// when DaemonMode = none, nydusd and nydus-image are not required
```
imeoer, line 89: `when using DaemonMode = none, nydusd and nydus-image binaries are not required`.

pkg/filesystem/nydus/config_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func TestWithNydusdBinaryPath(t *testing.T) {
 	var fs filesystem
-	opt := WithNydusdBinaryPath("/bin/nydusd")
+	opt := WithNydusdBinaryPath("/bin/nydusd", "multiple")
```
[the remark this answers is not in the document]
power-more: fixed in https://github.com/containerd/nydus-snapshotter/pull/57

config/config.go
power-more: fixed in https://github.com/containerd/nydus-snapshotter/pull/57

PR #55, opened 2022-05-26
add support for erofs with fscache

imeoer:
Ref: https://github.com/dragonflyoss/image-service/blob/master/docs/nydus-fscache.md

Add boot option `--daemon-backend erofs` for snapshotter to support
erofs with fscache daemon (nydusd daemon subcommand).

This daemon backend must be used with shared daemon mode, so a general
command usage should like this:

```
containerd-nydus-grpc \
  --config-path //path/to/nydus-erofs-config.json \
  --daemon-mode shared \
  --daemon-backend erofs \
  --log-level info \
  --root /var/lib/containerd/io.containerd.snapshotter.v1.nydus \
  --cache-dir /var/lib/nydus/cache \
  --address /run/containerd/containerd-nydus-grpc.sock \
  --nydusd-path /usr/bin/nydusd \
  --nydusimg-path /usr/bin/nydus-image \
  --log-to-stdout
```

### DIFF
--- a/cmd/containerd-nydus-grpc/pkg/command/flags.go
+++ b/cmd/containerd-nydus-grpc/pkg/command/flags.go
@@ -40,6 +40,7 @@ type Args struct {
 	NydusImageBinaryPath string
 	SharedDaemon         bool
 	DaemonMode           string
+	DaemonBackend        string
 	AsyncRemove          bool
 	EnableMetrics        bool
 	MetricsFile          string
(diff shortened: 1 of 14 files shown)


imeoer:
Renamed daemon backend name "erofs" to "fscache", let's update [doc](https://github.com/dragonflyoss/image-service/blob/master/docs/nydus-fscache.md) once the PR is merged. cc @hsiangkao

PR #56, opened 2022-05-27
config: support skip_verify backend option

imeoer:
Add `skip_verify: true` option to enable skipping SSL certificate
validation of HTTPS scheme for registry/oss backend.


eryugey (approved):
LGTM

PR #67, opened 2022-06-13
Convert: support nydus v6 fs version convert

zyfjeff:
1. add nydus v6 fs version support
2. add nydus v6 tests

Signed-off-by: zyfjeff <zyfjeff@linux.alibaba.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 env:
-  NYDUS_VERSION: v2.0.0-rc.5
+  NYDUS_VERSION: v2.1.0-alpha.0
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


pkg/converter/convert_unix.go
```diff
@@ -37,7 +37,7 @@ const envNydusWorkdir = "NYDUS_WORKDIR"
 func getBuilder() string {
 	builderPath := os.Getenv(envNydusBuilder)
 	if builderPath == "" {
-		builderPath = "nydus-image"
+		builderPath = "/root/project/image-service/target-fusedev/release/nydus-image"
```
hsiangkao: it'd be better to drop the immediate temporary modification?

zyfjeff: It's been fixed and what you seem to see is stale code

zyfjeff: Frankly speaking, some of the small projects I am involved in general only require a detailed description in the PR, and multiple commits in the PR are more casual and do not need to be merged and sorted out, and the commit message does not need to be so detailed.

hsiangkao (approved):

PR #69, opened 2022-06-14
args: change async-remove to sync-remove

power-more:
According to the usage of boolflag, the default value should
be false. While async-remove is enabled by default, so we modify args
to sync-remove.

### DIFF
--- a/cmd/containerd-nydus-grpc/pkg/command/flags.go
+++ b/cmd/containerd-nydus-grpc/pkg/command/flags.go
@@ -42,7 +42,7 @@ type Args struct {
 	SharedDaemon         bool
 	DaemonMode           string
 	DaemonBackend        string
-	AsyncRemove          bool
+	SyncRemove           bool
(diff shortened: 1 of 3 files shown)


cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -147,10 +147,9 @@ func buildFlags(args *Args) []cli.Flag {
 			Destination: &args.DaemonBackend,
 		},
 		&cli.BoolFlag{
-			Name:        "async-remove",
-			Value:       true,
+			Name:        "sync-remove",
 			Usage:       "whether to cleanup snapshots asynchronously",
```
imeoer, line 151: Should be `synchronously`.

power-more: fixed

imeoer (approved):
LGTM

PR #72, opened 2022-06-16
[WIP] Filesytem: support estargz for rafs v6

imeoer:


imeoer:
Waiting for https://github.com/dragonflyoss/image-service/pull/486 to be merged, and then we can add a smoke test case.

PR #74, opened 2022-06-20
Overlayfs: add volatile option when creating Active snapshot

power-more:
cri callers can add containerd.io/snapshot.volatile annotation to make
    corresponding snapshot volatile when mounted to rootfs.

    Overlayfs volatile mount is a new feature with Linux 5.10, for more
    details: https://www.redhat.com/sysadmin/container-volatile-overlay-mounts

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -239,7 +239,11 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 		}
 		return o.remoteMounts(ctx, *s, id, info.Labels)
 	}
-	return o.mounts(ctx, *s)
+	_, snap, _, err := snapshot.GetSnapshotInfo(ctx, o.ms, key)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("failed to get info for snapshot %s", key))
+	}
+	return o.mounts(ctx, &snap, *s)
 }
 
 func (o *snapshotter) prepareRemoteSnapshot(ctx context.Context, id string, labels map[string]string) error {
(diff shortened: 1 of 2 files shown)


hsiangkao (approved):

snapshot/snapshot.go
```diff
@@ -48,6 +48,10 @@ const (
 	NydusRootfsV6      string = "v6"
 	RafsV5SuperMagic   uint32 = 0x5241_4653
 	RafsSuperVersionV5 uint32 = 0x500
+	// volatileOpt is a key of an optional lablel to each snapshot.
+	// If this optional label of a snapshot is specified, when mounted to rootdir
+	// this snapshot will include volatile option
+	volatileOpt string = "containerd.io/snapshot/overlay.volatile"
```
imeoer, line 54: Better to move it to `pkg/label`.

snapshot/snapshot.go
```diff
@@ -274,7 +278,11 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 			return o.remoteMounts(ctx, *s, id, info.Labels)
 		}
 	}
-	return o.mounts(ctx, *s)
+	_, snap, _, err := snapshot.GetSnapshotInfo(ctx, o.ms, key)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get info for snapshot %s: %w", key, err)
```
imeoer, line 283: Should use `errors.Wrap`.

snapshot/snapshot.go
power-more: done

snapshot/snapshot.go
power-more: done

jiangliu:
What happens if the `volatile` option is used with older kernel? Will it cause breaks on older kernel?

power-more:
> What happens if the `volatile` option is used with older kernel? Will it cause breaks on older kernel?

it seems to print pr_err in logs, but would not break.

image
ref: https://lore.kernel.org/linux-unionfs/20200722175024.GA608248@redhat.com/

jiangliu (approved):

imeoer (approved):
LGTM

hsiangkao:
> > What happens if the `volatile` option is used with older kernel? Will it cause breaks on older kernel?
> 
> it seems to print pr_err in logs, but would not break.
> 
> image ref: https://lore.kernel.org/linux-unionfs/20200722175024.GA608248@redhat.com/

Nope, I think it will return -EINVAL directly. We should mount again if it happens.
``` c
 600                 default:
 601                         pr_err("unrecognized mount option \"%s\" or missing value\n",
 602                                         p);
 603                         return -EINVAL;
 604                 }
```

PR #75, opened 2022-06-20
config: Fix workdir json key not consist with nydusd fscache daemon

kevinXYin:
Use work_dir for json key as what nydusd needed, and create fscache work
dir before call add blob cmd to nydusd.

Signed-off-by: Xin Yin <yinxin.x@bytedance.com>
Reported-by: Yongqing Li <liyongqing@bytedance.com>

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -164,6 +164,10 @@ func (d *Daemon) sharedErofsMount() error {
 		return err
 	}
 
+	if err := os.MkdirAll(d.FscacheWorkDir(), 0755); err != nil {
+		return errors.Wrapf(err, "failed to create fscache work dir %s", d.FscacheWorkDir())
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -164,6 +164,10 @@ func (d *Daemon) sharedErofsMount() error {
 		return err
 	}
 
+	if err := os.MkdirAll(d.FscacheWorkDir(), 0755); err != nil {
```
imeoer, line 167: Maybe we can ensure this directory on snapshotter startup?

kevinXYin: This dir seems pre-image , and is generated by SnapshotDir and SnapshotID, should we use a fixed directory?

imeoer: Ok, indeed it's per image, sorry I miss that.

imeoer (approved):

Issue #54, opened 2022-05-25
Replace  standard http client to go-retryablehttp


zyfjeff:
At present, our snapshot will download the image through the go standard http client, if the http download fails and does not do a retry, ideally we should do a retry internally, we can replace the standard http client with go-retryablehttp

PR #76, opened 2022-06-23
resolver: replace standard http client to go-retryablehttp

power-more:
Previously, our snapshot will download the image through the go standard http
client, if the http download fails and does not do a retry. Now we retry
for defaultRetryMax = 4 times.

Fixes: #54


power-more:
> Could you please help to rebase the PR?

done

imeoer:
Thanks for the contribution, how about also replacing all `http.NewRequestWithContext` in [here](https://github.com/containerd/nydus-snapshotter/blob/main/pkg/filesystem/fs/stargz/resolver.go)?

imeoer (approved):

PR #77, opened 2022-06-24
Filesystem: refactor estargz support

imeoer:
Since the stargz filesystem implementation is very similar
to the nydus filesystem implementation, much of the code
can be shared, for example the shared daemon related codes.

We are going to merge them together, this PR moves codes,
and adjust the code path for stargz support.


pkg/filesystem/fs/fs.go
```diff
@@ -1,18 +1,59 @@
...
 
+// TODO: Move snapshotter needed all image annotations to nydus-snapshotter.
+const LayerAnnotationNydusBlobIDs = "containerd.io/snapshot/nydus-blob-ids"
+
+// RafsV6 layout: 1k + SuperBlock(128) + SuperBlockExtener(256)
+// RafsV5 layout: 8K superblock
+// So we only need to read the MaxSuperBlockSize size to include both v5 and v6 superblocks
+const MaxSuperBlockSize = 8 * 1024
+const RafsV6Magic = 0xE0F5E1E2
+const ChunkInfoOffset = 1024 + 128 + 24
```
bergwolf, line 50: RafsV6ChunkInfoOffset?

pkg/filesystem/fs/fs.go
```diff
@@ -22,16 +63,732 @@ const (
...
+}
+
+func (fs *Filesystem) PrepareBlobLayer(ctx context.Context, snapshot storage.Snapshot, labels map[string]string) error {
+	start := time.Now()
+	defer func() {
+		duration := time.Since(start)
+		log.G(ctx).Infof("total nydus prepare data layer duration %d", duration.Milliseconds())
+	}()
+
+	if fs.imageMode == OnDemand {
```
bergwolf, line 157: Move it to before the duration calculation? No need to calculate duration for preload mode.

pkg/filesystem/fs/fs.go
```diff
@@ -22,16 +63,732 @@ const (
...
+		return errors.Wrap(err, "Unexpected error, we can't handle it")
+	}
+
+	readerCloser, err := fs.resolver.Resolve(ref, layerDigest, labels)
+	if err != nil {
+		return errors.Wrapf(err, "failed to resolve from ref %s, digest %s", ref, layerDigest)
+	}
+	defer readerCloser.Close()
+
+	blobFile, err := os.OpenFile(blobPath, os.O_CREATE|os.O_RDWR, 0755)
```
bergwolf, line 183: `0666`?

pkg/filesystem/fs/fs.go
bergwolf: Also it's better to check `fs.imageMode == Preload` directly.

pkg/filesystem/fs/fs.go
```diff
@@ -22,16 +63,732 @@ const (
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create blob file")
+	}
+	defer blobFile.Close()
+	_, err = io.Copy(blobFile, readerCloser)
```
bergwolf, line 188: remove blob file upon error?

pkg/filesystem/fs/fs.go
imeoer: Fixed, actually, these codes just moved from other place. :)

pkg/filesystem/fs/fs.go
imeoer: Fixed

pkg/filesystem/fs/fs.go
imeoer: Fixed

hsiangkao (approved):

PR #80, opened 2022-06-27
feat: NYDUS files are convertible to OCI tar

ccx1024cc:
BACKGROUND
OCI layer is convertible into bootstrap and blob of NYDUS. However there is no way to convert bootstrap and blob to OCI layer.

REQUIREMENT
BuildKit already provides ability to convert OCI layer to NYDUS files. It seeks for the ability to convert back.

FEATURE
Excluding hard link, the hash digests of result and original OCI tar are same .
Including hard link, the unpack result of result and original OCI tar are same.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -30,6 +30,7 @@ import (
 )
 
 const bootstrapNameInTar = "image.boot"
+const blobNameInTar = "image.blob"
 
 const envNydusBuilder = "NYDUS_BUILDER"
 const envNydusWorkdir = "NYDUS_WORKDIR"
@@ -46,7 +47,7 @@ func getBuilder() string {
 func unpackOciTar(ctx context.Context, dst string, reader io.Reader) error {
 	ds, err := compression.DecompressStream(reader)
 	if err != nil {
-		return errors.Wrap(err, "decompress stream")
+		return errors.Wrap(err, "unpack stream")
 	}
 	defer ds.Close()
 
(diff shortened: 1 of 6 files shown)


pkg/converter/convert_unix.go
```diff
@@ -124,7 +151,62 @@ func unpackBootstrapFromNydusTar(ctx context.Context, ra content.ReaderAt, targe
...
+				return errors.Wrap(err, "seek to blob data offset")
+			}
+			if _, err := io.CopyN(target, reader, hdr.Size); err != nil {
+				return errors.Wrap(err, "copy blob data to reader")
+			}
+			return nil
+		}
+	}
+
+	logrus.Info("Blob is not found in nydus tar")
```
imeoer, line 204: It is not necessary to print this log.

pkg/converter/convert_unix.go
```diff
@@ -279,3 +361,43 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
...
+	defer blobFifo.Close()
+
+	go func() {
+		err := tool.Decompress(tool.DecompressOption{
+			BuilderPath:   getBuilder(),
+			BootstrapPath: bootPath,
+			BlobPath:      blobPath,
+			TarPath:       tarPath,
+		})
+		if err != nil {
```
imeoer, line 393: Can we return this `err`?

tests/converter_test.go
```diff
@@ -150,13 +151,21 @@ func buildOCILowerTar(t *testing.T) io.ReadCloser {
...
+func buildOCIUpperTar(t *testing.T, teePath string) io.ReadCloser {
 	pr, pw := io.Pipe()
-	tw := tar.NewWriter(pw)
 
 	go func() {
+		tw := tar.NewWriter(pw)
 		defer pw.Close()
 
+		if len(teePath) > 0 {
+			teew, err := os.OpenFile(teePath, os.O_WRONLY|os.O_CREATE, 0755)
```
imeoer, line 162: `0644`?

imeoer:
Should we name it pack/unpack?

pkg/converter/convert_unix.go
```diff
@@ -279,3 +358,49 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
...
+			blobFifo.Close()
+			unpackErrChan <- err
+		}
+	}()
+
+	if _, err := io.Copy(dest, blobFifo); err != nil {
+		if unpackErr := <-unpackErrChan; unpackErr != nil {
+			return errors.Wrap(err, "unpack")
```
imeoer, line 400: `return errors.Wrap(unpackErr, "unpack")` ?

imeoer (approved):
LGTM！

PR #81, opened 2022-06-28
fscache: fix possible conflict fscache id

imeoer:
Maybe image ref causes fscache id conflict when user updates
image content using same image ref. So it's more safe to use
snapshot id instead of image ref as the fscache id.

### DIFF
--- a/config/daemonconfig.go
+++ b/config/daemonconfig.go
@@ -132,7 +132,7 @@ func SaveConfig(c interface{}, configFile string) error {
 	return ioutil.WriteFile(configFile, b, 0755)
 }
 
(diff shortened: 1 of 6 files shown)


jiangliu (approved):

imeoer:
> > Maybe image ref causes fscache id conflict when user updates image content using same image ref. So it's more safe to use snapshot id instead of image ref as the fscache id.
> > Signed-off-by: Yan Song [yansong.ys@antfin.com](mailto:yansong.ys@antfin.com)
> 
> Will this PR solve the possible confliction with OCIv1 layers?

No, it's an unrelated problem.

PR #82, opened 2022-06-29
fscache: fix fscache workdir path for estargz

imeoer:
Before this patch, we share the same fscache workdir for all estargz
images, this workdir stores all blob meta files for each estargz image.

The fatal problem is that once an estargz image is removed by user, the
all blob meta files will be cleaned by containerd GC, this behavior affects
all existing estargz images.

This patch stores the blob meta files into each snapshot dir, and then
copy them into topmost snapshot dir if user wants to run a container.


hsiangkao:
Yeah, I noticed this issue as well, thanks for this!

hsiangkao (approved):

PR #98, opened 2022-07-15
Enable log for fscache

jiangliu:
Several enhancements:
1) enable log for fscache
2) avoid deleting non-socket file when removing old socket file
3) refine error handling logic in createSnapshot

### DIFF
--- a/cmd/containerd-nydus-grpc/app/snapshotter/serve.go
+++ b/cmd/containerd-nydus-grpc/app/snapshotter/serve.go
@@ -50,9 +50,12 @@ func ensureSocketNotExists(listeningSocketPath string) error {
 	if err := os.MkdirAll(filepath.Dir(listeningSocketPath), 0700); err != nil {
 		return errors.Wrapf(err, "failed to create directory %q", filepath.Dir(listeningSocketPath))
 	}
-	_, err := os.Stat(listeningSocketPath)
+	finfo, err := os.Stat(listeningSocketPath)
 	// err is nil means listening socket path exists, remove before serve
 	if err == nil {
+		if finfo.Mode()&os.ModeSocket == 0 {
+			return errors.Errorf("file %s is not a socket", listeningSocketPath)
+		}
 		err := os.Remove(listeningSocketPath)
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


pkg/process/manager.go
```diff
@@ -180,6 +167,16 @@ func (m *Manager) buildStartCommand(d *daemon.Daemon) (*exec.Cmd, error) {
...
 	}
 
+	args = append(args, "--apisock", d.GetAPISock())
+	args = append(args, "--log-level", d.LogLevel)
+	if !d.LogToStdout {
+		args = append(args, "--log-file", d.LogFile())
+	}
+	nydusdThreadNum := d.NydusdThreadNum()
+	if nydusdThreadNum != "" {
+		args = append(args, "--thread-num", nydusdThreadNum)
```
changweige, line 177: `--thread-num` only works for FUSE daemon?

jiangliu: sorry, it should be "--fscache-threads" for fscache, fixed.

cmd/containerd-nydus-grpc/app/snapshotter/serve.go
```diff
@@ -50,9 +50,13 @@ func ensureSocketNotExists(listeningSocketPath string) error {
 	if err := os.MkdirAll(filepath.Dir(listeningSocketPath), 0700); err != nil {
 		return errors.Wrapf(err, "failed to create directory %q", filepath.Dir(listeningSocketPath))
 	}
-	_, err := os.Stat(listeningSocketPath)
+	finfo, err := os.Stat(listeningSocketPath)
 	// err is nil means listening socket path exists, remove before serve
 	if err == nil {
+		if finfo.Mode()&os.ModeSocket == 0 {
+			return errors.Errorf("file %s is not a socket", listeningSocketPath)
+
```
imeoer, line 58: Maybe format the code.

snapshot/snapshot.go
```diff
@@ -539,10 +536,8 @@ func (o *snapshotter) createSnapshot(ctx context.Context, kind snapshots.Kind, k
 			return nil, storage.Snapshot{}, errors.Wrap(err, "failed to stat parent")
 		}
 
+		// FIXME: Why only change owner of having parent?
 		if err := lchown(filepath.Join(td, "fs"), st); err != nil {
```
imeoer, line 540: It seems not to change the `err` in func's topmost scope, so I think `cleanupSnapshotDirectory` can't be called properly.

changweige (approved):
LGTM

PR #99, opened 2022-07-15
Fix some spelling mistakes, rename nydusd log file

changweige:
Provide a flag FS_CACHE=1 to `make install` thus to easily setup nydus-snapshotter with fs driver  fscache.


jiangliu (approved):

hsiangkao (approved):

misc/snapshotter/nydus-snapshotter.fscache.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=nydus snapshotter
+After=network.target
+Before=containerd.service
+
+[Service]
+Type=simple
+Environment=HOME=/root
+ExecStart=/usr/local/bin/containerd-nydus-grpc --fs-driver fscache --daemon-mode shared --config-path /etc/nydus/config.json
```
imeoer, line 9: Can users install both `fusedev` and `fscache`? If so, we'd better distinguish the config file.

PR #101, opened 2022-07-18
Second PR to enhance the snapshotter implementation

jiangliu:
Second PR to enhance the snapshotter implementation with:
1) ensure atomicity when downloading/generating files
2) remove dead code
3) optimize stargz implementation by reducing duplicated work
4) improve documentation and const definitions
5) enforce stricter permission for files/directories


snapshot/snapshot.go
```diff
@@ -254,27 +254,17 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
-	logCtx.Infof("Preparing. Key=%s, Parent=%s, Labels %v", key, parent, base.Labels)
+	logCtx.Infof("Preparing snapshot with labels %v", base.Labels)
 	if target, ok := base.Labels[label.TargetSnapshotRef]; ok {
-		// check if image layer is nydus layer
+		// check if image layer is nydus data layer
 		if o.fs.Support(ctx, base.Labels) {
-			logCtx.Infof("nydus data/meta layer, skip download and unpack %s", key)
-			_, isBootstrapLayer := base.Labels[label.NydusMetaLayer]
-			if isBootstrapLayer {
-				err = o.fs.PrepareMetaLayer(ctx, s, base.Labels)
```
imeoer, line 269: Some confusion, why not prepare the meta layer?

pkg/filesystem/fs/fs.go
```diff
@@ -236,6 +236,10 @@ func (fs *Filesystem) newSharedDaemon() (*daemon.Daemon, error) {
 	return d, nil
 }
 
+func (fs *Filesystem) StargzEnabled() bool {
+	return fs.stargzResolver != nil
+}
+
 func (fs *Filesystem) SupportStargz(ctx context.Context, labels map[string]string) (bool, string, string, *stargz.Blob) {
 	if fs.stargzResolver == nil {
```
imeoer, line 244: Maybe use `if !fs.StargzEnabled() {`

snapshot/snapshot.go
```diff
@@ -275,14 +275,14 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
-				logCtx.Errorf("failed to prepare stargz layer of snapshot ID %s, err: %v", s.ID, err)
-			} else {
-				// Mark this snapshot as stargz layer
-				base.Labels[label.StargzLayer] = "true"
-				err := o.Commit(ctx, target, key, append(opts, snapshots.WithLabels(base.Labels))...)
-				if err == nil || errdefs.IsAlreadyExists(err) {
-					return nil, errors.Wrapf(errdefs.ErrAlreadyExists, "target snapshot %q", target)
-				}
+				logCtx.Errorf("prepare stargz layer of snapshot ID %s, err: %v", s.ID, err)
+				return nil, errors.Wrap(err, "prepare stargz image layer")
```
imeoer, line 279: If prepare stargz meta layer fails, we need to fall back to OCI non-lazy mode, so should ignore the error here.

snapshot/snapshot.go
jiangliu: The bootstrap file is handled by containerd.
And the `o.fs.Support()` checks `labels[label.NydusDataLayer]`, so it filters out bootstrap files too.

snapshot/snapshot.go
jiangliu: Ok, didn't realized that and will revert it.

snapshot/snapshot.go
imeoer: Ok, that make sense.

pkg/filesystem/fs/fs.go
jiangliu: Fixed

PR #114, opened 2022-07-28
Optimize snapshotter to reduce container startup time

jiangliu:
Optimize snapshotter to reduce container startup time by
1) avoid probing stargz layer for nydus metadata layer.
2) start shared nydusd daemon early, so it doesn't penalize the first container.


imeoer (approved):
LGTM!

changweige (approved):
LGTM

Issue #110, opened 2022-07-26
v6 meta files are not deleted from cache dir when removing container images


changweige:
Perform removing container images by `nerdctl image prune --all`, only data blob files are deleted while meta files are residual.

PR #115, opened 2022-07-28
Remove blob.meta files when GC nydus blob files

jiangliu:
Remove blob.meta files when deleting nydus blob files.

Fixes: https://github.com/containerd/nydus-snapshotter/issues/110

Signed-off-by: Jiang Liu <gerry@linux.alibaba.com>

### DIFF
--- a/pkg/cache/store.go
+++ b/pkg/cache/store.go
@@ -9,6 +9,7 @@ import (
 
 const (
 	chunkMapFileSuffix = ".chunk_map"
+	metaFileSuffix     = ".blob.meta"
 )
 
 type Store struct {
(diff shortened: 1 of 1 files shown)


pkg/cache/store.go
```diff
@@ -37,6 +38,11 @@ func (cs *Store) DelBlob(blob string) error {
 		return errors.Wrapf(err, "remove blob %v err", blobPath)
```
changweige, line 38: Even blob and chunk map deletion fails, we still try to remove blob.meta?

jiangliu: It would be safer to keep the blob.meta file in such a case.

changweige (approved):
LGTM

PR #118, opened 2022-08-01
enhance how nydusd's log are logged and redirected

changweige:
If nydusd does not make to reach logging setup when starting, it can not log any errors to its logger.
At present, nydusd's stderr is only redirected to snapshotter with `--log-to-stdout` enabled. But the option 
should only control nydusd's logs rather than its stderr and stdout.

This helps us with analyzing  failed started container

### DIFF
--- a/pkg/process/manager.go
+++ b/pkg/process/manager.go
@@ -184,10 +184,11 @@ func (m *Manager) buildStartCommand(d *daemon.Daemon) (*exec.Cmd, error) {
 	log.L.Infof("start nydus daemon: %s %s", m.nydusdBinaryPath, strings.Join(args, " "))
 
 	cmd := exec.Command(m.nydusdBinaryPath, args...)
-	if d.LogToStdout {
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stdout
-	}
+	// nydusd standard output and standard error rather than its logs are
+	// always redirected to snapshotter's respectively
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
 	return cmd, nil
 }
 
--- a/misc/snapshotter/nydus-snapshotter.fscache.service
+++ b/misc/snapshotter/nydus-snapshotter.fscache.service
@@ -11,6 +11,8 @@ Restart=always
 RestartSec=1
 KillMode=process
 OOMScoreAdjust=-999
+StandardOutput=journal
+StandardError=journal
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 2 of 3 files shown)


pkg/process/manager.go
```diff
@@ -234,10 +241,14 @@ func (m *Manager) DestroyDaemon(d *daemon.Daemon) error {
...
-		// if nydus-snapshotter restart, it will break the relationship between nydusd and
+
+		// if nydus-snapshotter restarts, it will break the relationship between nydusd and
 		// nydus-snapshotter, p.Wait() will return err, so here should exclude this case
-		if _, err = p.Wait(); err != nil && !stderrors.Is(err, syscall.ECHILD) {
-			log.L.Errorf("failed to process wait, %v", err)
+		if state, err := p.Wait(); err == nil {
+			// Expect to catch failure exit status.
+			d.SuccessStatus = state.Success()
+		} else if err != nil && !stderrors.Is(err, syscall.ECHILD) {
```
imeoer, line 250: Maybe we can give a warning for `ECHILD` status.

pkg/process/manager.go
```diff
@@ -184,10 +190,11 @@ func (m *Manager) buildStartCommand(d *daemon.Daemon) (*exec.Cmd, error) {
 	log.L.Infof("start nydus daemon: %s %s", m.nydusdBinaryPath, strings.Join(args, " "))
 
 	cmd := exec.Command(m.nydusdBinaryPath, args...)
-	if d.LogToStdout {
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stdout
-	}
+	// nydusd standard output and standard error rather than its logs are
+	// always redirected to snapshotter's respectively
+	cmd.Stdout = os.Stdout
```
imeoer, line 195: Seems no logs will be output to stdout/stderr when users specify `--log-file`.

pkg/process/manager.go
changweige: Generally, it is a normal situation. But I can add a warning log entry there.

imeoer:
My suggestion is that snapshotter's own --log-file option will determine whether snapshotter + nydusd logs are put in stderr/stdout or log file, which is easiest to check for errors and easier for the user to understand.

pkg/process/manager.go
changweige: This code snippet is no longer touched.

changweige:
> My suggestion is that snapshotter's own --log-file option will determine whether snapshotter + nydusd logs are put in stderr/stdout or log file, which is easiest to check for errors and easier for the user to understand.

SGTM. We can raise an issue to track the idea.

imeoer (approved):

PR #119, opened 2022-08-01
remove nydusd daemon mode fuse when starting

changweige:
`fuse` subcommand is not compatible with old nydusd.
So remove it and new nydusd can use fuse as default mode.


imeoer (approved):

hsiangkao:
please revert this commit, thanks!

PR #120, opened 2022-08-01
Fix BlobMgr Remove

power-more:
Previously in the localfs backend mode, when blobMgr delete the blob
corresponding to the snapshot, it mistakenly used the key (chainID)
issued by containerd as the blob digest, resulting in only the first
layer of blob being deleted.
Now
1. For the scenario of converting OCI to Nydus image, get the blob
digest, in fact the snapshot key by label NydusDataLayer, the kv pair
(NydusDataLayer --> snapshot key) has been set after the convert.
2. For the scenario of full download of nydus image, get the correct
blob digest by label CRILayerDigest.

### DIFF
--- a/misc/snapshotter/nydusd-config-localfs.json
+++ b/misc/snapshotter/nydusd-config-localfs.json
@@ -3,7 +3,7 @@
       "backend": {
         "type": "localfs",
         "config": {
-            "dir": "/tmp/"
+            "dir": "/tmp/blobs/"
         }
       },
       "cache": {
--- a/pkg/filesystem/fs/fs.go
+++ b/pkg/filesystem/fs/fs.go
@@ -144,11 +144,11 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
 	return &fs, nil
 }
 
-func (fs *Filesystem) CleanupBlobLayer(ctx context.Context, key string, async bool) error {
+func (fs *Filesystem) CleanupBlobLayer(ctx context.Context, blobDigest string, async bool) error {
 	if fs.blobMgr == nil {
 		return nil
 	}
-	return fs.blobMgr.Remove(key, async)
+	return fs.blobMgr.Remove(blobDigest, async)
 }
 
 func (fs *Filesystem) PrepareBlobLayer(ctx context.Context, snapshot storage.Snapshot, labels map[string]string) error {
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -402,9 +402,10 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to remove")
 	}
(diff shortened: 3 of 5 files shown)


zyfjeff:
Can you help improve this integration test and add a case for image deletion?

https://github.com/containerd/nydus-snapshotter/blob/main/.github/workflows/image.yml#L117

power-more:
> Can you help improve this integration test and add a case for image deletion?
> 
> https://github.com/containerd/nydus-snapshotter/blob/main/.github/workflows/image.yml#L117

done

zyfjeff (approved):
LGTM

misc/snapshotter/nydusd-config-localfs.json
```diff
@@ -3,7 +3,7 @@
       "backend": {
         "type": "localfs",
         "config": {
-            "dir": "/tmp/"
+            "dir": "/tmp/blobfs/"
```
imeoer, line 6: Maybe rename it to `/tmp/blobs`.

pkg/filesystem/fs/fs.go
```diff
@@ -143,11 +143,11 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
 	return &fs, nil
 }
 
-func (fs *Filesystem) CleanupBlobLayer(ctx context.Context, key string, async bool) error {
+func (fs *Filesystem) CleanupBlobLayer(ctx context.Context, id string, async bool) error {
```
imeoer, line 146: `id` -> `blobDigest` is better.

snapshot/snapshot.go
```diff
@@ -403,8 +403,9 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
 		return errors.Wrap(err, "failed to remove")
 	}
 	var cleanupBlob bool
-	if _, ok := snap.Labels[label.NydusDataLayer]; ok {
-		cleanupBlob = true
+	blobDigestString, cleanupBlob := snap.Labels[label.NydusDataLayer]
```
imeoer, line 406: No type is required in a variable name, just use `blobDigest`.

pkg/filesystem/fs/fs.go
power-more: Fixed.
The reason I didn't use blobDigest is that i thought, the type of blobDigest is preferably digest.Digest in general.

snapshot/snapshot.go
```diff
@@ -403,8 +403,9 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
...
+	if cleanupBlob && blobDigestString == "true" {
```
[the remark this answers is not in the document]
power-more: Fixed.
In fact, we don't even need to care what value is, so i update it to _

snapshot/snapshot.go
power-more: fixed

imeoer (approved):

PR #124, opened 2022-08-04
cache manager: fix cache manager failure in none daemon mode

power-more:
Previously, both AddSnapshot and DelSnapshot will not be executed in
the none daemon mode. The fundamental reason is that the incoming
parameter is the imageID obtained from the daemon. Now we adjust the
function call path to bypass the daemon and get the imageID directly
from snapshot labels.

For AddSnapshot, we separate it from prepareRemoteSnapshot, run it
separately, and parse the imageID directly from the snapshot label.

For DelSnapshot, we advance it from Cleanup to Remove.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -300,6 +300,11 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 			}
 
 			logCtx.Infof("found nydus meta layer id %s, parpare remote snapshot", id)
+
+			if err := o.fs.AddSnapshot(info.Labels); err != nil {
+				return nil, errors.Wrap(err, "cache manager failed to add snapshot")
+			}
+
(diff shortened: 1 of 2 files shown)


snapshot/snapshot.go
```diff
@@ -300,6 +300,11 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 			}
 
 			logCtx.Infof("found nydus meta layer id %s, parpare remote snapshot", id)
+
+			if err := o.fs.AddSnapshot(info.Labels); err != nil {
```
imeoer, line 304: Why only do `AddSnapshot` when the container running? If the user deletes the image before any container run, can nydus blobs be GC-ed?

imeoer: Okay I miss it, blob cache is generated at only container running.

imeoer (approved):
LGTM

PR #127, opened 2022-08-10
Add helper functions for containerd image convertion

loheagn:
Add some functions which are helpful for using the `Convert` function in the containerd client library to convert OCI images to Nydus images.

For example, we can use these functions to support nerdctl to convert Nydus images.


pkg/converter/constant.go
```diff
@@ -13,6 +13,9 @@ const (
 
 	ManifestNydusCache = "containerd.io/snapshot/nydus-cache"
 
+	ConfigGCLabelKey = "containerd.io/gc.ref.content.config"
```
imeoer, line 16: Seems it's a private constant, maybe put it into `convert_unix.go` and name `configGCLabelKey`?

pkg/converter/convert_unix.go
```diff
@@ -423,3 +431,237 @@ func Unpack(ctx context.Context, ia content.ReaderAt, dest io.Writer, opt Unpack
 
 	return nil
 }
+
+// ConvertToNydusLayer returns a function which converts an OCI image layer to a nydus blob layer, and set the media type to "application/vnd.oci.image.layer.nydus.blob.v1".
+func ConvertToNydusLayer(opt PackOption) converter.ConvertFunc {
```
imeoer, line 436: Maybe name it `LayerConvertFunc`.

pkg/converter/convert_unix.go
```diff
@@ -423,3 +431,237 @@ func Unpack(ctx context.Context, ia content.ReaderAt, dest io.Writer, opt Unpack
...
+		}
+
+		return &newDesc, nil
+	}
+}
+
+// ConversionHook returns a function which will be used as a callback called for each blob after conversion is done.
+// The function only hooks the manifest conversion.
+// The function merges all the nydus blob layers into a nydus bootstrap layer and modify the config.
+func ConversionHook(opt PackOption) converter.ConvertHookFunc {
```
imeoer, line 513: Better to name it `ConvertHookFunc`.

imeoer (commented):
Thanks for the contribution, others LGTM. Could you help to update `https://github.com/goharbor/acceleration-service` to import these funcs? Thanks!

loheagn:
Thanks for the review @imeoer . I think it's ok to update `acceleration-service` with these functions, but would require some modifications to the current code, like add `backend` option.

imeoer:
> Thanks for the review @imeoer . I think it's ok to update `acceleration-service` with these functions, but would require some modifications to the current code, like add `backend` option.

@loheagn Maybe we can move the backend codes after the LayerConvertFunc function call. :)

loheagn:
> @loheagn Maybe we can move the backend codes after the LayerConvertFunc function call. :)

The `backend` option also has some side effects in the manifest convertion:

[nydus.go#L178](https://github.com/goharbor/acceleration-service/blob/6c796890b3c470fb8c8d48c11dd5cce0958efac9/pkg/driver/nydus/nydus.go#L178)

[nydus.go#L197](https://github.com/goharbor/acceleration-service/blob/6c796890b3c470fb8c8d48c11dd5cce0958efac9/pkg/driver/nydus/nydus.go#L197)

imeoer:
@loheagn Ok, sorry I miss that, how about moving backend codes into snapshotter as well?

loheagn:
@imeoer I think it's ok. We just need to add a param to the function, and the default value of the param can be nil.

imeoer:
> @imeoer I think it's ok. We just need to add a param to the function, and the default value of the param can be nil.

Thanks for the work!

pkg/converter/convert_unix.go
```diff
@@ -423,3 +434,259 @@ func Unpack(ctx context.Context, ia content.ReaderAt, dest io.Writer, opt Unpack
 
 	return nil
 }
+
+// LayerConvertFunc returns a function which converts an OCI image layer to a nydus blob layer, and set the media type to "application/vnd.oci.image.layer.nydus.blob.v1".
+func LayerConvertFunc(opt PackOption, backend backend.Backend) converter.ConvertFunc {
```
imeoer, line 439: Would be better if put `backend` into `PackOption`.

pkg/converter/convert_unix.go
```diff
@@ -423,3 +434,259 @@ func Unpack(ctx context.Context, ia content.ReaderAt, dest io.Writer, opt Unpack
...
+		}
+
+		return &newDesc, nil
+	}
+}
+
+// ConvertHookFunc returns a function which will be used as a callback called for each blob after conversion is done.
+// The function only hooks the manifest conversion.
+// The function merges all the nydus blob layers into a nydus bootstrap layer and modify the config.
+func ConvertHookFunc(opt PackOption, backend backend.Backend) converter.ConvertHookFunc {
```
imeoer, line 528: Ditto.

go.mod
```diff
@@ -32,9 +33,9 @@ require (
 )
 
 require (
-	github.com/KarpelesLab/reflink v0.0.2 // indirect
 	github.com/Microsoft/go-winio v0.5.1 // indirect
 	github.com/Microsoft/hcsshim v0.9.3 // indirect
+	github.com/baiyubin/aliyun-sts-go-sdk v0.0.0-20180326062324-cfa1a18b161f // indirect
```
imeoer, line 38: Is this package still necessary after `go mod tidy`?

jiangliu, line 38: This dependency seems a little strange, like a personal repo?

pkg/converter/convert_unix.go
loheagn: done

pkg/converter/convert_unix.go
loheagn: done

go.mod
loheagn: Yes. This package seems to be imported by `github.com/aliyun/aliyun-oss-go-sdk`.

loheagn: This package seems to be imported by `github.com/aliyun/aliyun-oss-go-sdk`.

imeoer (approved):
LGTM, one remaining suggestion is to explain why & how in the commit message if possible, like https://github.com/containerd/nydus-snapshotter/commit/60cc64fa274cf777876659bb0cab9ab2aa2cb338.

loheagn:
> LGTM, one remaining suggestion is to explain why & how in the commit message if possible, like [60cc64f](https://github.com/containerd/nydus-snapshotter/commit/60cc64fa274cf777876659bb0cab9ab2aa2cb338).

Thanks @imeoer ! I have squshed the commits and modified the commit message.

imeoer (approved):

changweige (approved):

PR #129, opened 2022-08-12
close pipe returned files objects.

changweige:
Otherwise, they are leaked.


imeoer:
It's not a real problem, we have commented here about Closer:

https://github.com/containerd/nydus-snapshotter/blob/e6cc1fa69ab35dcd88424d680e2171ad978c7ce1/pkg/converter/convert_unix.go#L246-L247

And io.PipeReader does not need to be closed explicitly.

PR #132, opened 2022-08-15
clean up and fix how nydus-snapshotter prepares mountpoints for nydusd

changweige:
1. Erofs/fscache can't work in ordinary SharedMount mode since each erofs has a seperated mountpoint while shareMode means one nydusd one mountpoint.
2. rafs as lowerdir and read-only underlying FS should be mounted at `/mnt`. `fs` is used to store bootstrap for meta layer and writeable for Active snapshot
3. `fs` has permission 0700, which causes other user in the container can't read /bin dir

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -77,6 +77,8 @@ func WithLogLevel(logLevel string) NewDaemonOpt {
 	}
 }
 
+// In Shared mode, it is the mountpoint where single nydusd mounts fusedev. It has nothing
+// with fscache driver.
 func WithRootMountPoint(rootMountPoint string) NewDaemonOpt {
 	return func(d *Daemon) error {
 		if err := os.MkdirAll(rootMountPoint, 0755); err != nil {
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -77,8 +77,7 @@ func WithLogLevel(logLevel string) NewDaemonOpt {
 	}
 }
 
-// In Shared mode, it is the mountpoint where single nydusd mounts fusedev. It has nothing
-// with fscache driver.
+// In Shared mode, it is the mountpoint where single nydusd mounts
 func WithRootMountPoint(rootMountPoint string) NewDaemonOpt {
 	return func(d *Daemon) error {
 		if err := os.MkdirAll(rootMountPoint, 0755); err != nil {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -54,8 +54,17 @@ type Daemon struct {
 	Once   *sync.Once         `json:"-"`
 }
 
+// Mountpoint for nydusd within single kernel mountpoint(FUSE mount). Each mountpoint
+// is create by API based pseudo mount. `RootMountPoint` is real mountpoint
+// where to perform the kernel mount.
+// Nydusd API based mountpoint must start with "/", otherwise nydusd API server returns error.
 func (d *Daemon) SharedMountPoint() string {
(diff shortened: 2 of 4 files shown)


snapshot/snapshot.go
```diff
@@ -479,11 +479,16 @@ func (o *snapshotter) Close() error {
 }
 
 func (o *snapshotter) upperPath(id string) string {
+
```
imeoer, line 482: Should remove this blank line.

pkg/filesystem/fs/fs.go
```diff
@@ -62,9 +62,15 @@ var nativeEndian binary.ByteOrder
 type Mode int
 
 const (
+	// A single nydusd serves all container images
 	SharedInstance Mode = iota
+	// One container image one nydusd topology
 	MultiInstance
+	// Rafs is integrated with other custom components.
+	// Nydusd is not needed to work as FUSE server or other backend service.
```
imeoer, line 70: Nydusd is not needed to work as FUSE server (fusedev) or other fs drivers, just provide the mount slices to containerd, and do real mount by other components.

pkg/filesystem/fs/fs.go
```diff
@@ -62,9 +62,15 @@ var nativeEndian binary.ByteOrder
...
+	// Snapshotter just cooperates with containerd with mount slice
 	NoneInstance
+	// Nydusd does not fulfill any concrete image service but prefetch data layers into local cache
```
imeoer, line 73: Nydusd does not fulfill any lazy pull (on-daemon) ability but only prefetch image data into local blob cache for other components.

pkg/daemon/daemon.go
```diff
@@ -67,14 +67,14 @@ func (d *Daemon) SharedAbsMountPoint() string {
 	return filepath.Join(*d.RootMountPoint, d.SharedMountPoint())
 }
 
+// Mountpoint of Erofs/RafsV6 and per-image nydusd/rafs
```
imeoer, line 70: Does RAFS v6 support shared mode?

pkg/daemon/config.go
```diff
@@ -77,6 +77,8 @@ func WithLogLevel(logLevel string) NewDaemonOpt {
 	}
 }
 
+// In Shared mode, it is the mountpoint where single nydusd mounts fusedev. It has nothing
+// with fscache driver.
 func WithRootMountPoint(rootMountPoint string) NewDaemonOpt {
```
imeoer, line 82: We cannot assume that the fscache driver does not support multiple daemon mode, it is better not to bind the mountpoint implementation with the meaning of daemon mode.

snapshot/snapshot.go
changweige: fixed

pkg/filesystem/fs/fs.go
changweige: fixed

pkg/filesystem/fs/fs.go
changweige: fixed

pkg/daemon/daemon.go
changweige: fixed. New comment is 
// Mountpoint of per-image nydusd/rafs. It is a kernel mountpoint for each
// nydus meta layer. Each meta layer is associated with a nydusd.

pkg/daemon/config.go
changweige: fixed

imeoer (approved):
LGTM, thanks for this.

PR #136, opened 2022-08-16
use xid to generate nydusd's internal ID

changweige:
The original ID generation method spawns a string with
lower case, upper case and undersocre, which is too complicated.


imeoer (approved):
LGTM

PR #138, opened 2022-08-18
converter: make backend package independent

imeoer:
Move `pkg/converter/backend` to `pkg/backend`.

The advantage of this change is that if a component imports the
converter package, but it doesn't need backend support, the non-direct
dependencies in `pkg/backend` don't have to be imported either.


changweige:
Can we rename backend to storage?

changweige (approved):
LGTM

PR #139, opened 2022-08-22
Change upperPath permission from 0750 to 0755

zyfjeff:
The user of ctr/nerdctl may be different from the user of nydus snapshotter.
In order to allow users of ctr and nerdctl to access rootfs, other group permissions need to be given here,
and the default permission of overlayfs is also 0755

Signed-off-by: zyfjeff <zyfjeff@linux.alibaba-inc.com>


changweige (approved):
LGTM

changweige:
Good catch, thanks

PR #140, opened 2022-08-23
Add containerized e2e test

changweige:
Then developers can run e2e anywhere rather than only github action


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    branches:
+      - "main"
+
+env:
+  CARGO_TERM_COLOR: always
+  REGISTRY: ghcr.io
+
+jobs:
+  run-e2e:
```
imeoer, line 18: Is the job duplicated with [this](https://github.com/containerd/nydus-snapshotter/blob/main/.github/workflows/image.yml) ?

changweige: Yes, you are right. In fact, most containerized e2e test cases are copied from it with replacing crictl with nerdctl. The stargz case is still not migrated. So let's keep them two run concurrently for a while until stargz case is added.

imeoer: Maybe we should remove it from this PR? otherwise nobody knows we need to remove the duplicated job in the future.

changweige: It is deleted.

imeoer:
LGTM

Makefile
```diff
@@ -35,11 +35,12 @@ clear:
...
 
 .PHONY: install
-install: static-release
+install:
 	sudo install -D -m 755 bin/containerd-nydus-grpc /usr/local/bin/containerd-nydus-grpc
 	sudo install -D -m 755 misc/snapshotter/nydusd-config.${FS_DRIVER}.json /etc/nydus/config.json
 	sudo install -D -m 644 misc/snapshotter/nydus-snapshotter.${FS_DRIVER}.service /etc/systemd/system/nydus-snapshotter.service
-	sudo systemctl enable /etc/systemd/system/nydus-snapshotter.service
+
+	@if which systemctl; then sudo systemctl enable /etc/systemd/system/nydus-snapshotter.service;fi
```
imeoer, line 43: If no systemd is installed, any error should be thrown?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: "1.18"
```
imeoer, line 24: go 1.17 has been defined in `go.mod`.

integration/entrypoint.sh
```diff
@@ -0,0 +1,208 @@
...
+        sleep ${RETRYINTERVAL}
+    done
+    if [ "${SUCCESS}" == "true" ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+function reboot_containerd {
```
imeoer, line 80: Maybe name it `cleanup` or something?

Makefile
changweige: no need

.github/workflows/e2e.yml
changweige: What's its effect? I didn't see any error when running the tests

integration/entrypoint.sh
changweige: No. It is truly restring containerd

Makefile
changweige: The integration test needs this Makefile target and the base container image has no systemd packed. The user should know if he/she has systemd installed as a system administrator.

.github/workflows/e2e.yml
changweige: Fine. changed to 1.17.y As containerd has increased its dependent golang version to 1.18.y.
We should consider upgrading it.  Then we won't have two kinds of build tags like:
```
//go:build !windows
// +build !windows
```

imeoer (approved):
LGTM

PR #143, opened 2022-08-24
change default work root dir to /var/lib/containerd-nydus

changweige:
Thus to keep consistent with /run/containerd-nydus

### DIFF
--- a/cmd/containerd-nydus-grpc/pkg/command/flags.go
+++ b/cmd/containerd-nydus-grpc/pkg/command/flags.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/containerd/containerd/log"
 	"github.com/containerd/nydus-snapshotter/cmd/containerd-nydus-grpc/pkg/logging"
 	"github.com/containerd/nydus-snapshotter/config"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


liubogithub:
Hmm, if one had deployed an old version of nydus-snapshotter, after this upgrade, his setup won't work any more.

Can we leave some error messages to let users know that it's probably due to the change of root dir?

changweige:
> Hmm, if one had deployed an old version of nydus-snapshotter, after this upgrade, his setup won't work anymore.
> 
> Can we leave some error messages to let users know that it's probably due to the change of root dir?

Reasonable. Fixed.

cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -263,6 +264,14 @@ func Validate(args *Args, cfg *config.Config) error {
 		return errors.New("invalid empty root directory")
 	}
 
+	if args.RootDir == defaultRootDir {
+		if entries, err := os.ReadDir(oldDefaultRootDir); err == nil {
```
imeoer, line 268: Old users will confuse if the error `default root directory is changed to /var/lib/containerd-nydus` be thrown and snapshotter exits, just give a warning for that?

changweige: Changed to a warning message

imeoer (approved):

Issue #142, opened 2022-08-23
Duplicated from nydusd repo: The container does not start itself with the host boot-up


changweige:
https://github.com/dragonflyoss/image-service/issues/691

PR #145, opened 2022-08-25
clean up nydus vestige when starting

changweige:
Nydusd judges if it should enter failover phrase by
checking residual disconnected unix socket file and mounted
mountpoint. But if snapshotter is restarting, nydusd can't fetch
back its states. So just start a fresh nydusd

Addressing #142 

### DIFF
--- a/pkg/process/manager.go
+++ b/pkg/process/manager.go
@@ -270,7 +270,7 @@ func (m *Manager) IsPrefetchDaemon() bool {
 	return m.daemonMode == config.DaemonModePrefetch
 }
 
-// Reconnect already running daemons，and rebuild daemons management structs.
+// Reconnect running daemons and rebuild daemons management states
 func (m *Manager) Reconnect(ctx context.Context) error {
 	var (
 		daemons      []*daemon.Daemon
@@ -296,6 +296,22 @@ func (m *Manager) Reconnect(ctx context.Context) error {
 		info, err := d.CheckStatus()
 		if err != nil {
 			log.L.WithField("daemon", d.ID).Warnf("failed to check daemon status: %v", err)
+
+			if d.ID == daemon.SharedNydusDaemonID {
+				// The only reason that nydusd can't be connected is it's not running.
+				// Moreover, snapshotter is restarting. So no nydusd states can be returned to each nydusd.
+				// Nyudsd can't do failover any more.
+				// We can safely try to umount its mountpoint to avoid nydusd pausing in INIT state.
+				log.L.Warnf("Nydusd died somehow. Clean up its vestige!")
+
+				mounter := mount.Mounter{}
+				// This is best effort. So no need to handle its error.
+				mounter.Umount(*d.RootMountPoint)
+				// Nydusd judges if it should enter failover phrase by checking if unix sockt is existed and it
+				// can't be connected.
+				os.Remove(d.GetAPISock())
+			}
+
 			return nil
 		}
 		if !info.Running() {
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -69,7 +69,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        GOOS: ['linux', 'windows', 'darwin']
+        GOOS: ['linux', 'darwin']
```
imeoer, line 72: We need to ensure the snapshooter can be compiled on windows platform, so that it can be imported by other components (buildkit/nerdctl).

pkg/process/manager.go
```diff
@@ -296,6 +297,20 @@ func (m *Manager) Reconnect(ctx context.Context) error {
 		info, err := d.CheckStatus()
 		if err != nil {
 			log.L.WithField("daemon", d.ID).Warnf("failed to check daemon status: %v", err)
+
+			if d.ID == daemon.SharedNydusDaemonID {
+				// The only season that nydusd can't be connected is it's not running.
```
imeoer, line 301: typo `reason`.

pkg/process/manager.go
```diff
@@ -296,6 +297,20 @@ func (m *Manager) Reconnect(ctx context.Context) error {
 		info, err := d.CheckStatus()
 		if err != nil {
 			log.L.WithField("daemon", d.ID).Warnf("failed to check daemon status: %v", err)
+
+			if d.ID == daemon.SharedNydusDaemonID {
```
imeoer, line 300: Why only handle the shared daemon case?

.github/workflows/ci.yml
changweige: Windows compilation is back

pkg/process/manager.go
```diff
@@ -296,6 +296,22 @@ func (m *Manager) Reconnect(ctx context.Context) error {
...
+				// The only season that nydusd can't be connected is it's not running.
+				// Moreover, snapshotter is restarting. So no nydusd states can be returned to each nydusd.
+				// Nyudsd can't do failover any more.
+				// We can safely try to umount its mountpoint to avoid nydusd pausing in INIT state.
+				log.L.Infof("Nydusd died somehow. Clean up its vestige.")
```
imeoer, line 305: `Nydusd in shared mode died somehow, clean up its vestige.`, and maybe use warning level?

pkg/process/manager.go
changweige: In fact, multiple daemons mod also has this issue. A following PR is going to resolve it

changweige: > Why only handle the shared daemon case?

In fact, multiple daemons mod also has this issue. A following PR is going to resolve it

pkg/process/manager.go
```diff
@@ -296,6 +296,22 @@ func (m *Manager) Reconnect(ctx context.Context) error {
...
+			if d.ID == daemon.SharedNydusDaemonID {
+				// The only reason that nydusd can't be connected is it's not running.
+				// Moreover, snapshotter is restarting. So no nydusd states can be returned to each nydusd.
+				// Nyudsd can't do failover any more.
+				// We can safely try to umount its mountpoint to avoid nydusd pausing in INIT state.
+				log.L.Warnf("Nydusd died somehow. Clean up its vestige!")
+
+				mounter := mount.Mounter{}
+				// This is best effort. So no need to handle its error.
+				mounter.Umount(*d.RootMountPoint)
```
imeoer, line 309: We'd better not let any errors go, at least give a warning log.

changweige: No need to give a warning. It does not have side effects.  It can upset the end user.

PR #147, opened 2022-08-26
converter: create work directory if not exists

imeoer:
In case users forget to create the work directory before using converter.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -41,7 +41,7 @@ const bootstrapNameInTar = "image.boot"
 const blobNameInTar = "image.blob"
 
 const envNydusBuilder = "NYDUS_BUILDER"
-const envNydusWorkdir = "NYDUS_WORKDIR"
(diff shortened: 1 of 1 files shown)


pkg/converter/convert_unix.go
```diff
@@ -58,17 +58,28 @@ func getBuilder(specifiedPath string) string {
...
-func getWorkdir(specifiedPath string) string {
-	if specifiedPath != "" {
-		return specifiedPath
+func ensureWorkDir(specifiedBasePath string) (string, error) {
+	var baseWorkDir string
+
+	if specifiedBasePath != "" {
+		baseWorkDir = specifiedBasePath
+	} else {
+		baseWorkDir = os.Getenv(envNydusWorkdir)
```
changweige, line 67: Can we also name it as envNydusWorkDir like L62

imeoer: Fixed

changweige (approved):
LGTM

PR #148, opened 2022-08-26
clean up nydus vestige when starting

changweige:
Nydusd judges if it should enter failover phrase by
checking residual disconnected unix socket file and mounted
mountpoint. But if snapshotter is restarting, nydusd can't fetch
back its states. So just start a fresh nydusd


imeoer (approved):
LGTM!

Issue #128, opened 2022-08-11
Improve converter performance by using bufio


imeoer:
We can wrap the [dest writer](https://github.com/containerd/nydus-snapshotter/blob/main/pkg/converter/convert_unix.go#L238) with [bufio](https://pkg.go.dev/bufio) to improve the performance on writing blob to containerd content store.

sctb512:
Is it necessary to wrap all "io.Reader" and "io.Writer" in bufio?

imeoer:
> Is it necessary to wrap all "io.Reader" and "io.Writer" in bufio?

Seem it's ok, maybe we can use https://pkg.go.dev/io#CopyBuffer directly.

sctb512:
> > Is it necessary to wrap all "io.Reader" and "io.Writer" in bufio?
> 
> Seem it's ok, maybe we can use https://pkg.go.dev/io#CopyBuffer directly.

Fine, I get it.

PR #150, opened 2022-08-26
converter: add buffer for io.Writer

sctb512:
add buffer for io.Writer to improve
nydus image conversion performation.

This PR is addressing #128 

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sync"
 	"syscall"
 
 	"github.com/containerd/containerd/archive"
@@ -44,6 +45,13 @@ const envNydusBuilder = "NYDUS_BUILDER"
 const envNydusWorkdir = "NYDUS_WORKDIR"
 
 const configGCLabelKey = "containerd.io/gc.ref.content.config"
+
+var bufPool = sync.Pool{
+	New: func() interface{} {
+		buffer := make([]byte, 1<<20)
+		return &buffer
+	},
+}
 
 func getBuilder(specifiedPath string) string {
 	if specifiedPath != "" {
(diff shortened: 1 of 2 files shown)


pkg/converter/convert_unix.go
```diff
@@ -304,7 +306,8 @@ func Pack(ctx context.Context, dest io.Writer, opt PackOption) (io.WriteCloser,
 			}
 		}()
 
-		if _, err := io.Copy(dest, blobFifo); err != nil {
+		buffer := make([]byte, defaultBufferSize)
```
imeoer, line 309: Maybe we can use [sync.Pool](https://github.com/containerd/containerd/blob/765351ac4def77c5c18e808ede3e71c99266fc6f/content/local/store.go#L514) to avoid allocating new buffer repeatedly, and do you have any performance testing, how about 1MB buffer?

pkg/converter/convert_unix.go
```diff
@@ -378,7 +381,8 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
 	}
 	defer rc.Close()
 
-	if _, err = io.Copy(dest, rc); err != nil {
```
imeoer, line 381: How about replacing all `io.Copy` to `io.CopyBuffer` in `convert_unix.go`?

pkg/converter/convert_unix.go
```diff
@@ -304,9 +312,11 @@ func Pack(ctx context.Context, dest io.Writer, opt PackOption) (io.WriteCloser,
 			}
 		}()
 
-		if _, err := io.Copy(dest, blobFifo); err != nil {
+		buffer := bufPool.Get().(*[]byte)
+		if _, err := io.CopyBuffer(dest, blobFifo, *buffer); err != nil {
```
imeoer, line 316: Better to use `defer bufPool.Put(buffer)` after `buffer := bufPool.Get().(*[]byte)`.

imeoer (approved):

PR #151, opened 2022-08-29
converter: fix incorrect blob list from merge

imeoer:
Affected by chunk dict, the blob list referenced by final bootstrap
are from different layers, part of them are from original layers,
part from chunk dict bootstrap.

So we need to rewrite manifest's layers using the real blob list
referenced by final bootstrap.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -341,11 +341,14 @@ func Pack(ctx context.Context, dest io.Writer, opt PackOption) (io.WriteCloser, 
 	return wc, nil
 }
 
-// Merge multiple nydus bootstraps (from each blob layer of image) to a final bootstrap.
-func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption) error {
+// Merge multiple nydus bootstraps (from each layer of image) to a final
+// bootstrap. And due to the possibility of enabling the `ChunkDictPath`
+// option causes the data deduplication, it will return the actual blob
+// digests referenced by the bootstrap.
+func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption) ([]digest.Digest, error) {
 	workDir, err := ensureWorkDir(opt.WorkDir)
 	if err != nil {
-		return errors.Wrap(err, "ensure work directory")
+		return nil, errors.Wrap(err, "ensure work directory")
 	}
 	defer os.RemoveAll(workDir)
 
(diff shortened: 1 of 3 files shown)


pkg/converter/convert_unix.go
```diff
@@ -712,41 +726,61 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 	buffer := bufPool.Get().(*[]byte)
 	defer bufPool.Put(buffer)
 	if _, err := io.CopyBuffer(compressed, pr, *buffer); err != nil {
-		return nil, errors.Wrapf(err, "copy bootstrap targz into content store")
+		return nil, nil, errors.Wrapf(err, "copy bootstrap targz into content store")
 	}
 	if err := gw.Close(); err != nil {
-		return nil, errors.Wrap(err, "close gzip writer")
+		return nil, nil, errors.Wrap(err, "close gzip writer")
```
changweige, line 732: Not related to this PR. But `gw` can be nil. We should handle the exception.
In addition, `gw` should also be closed when returning from L729

pkg/converter/convert_unix.go
```diff
@@ -712,41 +726,61 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
...
+	bootstrapInfo, err := cs.Info(ctx, compressedDgst)
 	if err != nil {
-		return nil, errors.Wrap(err, "get info from content store")
+		return nil, nil, errors.Wrap(err, "get info from content store")
+	}
+
+	blobDigests := <-blobDigestChan
+	blobDescs := []ocispec.Descriptor{}
+	blobIDs := []string{}
+	for _, blobDigest := range blobDigests {
```
changweige, line 755: If `blobDigests` is empty or nil, which means another processing goroutine fails . Should it abort the following handle?

pkg/converter/convert_unix.go
```diff
@@ -621,24 +626,32 @@ func ConvertHookFunc(opt MergeOption) converter.ConvertHookFunc {
...
 			manifest.Layers = []ocispec.Descriptor{*bootstrapDesc}
 		} else {
-			manifest.Layers = append(manifest.Layers, *bootstrapDesc)
+			for idx, blobDesc := range blobDescs {
+				blobGCLabelKey := fmt.Sprintf("containerd.io/gc.ref.content.l.%d", idx)
+				labels[blobGCLabelKey] = blobDesc.Digest.String()
+			}
+			// Affected by chunk dict, the blob list referenced by final bootstrap
+			// are from different layers, part of them are from original layers, part
+			// from chunk dict bootstrap, so we need to rewrite manifest's layers here.
```
changweige, line 635: I think overwrite is better than rewrite

changweige (commented):
Better we can add an e2e test case to the integration test.

pkg/converter/convert_unix.go
imeoer: image

It's ok if `gw.Close` not be called when returning from L729.

pkg/converter/convert_unix.go
imeoer: Will abort at L728: `if _, err := io.CopyBuffer(compressed, pr, *buffer); err != nil {`

pkg/converter/convert_unix.go
imeoer: It means just generating a new manifest, not covering it. :)

pkg/converter/convert_unix.go
changweige: Then where does it flush the data?

imeoer: If an error happens, no data need to be flushed.

changweige (approved):
LGTM

Issue #125, opened 2022-08-06
can not start a rafs v6 container in fuse mode with --shared-daemon option


changweige:
Duplicate one here from https://github.com/dragonflyoss/image-service/issues/638

PR #152, opened 2022-08-30
Enhance daemon states management to improve usability and stability

changweige:
Several community users and nydus contributors have complained about nydus-snapshotter's usability and stability.
1. nydusd is not ready so the container image can't be accessed
2. nydus image can't be accessed after rebooting the host

To solve the issues once and for all, this PR refactors nydusd daemons states management. It decouples in-memory states cache from DB persistence storage. Snapshotter never empties the DB when restarting but to rebuild the in-memory daemon states cache to get consistent with DB persistent storage. In addition, it will request API to restarted nydusd daemon to recover the API mounts.

Several E2E cases are accompanied.

Address: https://github.com/containerd/nydus-snapshotter/issues/142 https://github.com/containerd/nydus-snapshotter/issues/125

### DIFF
--- a/pkg/filesystem/fs/fs.go
+++ b/pkg/filesystem/fs/fs.go
@@ -148,25 +148,32 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
 	if recoveringDaemons, err = fs.manager.Reconnect(ctx); err != nil {
 		return nil, errors.Wrap(err, "failed to reconnect daemons")
 	}
+
+	var sharedDaemonConnected = false
+
 	// Try to bring up the shared daemon early.
 	if fs.mode == SharedInstance {
-		if d, _ := fs.getSharedDaemon(); d != nil {
+		if d, _ := fs.getSharedDaemon(); d != nil && !d.Connected {
 			log.G(ctx).Infof("initialize the shared nydus daemon")
 			fs.initSharedDaemon(ctx)
+
+			sharedDaemonConnected = true
 		}
 	}
 
-	// Try to bring all persisted and stopped nydusd up and remount Rafs
+	// Try to bring all persisted and stopped nydusds up and remount Rafs
 	if len(recoveringDaemons) != 0 {
 		for _, d := range recoveringDaemons {
 			if fs.mode == SharedInstance {
-				if d.ID != daemon.SharedNydusDaemonID {
+				if d.ID != daemon.SharedNydusDaemonID && sharedDaemonConnected {
 					if err := d.SharedMount(); err != nil {
 						return nil, errors.Errorf("failed to mount rafs when recovering, %v", err)
 					}
 				}
 			} else {
-				fs.manager.StartDaemon(d)
+				if !d.Connected {
+					fs.manager.StartDaemon(d)
+				}
 			}
 		}
 	}
@@ -700,9 +707,12 @@ func (fs *Filesystem) initSharedDaemon(ctx context.Context) (_ *daemon.Daemon, r
 
 	defer func() {
 		if retErr != nil {
-			fs.manager.DeleteDaemon(d)
+			if err := fs.manager.DeleteDaemon(d); err != nil {
+				log.L.Errorf("%v", err)
+			}
 		}
 	}()
+
 	if err := fs.manager.StartDaemon(d); err != nil {
 		return nil, errors.Wrap(err, "failed to start shared daemon")
 	}
--- a/pkg/filesystem/fs/fs.go
+++ b/pkg/filesystem/fs/fs.go
@@ -153,9 +153,11 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
 
 	// Try to bring up the shared daemon early.
 	if fs.mode == SharedInstance {
-		if d, _ := fs.getSharedDaemon(); d != nil && !d.Connected {
-			log.G(ctx).Infof("initialize the shared nydus daemon")
-			fs.initSharedDaemon(ctx)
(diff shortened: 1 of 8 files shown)


pkg/process/manager.go
```diff
@@ -26,14 +26,164 @@ import (
...
+		}
+	} else {
+		s.daemons = append(s.daemons, daemon)
+	}
+
+	s.idxByDaemonID[daemon.ID] = daemon
+	s.idxBySnapshotID[daemon.SnapshotID] = daemon
+
+	if ok {
+		return old
```
imeoer, line 67: Why do we need to return `old`? seems it doesn't to be used.

pkg/process/manager.go
```diff
@@ -26,14 +26,164 @@ import (
...
+
+	var deleted *daemon.Daemon
+
+	ds := s.daemons[:0]
+	for _, victim := range s.daemons {
+		if victim == d {
+			deleted = victim
+			continue
+		}
+		ds = append(ds, victim)
```
imeoer, line 85: Why is the victim added to the survivor list? :)

pkg/process/manager.go
```diff
@@ -26,14 +26,164 @@ import (
...
+
+func (s *DaemonStates) Remove(d *daemon.Daemon) *daemon.Daemon {
+	s.mu.Lock()
+	old := s.removeUnlocked(d)
+	s.mu.Unlock()
+
+	return old
+}
+
+func (s *DaemonStates) RemoveByDaemonID(id string) *daemon.Daemon {
```
imeoer, line 101: Seem it not to be used.

pkg/process/manager.go
```diff
@@ -26,14 +26,164 @@ import (
...
+	if daemon != nil && op != nil {
+		op(s, daemon)
+	} else if daemon == nil {
+		log.L.Warnf("daemon daemon_id=%s is not found", id)
+	}
+
+	return daemon
+}
+
+func (s *DaemonStates) GetBySnapshotID(id string, op func(s *DaemonStates, d *daemon.Daemon)) *daemon.Daemon {
```
imeoer, line 134: Is it necessary to put `s` as a parameter of the callback func?

pkg/process/manager.go
```diff
@@ -26,14 +26,164 @@ import (
...
+func (s *DaemonStates) List() []*daemon.Daemon {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if len(s.daemons) == 0 {
+		return nil
+	}
+
+	listed := make([]*daemon.Daemon, len(s.daemons))
+	copy(listed, s.daemons)
```
imeoer, line 158: Is it necessary to copy one?

pkg/process/manager.go
changweige: Yes. This is just a hint to callers. It is up to the caller if they need to check the daemon is previously inserted.
By returning  old daemon pointer, caller has a chance to tell it is overwitten.

pkg/process/manager.go
changweige: The matching victim is already skipped. Please check it out on L83.

pkg/process/manager.go
changweige: Yes. It is the original semantic that daemon can be removed by its snapshot ID. So I transformed it a bit.

pkg/process/manager.go
changweige: Yes. For current usage, the callback is to delete an item from daemons states cache. So it has to access the s as daemon states manager

pkg/process/manager.go
changweige: Yes. It is protected by the mutex lock. If not copying the slice, we can't say the returned slice is persistent.

pkg/filesystem/fs/fs.go
```diff
@@ -138,17 +140,44 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
...
+
+	var sharedDaemonConnected = false
+
 	// Try to bring up the shared daemon early.
 	if fs.mode == SharedInstance {
-		if _, e := fs.getSharedDaemon(); e != nil {
-			log.G(ctx).Warnf("initialize the shared nydus daemon")
+		if d := fs.getSharedDaemon(); d != nil && !d.Connected {
+			log.G(ctx).Infof("initialize the shared nydus daemon")
 			fs.initSharedDaemon(ctx)
```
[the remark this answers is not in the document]
changweige: fixed

pkg/filesystem/fs/fs.go
```diff
@@ -138,17 +140,44 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
...
+		}
+	}
+
+	// Try to bring all persisted and stopped nydusds up and remount Rafs
+	if len(recoveringDaemons) != 0 {
+		for _, d := range recoveringDaemons {
+			if fs.mode == SharedInstance {
+				if d.ID != daemon.SharedNydusDaemonID && sharedDaemonConnected {
+					if err := d.SharedMount(); err != nil {
+						return nil, errors.Errorf("failed to mount rafs when recovering, %v", err)
```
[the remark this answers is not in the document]
changweige: fixed

pkg/filesystem/fs/fs.go
```diff
@@ -138,17 +140,44 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
...
+					}
+				}
+			} else {
+				if !d.Connected {
+					fs.manager.StartDaemon(d)
```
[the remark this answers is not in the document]
changweige: fixed

imeoer (approved):

PR #153, opened 2022-08-31
add e2e test for fscache driver

changweige:


imeoer:
cc @hsiangkao

integration/entrypoint.sh
```diff
@@ -91,14 +96,15 @@ function reboot_containerd {
     fi
 
     local daemon_mode=${1}
+    local fs_driver=${2:-fuse}
```
imeoer, line 99: Should be `fusedev` instead of `fuse` ?

Makefile
```diff
@@ -67,4 +68,6 @@ smoke:
 integration:
 	CGO_ENABLED=1 ${PROXY} GOOS=${GOOS} GOARCH=${GOARCH} go build -ldflags '-s -w -X "main.Version=${VERSION}" -extldflags "-static"' -race -v -o bin/containerd-nydus-grpc ./cmd/containerd-nydus-grpc
 	$(SUDO) DOCKER_BUILDKIT=1 docker build -t nydus-snapshotter-e2e:0.1 -f integration/Dockerfile .
-	$(SUDO) docker run --privileged -v /root/.docker:/root/.docker -v `go env GOMODCACHE`:/go/pkg/mod -v `go env GOCACHE`:/root/.cache/go-build -v `pwd`:/nydus-snapshotter nydus-snapshotter-e2e:0.1
\ No newline at end of file
+	$(SUDO) docker run --rm --privileged -v /root/.docker:/root/.docker -v `go env GOMODCACHE`:/go/pkg/mod \
+	-v `go env GOCACHE`:/root/.cache/go-build -v `pwd`:/nydus-snapshotter \
+	-v /usr/src/linux-headers-${KERNEL_VER}:/usr/src/linux-headers-${KERNEL_VER} nydus-snapshotter-e2e:0.1
```
hsiangkao, line 73: why do we need kernel headers here?

integration/entrypoint.sh
```diff
@@ -77,6 +77,11 @@ function retry {
     fi
 }
 
+function can_erofs_ondemand_read {
+    grep CONFIG_EROFS_FS_ONDEMAND=y /usr/src/linux-headers-"$(uname -r)"/.config
```
imeoer, line 81: Can this test case run in github action? or just run in an internal host?

hsiangkao, line 81: oh, okay, I think we can just check whether /dev/cachefiles exists

hsiangkao (commented):
> cc @hsiangkao

Yeah, yet I think it needs the proper kernel first (>= 5.19)

Makefile
changweige: Because the test has to detect if the kernel is supporting erofs on-demand read

integration/entrypoint.sh
changweige: I am afraid it's not enough since `/dev/cachefiles ` can still work for netfs on-demand read

integration/entrypoint.sh
changweige: Yes. it's bug here

integration/entrypoint.sh
changweige: Github action does not have a runner host with the kernel of that new (>=5.19). So it only runs locally

integration/entrypoint.sh
changweige: fixed.

integration/entrypoint.sh
hsiangkao: `CONFIG_EROFS_FS_ONDEMAND=m` is also valid.

changweige: Fixed. Thanks

imeoer (approved):

PR #154, opened 2022-08-31
converter: improve performance by calculating diffid ahead of time

imeoer:
Write a diff id label of layer in content store for simplifying
diff id calculation to speed up the conversion.
See: https://github.com/containerd/containerd/blob/e4fefea5544d259177abb85b64e428702ac49c97/images/diffid.go#L49

Related performance test: https://github.com/goharbor/acceleration-service/pull/50


changweige (approved):
LGTM

PR #155, opened 2022-08-31
converter: improve method options

imeoer:
Update some comments for exported methods in the same time.


pkg/converter/convert_unix.go
```diff
@@ -684,13 +690,17 @@ func mergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 		return nil, errors.Wrap(err, "marshal blob ids")
 	}
 
+	if opt.FsVersion == "" {
+		opt.FsVersion = "5"
```
changweige, line 694: Should we move to fs version 6 as default?  Fs version 5 is really a burden for development so far.

pkg/converter/types.go
```diff
@@ -58,12 +59,18 @@ type MergeOption struct {
 	WorkDir string
 	// BuilderPath holds the path of `nydus-image` binary tool.
 	BuilderPath string
+	// FsVersion specifies nydus RAFS format version, possible
+	// values: `5`, `6` (EROFS-compatible), default is `5`.
```
changweige, line 63: I am suggesting using 6 as default

changweige (commented):
Otherwise looks good

pkg/converter/convert_unix.go
imeoer: We have left `v5` as the default option in all components (nydusify/buildkit/nerdctl/acceld) before, maybe we can push `v6` as default together in the future.

PR #159, opened 2022-09-05
converter: export IsNydusBlob and MergeLayers methods

imeoer:
Make them easier for buildkit/nerdctl/acceld components to use.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -450,6 +450,27 @@ func Unpack(ctx context.Context, ra content.ReaderAt, dest io.Writer, opt Unpack
 	return nil
 }
 
+// IsNydusBlobAndExists returns true when the specified digest of content exists in
+// the content store and it's nydus blob format.
+func IsNydusBlobAndExists(ctx context.Context, cs content.Store, desc ocispec.Descriptor) bool {
+	_, err := cs.Info(ctx, desc.Digest)
+	if err != nil {
+		return false
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/converter/convert_unix.go
```diff
@@ -450,6 +450,23 @@ func Unpack(ctx context.Context, ra content.ReaderAt, dest io.Writer, opt Unpack
...
+		return false
+	}
+
+	_, err := cs.Info(ctx, desc.Digest)
+	if err != nil {
+		return false
+	}
+
+	hasMediaType := desc.MediaType == MediaTypeNydusBlob
+	_, hasAnno := desc.Annotations[LayerAnnotationNydusBlob]
```
changweige, line 466: Why is only nydus blob media type not enough to judge it's a nydus blob layer?

imeoer: Maybe we could just check by media type, but recently, we won't assume that the media type must be `"application/vnd.oci.image.layer.nydus.blob.v1"`, maybe a tgz media type can be supported in future? So we should always trust the annotation.

changweige: Then how about not checking media type but just checking annotation? Otherwise, this method malfunctions when "application/vnd.oci.image.layer.nydus.blob.v1" is abandoned.

imeoer: Okay, it makes sense, fixed.

changweige (approved):
LGTM

PR #160, opened 2022-09-05
give NoneMode mount option with lowerdir=/dummy

changweige:
NoneMode nydus-snapshotter does not start nydusd and
no container image mounted contents are provided. Only
info in mount options are meaningful. So give a dummy
mount dir there.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -482,12 +482,12 @@ func (o *snapshotter) upperPath(id string) string {
 	return filepath.Join(o.root, "snapshots", id, "fs")
 }
 
-func (o *snapshotter) lowerPath(id string) string {
-	if mnt, err := o.fs.MountPoint(id); err == nil {
-		return mnt
+func (o *snapshotter) lowerPath(id string) (mnt string, err error) {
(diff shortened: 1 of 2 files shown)


snapshot/snapshot.go
```diff
@@ -608,7 +608,12 @@ func (o *snapshotter) remoteMounts(ctx context.Context, s storage.Snapshot, id s
 		return bindMount(o.upperPath(s.ParentIDs[0])), nil
 	}
 
-	lowerDirOption := fmt.Sprintf("lowerdir=%s", o.lowerPath(id))
+	lowerPath := o.lowerPath(id)
+	if lowerPath == "" {
```
imeoer, line 612: Maybe not related to this PR, just a code style suggestion, the error returned from `o.fs.MountPoint` is ignored in `o.lowerPath`, maybe we should return the error from `o.lowerPath` directly.

imeoer (approved):

snapshot/snapshot.go
changweige: Yes. Let's return the error to containerd

PR #162, opened 2022-09-07
Support shared domain

userzj:
https://lore.kernel.org/linux-erofs/20220902105305.79687-1-zhujia.zj@bytedance.com/T/#t
Signed-off-by: Jia Zhu <zhujia.zj@bytedance.com>

### DIFF
--- a/config/daemonconfig.go
+++ b/config/daemonconfig.go
@@ -167,7 +167,7 @@ func NewDaemonConfig(fsDriver string, cfg DaemonConfig, imageID, snapshotID stri
 			backendConfig = &cfg.Config.BackendConfig
 			fscacheID := erofs.FscacheID(snapshotID)
 			cfg.ID = fscacheID
-			cfg.DomainID = fscacheID
+			cfg.DomainID = erofs.GlobalDomainName
 			cfg.Config.ID = fscacheID
 		}
 		if keyChain != nil {
--- a/pkg/utils/erofs/erofs_linux.go
+++ b/pkg/utils/erofs/erofs_linux.go
@@ -12,12 +12,23 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const GlobalDomainName = "global-domain"
+
 func Mount(bootstrapPath, fsID, mountPoint string) error {
 	mount := unix.Mount
 
-	opts := "fsid=" + fsID
+	opts := fmt.Sprintf("domain_id=%s,fsid=%s", GlobalDomainName, fsID)
 	logrus.Infof("Mount erofs to %s with options %s", mountPoint, opts)
+
+	// If a mount with the option 'domain_id' fails, this indicates that
+	// the kernel may not support EROFS shared domain feature.
+	// Fallback to retry the mount without 'domain_id' option.
 	if err := mount("erofs", mountPoint, "erofs", 0, opts); err != nil {
+		if errors.Is(err, unix.EINVAL) {
+			logrus.Infof("Fallback to the mount without @domain_id option")
+			opts := "fsid=" + fsID
+			err = mount("erofs", mountPoint, "erofs", 0, opts)
+		}
 		return errors.Wrapf(err, "failed to mount erofs")
 	}
 
--- a/pkg/utils/erofs/erofs_other.go
+++ b/pkg/utils/erofs/erofs_other.go
@@ -5,6 +5,8 @@ package erofs
 
 import "fmt"
 
+const GlobalDomainName = ""
+
 func Mount(bootstrapPath, fsID, mountPoint string) error {
 	return fmt.Errorf("not implemented")
 }


hsiangkao:
``
golangci-lint run
pkg/utils/erofs/erofs_linux.go:21: File is not `gofmt`-ed with `-s` (gofmt)
		if err := mount("erofs", mountPoint, "erofs", 0, "fsid=" + fsID); err != nil {
``

config/daemonconfig.go
```diff
@@ -161,7 +161,7 @@ func NewDaemonConfig(fsDriver string, cfg DaemonConfig, imageID, snapshotID stri
 			backendConfig = &cfg.Config.BackendConfig
 			fscacheID := erofs.FscacheID(snapshotID)
 			cfg.ID = fscacheID
-			cfg.DomainID = fscacheID
+			cfg.DomainID = "global-domain"
```
changweige, line 164: Is domain ID specific to kernel/erofs? In other words, is it a predefined keyword in the Kernel?

pkg/utils/erofs/erofs_linux.go
```diff
@@ -15,10 +15,12 @@ import (
 func Mount(bootstrapPath, fsID, mountPoint string) error {
 	mount := unix.Mount
 
-	opts := "fsid=" + fsID
+	opts := fmt.Sprintf("domain_id=%s,fsid=%s", "global-domain", fsID)
 	logrus.Infof("Mount erofs to %s with options %s", mountPoint, opts)
 	if err := mount("erofs", mountPoint, "erofs", 0, opts); err != nil {
-		return errors.Wrapf(err, "failed to mount erofs")
+		if err := mount("erofs", mountPoint, "erofs", 0, "fsid="+fsID); err != nil {
+			return errors.Wrapf(err, "failed to mount erofs")
```
changweige, line 22: Why do you retry another mount if the L20 mount operation fails? It's better to add a comment to tell necessity and intention.

pkg/utils/erofs/erofs_linux.go
```diff
@@ -15,10 +15,12 @@ import (
 func Mount(bootstrapPath, fsID, mountPoint string) error {
 	mount := unix.Mount
 
-	opts := "fsid=" + fsID
+	opts := fmt.Sprintf("domain_id=%s,fsid=%s", "global-domain", fsID)
```
changweige, line 18: It's better to make "global-domain" as a global constant

config/daemonconfig.go
userzj: Yup, @domain_id is a mount option of EROFS.

pkg/utils/erofs/erofs_linux.go
userzj: Thanks, I'll revise it.

pkg/utils/erofs/erofs_linux.go
userzj: If a mount with the option "domain_id" fails, this indicates that the kernel may not support EROFS shared domain, and will fall back to normal mount. I'll add a comment about that. Thanks.

config/daemonconfig.go
changweige: Actually, I mean "global-domain" :-)

pkg/utils/erofs/erofs_linux.go
changweige: I see. Can it check return error code like unsupported or invalid ?

config/daemonconfig.go
userzj: 'global-domain' is not a keyword in the kernel. This is just a name(id) specified by us. In future work, 'domain_id' will be specified by the user.

pkg/utils/erofs/erofs_linux.go
```diff
@@ -12,15 +12,24 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const globalDomainName = "global-domain"
```
changweige, line 15: Um. erofs package should import the constant from daemonconfig.go

pkg/utils/erofs/erofs_linux.go
```diff
@@ -12,15 +12,24 @@ import (
...
+
 func Mount(bootstrapPath, fsID, mountPoint string) error {
 	mount := unix.Mount
+	var err error
 
-	opts := "fsid=" + fsID
+	opts := fmt.Sprintf("domain_id=%s,fsid=%s", globalDomainName, fsID)
 	logrus.Infof("Mount erofs to %s with options %s", mountPoint, opts)
-	if err := mount("erofs", mountPoint, "erofs", 0, opts); err != nil {
+	err = mount("erofs", mountPoint, "erofs", 0, opts)
```
changweige, line 23: We still can improve the code like:
```
if err := mount("erofs", mountPoint, "erofs", 0, opts); err!=nil {
	if errors.Is(err, unix.EINVAL) {
		err = mount("erofs", mountPoint, "erofs", 0, "fsid="+fsID)
                 if err != nil {
                      return errr
}
	}

         return errors.Wrapf(err, "failed to mount erofs")
}

```

userzj: Thanks.

changweige (approved):
LGTM, thanks

PR #163, opened 2022-09-08
Remove default http timeout limit

zyfjeff:
When the single-layer image is too large or the network bandwidth is poor, it is easy to reach this timeout period. The default resolver of docker has no timeout limit, so this timeout limit is removed here, and the timeout is controlled by upper-layer tools such as nerdctl/ctr/crictl.

Signed-off-by: zyfjeff <zyfjeff@linux.alibaba-inc.com>

### DIFF
--- a/pkg/filesystem/fs/resolver.go
+++ b/pkg/filesystem/fs/resolver.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"time"
 
 	"github.com/containerd/containerd/reference/docker"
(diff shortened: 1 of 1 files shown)


changweige:
How can the upper layer tool you metioned handle timeout? Is there a CLI parameter?

zyfjeff:
> How can the upper layer tool you metioned handle timeout? Is there a CLI parameter?

sure, ctr has timeout parameter,
   --timeout value              total timeout for ctr commands (default: 0s)

and, nerdctl、circtl also has timeout parameter

changweige (approved):
LGTM

PR #170, opened 2022-09-13
[Bugfix] Modify the logic of obtaining image ref when deleting snapshots

power-more:
In some cases, such as a user-defined label in the upper layer, the snapshot label is not empty, but does not contain
"containerd.io/snapshot/cri.image-ref", the function should continue to execute normally instead of returning an error.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -402,13 +402,18 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
 		return errors.Wrap(err, "failed to get snapshot")
 	}
 
+	// workaround for cacheMgr
+	// It is not necessary to execute successfully, even if it fails, it does not affect removing snapshot
 	if snap.Labels != nil {
-		if imageID, ok := snap.Labels[label.CRIImageRef]; ok {
-			if err := o.fs.DelSnapshot(imageID); err != nil {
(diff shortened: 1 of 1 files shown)


snapshot/snapshot.go
```diff
@@ -403,12 +403,14 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
 	}
 
 	if snap.Labels != nil {
-		if imageID, ok := snap.Labels[label.CRIImageRef]; ok {
-			if err := o.fs.DelSnapshot(imageID); err != nil {
-				return errors.Wrap(err, "failed to delete snapshot in cache manager")
+		if _, ok := snap.Labels[label.TargetSnapshotRef]; ok {
+			if imageID, ok := snap.Labels[label.CRIImageRef]; ok {
```
changweige, line 407: Why need to check if a particular label existed for the deleted snapshot?  What's the problem if we delete a snapshot which does not meet the labels combination?

imeoer: Does it work for OCI image layer?

changweige:
LGTM

changweige (approved):

snapshot/snapshot.go
changweige: As discussed with @power-more , it does not affect the OCI layer.

PR #172, opened 2022-09-14
detect go race condition when perform e2e test

changweige:
e2e test already builds nydus-snapshotter with `-race` flag. Now let's add a checkpoint for go race detection.

### DIFF
--- a/integration/entrypoint.sh
+++ b/integration/entrypoint.sh
@@ -30,6 +30,20 @@ RETRYNUM=30
 RETRYINTERVAL=1
 TIMEOUTSEC=180
 
+GORACE_REPORT="$(pwd)/go_race_report"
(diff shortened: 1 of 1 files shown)


integration/entrypoint.sh
```diff
@@ -30,6 +30,20 @@ RETRYNUM=30
 RETRYINTERVAL=1
 TIMEOUTSEC=180
 
+GORACE_REPORT="$(pwd)/go_race_report"
+export GORACE="log_path=${GORACE_REPORT}"
+
+function race_happens {
```
imeoer, line 36: Maybe name `detect_go_race`?

changweige: good idea

PR #173, opened 2022-09-15
add e2e test for nydus-snapshotter OCI/overlayfs compatibility

changweige:
nydus-snapshotter is compatible with overlayfs snapshotter, add a test for it

### DIFF
--- a/integration/entrypoint.sh
+++ b/integration/entrypoint.sh
@@ -20,11 +20,15 @@ JAVA_IMAGE=${JAVA_IMAGE:-ghcr.io/dragonflyoss/image-service/java:nydus-nightly-v
 WORDPRESS_IMAGE=${WORDPRESS_IMAGE:-ghcr.io/dragonflyoss/image-service/wordpress:nydus-nightly-v6}
 TOMCAT_IMAGE=${TOMCAT_IMAGE:-ghcr.io/dragonflyoss/image-service/tomcat:nydus-nightly-v5}
 STARGZ_IMAGE=${TOMCAT_IMAGE:-ghcr.io/stargz-containers/wordpress:5.9.2-esgz}
+REDIS_OCI_IMAGE=${REDIS_OCI_IMAGE:-ghcr.io/stargz-containers/redis:6.2.6-org}
+WORDPRESS_OCI_IMAGE=${WORDPRESS_OCI_IMAGE:-ghcr.io/stargz-containers/wordpress:5.9.2-org}
 
 # JAVA_IMAGE=${JAVA_IMAGE:-hub.byted.org/gechangwei/java:latest-nydus-v6}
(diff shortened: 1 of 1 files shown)


integration/entrypoint.sh
```diff
@@ -245,13 +264,19 @@ function pull_remove_multiple_images {
...
-    nerdctl --snapshotter nydus image rm "${JAVA_IMAGE}"
-    nerdctl --snapshotter nydus image rm "${WORDPRESS_IMAGE}"
+    nerdctl --snapshotter nydus create --rm --net none "${TOMCAT_IMAGE}"
+    nerdctl --snapshotter nydus create --rm --net none "${WORDPRESS_IMAGE}"
 
-    # TODO: Validate running nydusd number
+    nerdctl --snapshotter nydus image rm --force "${JAVA_IMAGE}"
+    nerdctl --snapshotter nydus image rm --force "${WORDPRESS_IMAGE}"
+
+    nerdctl --snapshotter nydus image rm --force --async "${TOMCAT_IMAGE}"
```
imeoer, line 273: Can we explain why need `--async`?

integration/entrypoint.sh
```diff
@@ -245,13 +264,19 @@ function pull_remove_multiple_images {
     nerdctl --snapshotter nydus image pull "${WORDPRESS_IMAGE}"
     nerdctl --snapshotter nydus image pull "${TOMCAT_IMAGE}"
 
-    nerdctl --snapshotter nydus image rm "${TOMCAT_IMAGE}"
-    nerdctl --snapshotter nydus image rm "${JAVA_IMAGE}"
-    nerdctl --snapshotter nydus image rm "${WORDPRESS_IMAGE}"
+    nerdctl --snapshotter nydus create --rm --net none "${TOMCAT_IMAGE}"
```
imeoer, line 267: Why need to do `image pull` -> `create` -> `image rm` -> `create` ?

integration/entrypoint.sh
changweige: No purpose to adding `--async` flag here, just want to give a fuzz flag to test if it works well

integration/entrypoint.sh
changweige: Because nydusd is not started right after `image pull`. Nydusd is started when preparing the writable active snapshot as the uppermost layer. So we must create a container to start nydusd. Then to test if snapshotter's nydusd daemons management works well

imeoer: Ok, maybe we can add some comments to describe these step by step.

integration/entrypoint.sh
imeoer: Ok, maybe we can add a comment.

integration/entrypoint.sh
changweige: comments all added

imeoer (approved):

PR #184, opened 2022-09-22
Fscache recover

changweige:
Support recover those nydusd daemon that started in fscache driver

### DIFF
--- a/pkg/process/manager.go
+++ b/pkg/process/manager.go
@@ -246,11 +246,11 @@ type Opt struct {
 }
 
 func (m *Manager) handleDaemonDeathEvent() {
-	for event := range m.LivenessNotifier {
-		log.L.Warnf("Daemon %s died! socket path %s", event.daemonID, event.path)
+	for ev := range m.LivenessNotifier {
+		log.L.Warnf("Daemon %s died! socket path %s", ev.daemonID, ev.path)
 
 		if m.RecoverPolicy == RecoverPolicyRestart {
-			go func() {
+			go func(event deathEvent) {
 				d := m.GetByDaemonID(event.daemonID)
 				if d == nil {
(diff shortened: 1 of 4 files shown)


pkg/process/manager.go
```diff
@@ -603,12 +603,34 @@ func (m *Manager) Reconnect(ctx context.Context) ([]*daemon.Daemon, error) {
 		m.daemonStates.RecoverDaemonState(d)
 
 		d.Once = &sync.Once{}
```
imeoer, line 605: Why do we need to prepare to initialize the client here?

pkg/process/manager.go
```diff
@@ -603,12 +603,34 @@ func (m *Manager) Reconnect(ctx context.Context) ([]*daemon.Daemon, error) {
...
+
+		defer func() {
+			// `CheckStatus->ensureClient` only checks if socket file is existed when building http client.
+			// But the socket file may be residual and will be cleared before starting a new nydusd.
+			// So clear the client by assigning nil
+			d.Client = nil
+			d.Once = &sync.Once{}
```
imeoer, line 612: Ditto.

pkg/process/manager.go
changweige: Because method `CheckStatus` is initializing HTTP client when reconnecting to daemons

pkg/process/manager.go
changweige: Because during iterating daemons in DB, their unix sockets might be deleted before restarting a new nydusd. So HTTP client has to be reinitialized

imeoer: Okay, can we put it into something like `d.DestroyClient`? Ditto.

imeoer (approved):

PR #186, opened 2022-09-23
remove ctx from snapshotter object

changweige:
The ctx is only used to print log.
Let's use log.L instead


imeoer (approved):
Thanks! LGTM!

PR #187, opened 2022-09-23
turn delay policy from Backoff to Fixed

changweige:
Backendoff delay policy brings longer and longer
delay when retrying. So it brings long latency to
set up images snapshots when creating containers.


imeoer (approved):
LGTM

PR #190, opened 2022-09-26
Fix daemon race

changweige:
Nydusd daemon record can be wrongly deleted from DB when nydusd fails to start up. A reliable and stable way to fix is to refactor nydusd daemon and rafs instance relationship. So the workaround is just start a new daemon when it can't be found from states cache and DB.

In addition, we should reset client when clear daemons vestige since the API socket has to be re-created by the new nydusd

### DIFF
--- a/integration/entrypoint.sh
+++ b/integration/entrypoint.sh
@@ -239,7 +239,7 @@ function start_multiple_containers_shared_daemon {
 function start_single_container_on_stargz {
     echo "testing $FUNCNAME"
     nerdctl_prune_images
-    reboot_containerd shared
+    reboot_containerd multiple
 
     killall "containerd-nydus-grpc" || true
     sleep 0.5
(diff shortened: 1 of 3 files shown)


integration/entrypoint.sh
```diff
@@ -239,7 +239,7 @@ function start_multiple_containers_shared_daemon {
 function start_single_container_on_stargz {
     echo "testing $FUNCNAME"
     nerdctl_prune_images
-    reboot_containerd shared
+    reboot_containerd multiple
```
imeoer, line 242: Why does the recovered snapshotter start in multiple mode?

changweige: See L 247. The recovered nydus-snapshotter is started in multiple mode

imeoer: Ok

imeoer (approved):

PR #192, opened 2022-09-27
upgrade golang version to 1.18.6

changweige:
No logic is changed, just upgrade golang


changweige:
I safely merge this since it doesn't contain any logics changes

PR #193, opened 2022-09-28
Fix some bugs in image conversion

loheagn:
This patch fixes some bugs in image conversion using `LayerConvertFunc` and `ConvertHookFunc`.

1. append `nydus.remoteimage.v1` to `os.features` of the manifest descriptor in the index
2. add `"containerd.io/snapshot/nydus-blob": "true"` to the annotations of the blob layer descriptor in the manifest
3. avoid appending duplicate bootstrap layer digest to the diffID list of the config

This patch also adds a smoke test `TestContainerdImageConvert` to make sure `LayerConvertFunc` and `ConvertHookFunc` work well with the containerd image conversion library.


pkg/converter/convert_unix.go
```diff
@@ -592,89 +592,130 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
...
-			return nil, errors.Wrap(err, "merge nydus layers")
+// convertIndex modifies the original index by appending "nydus.remoteimage.v1"
+// to the Platform.OSFeatures of each modified manifest descriptors.
+func convertIndex(ctx context.Context, cs content.Store, orgDesc ocispec.Descriptor, newDesc *ocispec.Descriptor) (*ocispec.Descriptor, error) {
+	var orgIndex ocispec.Index
+	if _, err := readJSON(ctx, cs, &orgIndex, orgDesc); err != nil {
+		return nil, errors.Wrap(err, "read target image index json")
+	}
+	// isManifestModified is a function to check whether the manifest is modified.
+	isManifestModified := func(manifest ocispec.Descriptor) bool {
```
imeoer, line 617: Maybe we can check whether it is nydus manifest (name it `isNydusManifest`) by recognizing bootstrap annotation, which would be safer?

pkg/converter/convert_unix.go
```diff
@@ -592,89 +592,130 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
...
+	config.RootFS.DiffIDs = []digest.Digest{}
+	for i, layer := range manifest.Layers {
+		config.RootFS.DiffIDs = append(config.RootFS.DiffIDs, digest.Digest(layer.Annotations[LayerAnnotationUncompressed]))
+		// Remove useless annotation.
+		delete(layer.Annotations, LayerAnnotationUncompressed)
+		manifest.Layers[i] = layer
+	}
+	if opt.Backend != nil {
+		config.RootFS.DiffIDs = []digest.Digest{digest.Digest(bootstrapDesc.Annotations[LayerAnnotationUncompressed])}
+	}
```
imeoer, line 702: Be better:

```
if opt.Backend != nil {
	config.RootFS.DiffIDs = []digest.Digest{digest.Digest(bootstrapDesc.Annotations[LayerAnnotationUncompressed])}
} else {
	config.RootFS.DiffIDs = []digest.Digest{}
	for i, layer := range manifest.Layers {
		config.RootFS.DiffIDs = append(config.RootFS.DiffIDs, digest.Digest(layer.Annotations[LayerAnnotationUncompressed]))
		// Remove useless annotation.
		delete(layer.Annotations, LayerAnnotationUncompressed)
		manifest.Layers[i] = layer
	}
}
```

imeoer (commented):
Thanks for the contribution!

pkg/converter/convert_unix.go
loheagn: Then we need to read json from the content store for each manifest descriptor.

loheagn: Maybe we can store the modified manifest digest in a map in the manifest conversion hook?

pkg/converter/convert_unix.go
loheagn: Fixed.

pkg/converter/convert_unix.go
imeoer: Well, just checking if the manifest is changed should also work.

pkg/converter/convert_unix.go
```diff
@@ -592,89 +592,131 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
...
+		return nil, errors.Wrap(err, "read image config")
+	}
+	if opt.Backend != nil {
+		config.RootFS.DiffIDs = []digest.Digest{digest.Digest(bootstrapDesc.Annotations[LayerAnnotationUncompressed])}
+	} else {
+		config.RootFS.DiffIDs = make([]digest.Digest, 0, len(manifest.Layers))
+		for i, layer := range manifest.Layers {
 			config.RootFS.DiffIDs = append(config.RootFS.DiffIDs, digest.Digest(layer.Annotations[LayerAnnotationUncompressed]))
 			// Remove useless annotation.
 			delete(layer.Annotations, LayerAnnotationUncompressed)
```
imeoer, line 700: Just use `delete(manifest.Layers[i].Annotations, LayerAnnotationUncompressed)`.

loheagn: OK

imeoer (approved):

loheagn:
@imeoer Can you please rerun the action? There may be a network error. Thanks :)

changweige:
The current main branch has Golang race problem, which fails this PR. I will fix it

changweige:
Please rebase to the latest main branch.

imeoer (approved):

PR #195, opened 2022-09-28
print which directories are being deleted

changweige:
This is very helpful when debug and trace how
nydus-snapshotter is releasing resources.

### DIFF
--- a/pkg/process/manager.go
+++ b/pkg/process/manager.go
@@ -275,6 +275,9 @@ func (m *Manager) handleDaemonDeathEvent() {
 					daemons := m.daemonStates.List()
 					for _, d := range daemons {
 						if d.ID != daemon.SharedNydusDaemonID {
+							// FIXME: Virtual daemon has a separated client, so it skips checking unix socket's existence.
+							// This is really hacky, but I don't have better solution until rafs object is decoupled from daemon.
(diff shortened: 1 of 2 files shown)


snapshot/snapshot.go
```diff
@@ -822,6 +822,8 @@ func (o *snapshotter) cleanupSnapshotDirectory(ctx context.Context, dir string)
 		log.G(ctx).WithError(err).WithField("dir", dir).Error("failed to unmount")
 	}
 
+	log.L.Infof("Delete snapshot directory %s", dir)
+
 	if err := os.RemoveAll(dir); err != nil {
```
imeoer, line 827: Move after here and say `Deleted snapshot directory %s`?

pkg/process/manager.go
```diff
@@ -400,6 +400,9 @@ func (m *Manager) CleanUpDaemonResources(d *daemon.Daemon) {
 	if d.IsMultipleDaemon() {
 		resource = append(resource, d.SocketDir)
 	}
+
+	log.L.Infof("deleting resources %v", resource)
+
 	for _, dir := range resource {
```
imeoer, line 406: Move after here and say `Deleted resources %v`?

PR #200, opened 2022-10-04
Add nydusd daemon command line builder

changweige:
Generate nydusd exec style command line via a daemonCommand object. It provides the flexibility to start nydusd as a library API. 
Move nydusd daemon startup logics to a new file `daemon_adapter.go`

### DIFF
--- a/pkg/daemon/command/command.go
+++ b/pkg/daemon/command/command.go
@@ -1,0 +1,169 @@
+/*
+ * Copyright (c) 2022. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package command
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+type Opt = func(cmd *DaemonCommand)
+
+// Define how to build a command line to start a nydusd daemon
+type DaemonCommand struct {
+	// "singleton" "fuse"
+	Mode string `type:"subcommand"`
+	// "blobcache" "fscache" "virtiofs"
+	FscacheDriver  string `type:"param" name:"fscache"`
+	FscacheThreads string `type:"param" name:"fscache-threads"`
+	Upgrade        bool   `type:"flag" name:"upgrade" default:""`
+	ThreadNum      string `type:"param" name:"thread-num"`
+	Config         string `type:"param" name:"config"`
+	Bootstrap      string `type:"param" name:"bootstrap"`
+	Mountpoint     string `type:"param" name:"mountpoint"`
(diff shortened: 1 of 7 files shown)


pkg/daemon/command/command.go
```diff
@@ -0,0 +1,161 @@
...
+	for _, o := range opts {
+		o(&cmd)
+	}
+
+	args := make([]string, 0, 32)
+	t := reflect.TypeOf(cmd)
+	v := reflect.ValueOf(cmd)
+
+	// Zero value will be skipped
+	if sc := v.FieldByName("Mode"); !sc.IsZero() {
```
imeoer, line 48: Can we handle `subcommand` instead of `Mode` here?

pkg/daemon/command/command.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+type Opt = func(cmd *DaemonCommand)
+
+// Define how to build a command line to start a nydusd daemon
+type DaemonCommand struct {
+	// "singleton" "fuse"
+	Mode string `type:"subcommand"`
+	// "blobcache" "fscache" "virtiofs"
+	FscacheDriver  string `type:"param" name:"fscache" default:""`
```
imeoer, line 22: Seems we haven't handled the `default` tag, can we remove it first?

changweige: Yes, lets remove it for now

pkg/daemon/command/command.go
```diff
@@ -0,0 +1,167 @@
...
+				continue
+			}
+		}
+	}
+
+	if subcommand != "" {
+		// Ensure subcommand is at the first place.
+		r := make([]string, 1)
+		r[0] = subcommand
+		args = append(r, args...)
```
imeoer, line 91: `args = append([]string{subcommand}, args...)`

pkg/daemon/command/command.go
```diff
@@ -0,0 +1,167 @@
...
+
+			v := v.Field(i).Bool()
+
+			if v {
+				flag := fmt.Sprintf("--%s", tag.Get("name"))
+				args = append(args, flag)
+			} else {
+				continue
+			}
+		}
```
imeoer, line 84: Better to handle `default` and give an error message, in case of an invalid option field is added.

changweige: good point

changweige: fixed

pkg/daemon/command/command.go
changweige: fixed

imeoer (approved):

PR #201, opened 2022-10-08
System http server

changweige:
Add a system controller which runs a HTTP server as a user interface to interact with nydus-snapshotter. 
With the system controller, we can fetch all managed nydusds's working state and control nydusd workflows.

Nydus-snapshotter might manage dozens of running nydus daemons, each daemon may have multiple
file system instances attached. For easy maintenance, the system controller can interact with
all the daemons in a consistent and automatic way.


imeoer:
Why not reuse the metrics server?

changweige:
> Why not reuse the metrics server?

Looks like we can merge the two HTTP servers.

changweige:
Prometheus metrics server and system contorller api server are merged

cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -212,6 +214,12 @@ func buildFlags(args *Args) []cli.Flag {
 			Usage:       "whether to validate integrity of image bootstrap",
 			Destination: &args.ValidateSignature,
 		},
+		&cli.StringFlag{
+			Name:        "api-socket",
+			Value:       defaultAPISocket,
+			Usage:       "path to the nydus-snapshotter HTTP server",
```
imeoer, line 220: `unix domain socket path to serve HTTP-based system management interface?`
In addition, can we also support `ip:port`? so that allows prometheus server to access metrics from remote.

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+	APISock    string `json:"api_socket"`
+	// TODO: trim me, only for trace and analyze purpose at this stage
+	RootMountPoint   string
+	CustomMountPoint string
+	SupervisorPath   string
+	SnapshotDir      string
+}
+
+func NewSystemController(manager *manager.Manager, sock string) (*Controller, error) {
+	os.Remove(sock)
```
imeoer, line 71: The err should be handled.

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+
+		w.Write(respBody)
+	}
+}
+
+// TODO: Implement me!
+func (sc *Controller) getDaemonRecords() func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		daemons := sc.manager.ListDaemons()
+		log.L.Infof("list daemons %v", daemons)
```
imeoer, line 150: Please write a `501 Not Implemented` status code at least, otherwise the API call will hang/timeout.

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/containerd/containerd/log"
+	"github.com/containerd/nydus-snapshotter/pkg/daemon"
+	"github.com/containerd/nydus-snapshotter/pkg/manager"
+	"github.com/containerd/nydus-snapshotter/pkg/metrics/exporter"
+	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const (
```
imeoer, line 25: We'd better mark all these endpoints as `expirimental`, so that we have a chance to adjust the response JSON format.

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+// 5. If upgrading a certain nydusd fails, abort!
+// 6. Delete the old nydusd executive
+func (sc *Controller) upgradeDaemons() func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		daemons := sc.manager.ListDaemons()
+
+		var c upgradeRequest
+
+		err := json.NewDecoder(r.Body).Decode(&c)
+		if err != nil {
```
imeoer, line 179: `if err := json.NewDecoder(r.Body).Decode(&c); err != nil {`

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+			log.L.Errorf("decode error, %v", r)
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
imeoer, line 181: Why not `http.StatusInternalServerError`?

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+		// Create a dedicated directory storing nydusd of various versions?
+
+		// TODO: Let struct `Daemon` have a method to start a process rather than forking
+		// process in daemons manager
+		// TODO: daemon client has a method to query daemon version and information.
+		for _, d := range daemons {
+			err := upgradeNydusDaemon(d, c)
+			if err != nil {
+				log.L.Errorf("Upgrade daemon %s failed", d.ID)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
```
imeoer, line 196: Better to define the error structure like the one below, instead of responding with a plain error message:

```
Status Code: 500

Response Body:
{
    code: "ERR_UPGRADE_FAILED",
    message: "blabla..."
}
```

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+		// TODO: Keep the nydusd executive path in Daemon state and persis it since nydusd
+		// can run on both versions.
+		// Create a dedicated directory storing nydusd of various versions?
+
+		// TODO: Let struct `Daemon` have a method to start a process rather than forking
+		// process in daemons manager
+		// TODO: daemon client has a method to query daemon version and information.
+		for _, d := range daemons {
+			err := upgradeNydusDaemon(d, c)
+			if err != nil {
```
imeoer, line 194: `if err := upgradeNydusDaemon(d, c); err != nil {`

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+				log.L.Errorf("Upgrade daemon %s failed", d.ID)
```
imeoer, line 195: Should print with `err`?

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+// 6. Delete the old nydusd executive
+func (sc *Controller) upgradeDaemons() func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		daemons := sc.manager.ListDaemons()
+
+		var c upgradeRequest
+
+		err := json.NewDecoder(r.Body).Decode(&c)
+		if err != nil {
+			log.L.Errorf("decode error, %v", r)
```
imeoer, line 180: Why print `r` instead of `err`?

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+
+			info = append(info, i)
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Header().Set("Content-Type", "application/json")
+
+		respBody, err := json.Marshal(&info)
+		if err != nil {
+			panic("marshal")
```
imeoer, line 139: Please respond with `500` instead of panic.

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+				Pid:              d.Pid,
+				ImageID:          d.ImageID,
+				RootMountPoint:   RootMountPointGetter(),
+				CustomMountPoint: CustomMountPointGetter(),
+				SnapshotDir:      d.SnapshotDir}
+
+			info = append(info, i)
+		}
+
+		w.WriteHeader(http.StatusOK)
```
imeoer, line 134: Can we provide a simple method named `respondJSON(w http.ResponseWriter, payload interface{})`?

pkg/system/system.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+		}
+	}
+}
+
+func (sc *Controller) Run() error {
+	log.L.Infof("Start system controller API server on %s", sc.addr)
+	listener, err := net.ListenUnix("unix", sc.addr)
+	if err != nil {
+		return err
```
imeoer, line 207: We'd better errors.Wrap all `err`.

snapshot/snapshot.go
```diff
@@ -76,19 +77,24 @@ func NewSnapshotter(ctx context.Context, cfg *config.Config) (snapshots.Snapshot
...
 		CacheDir:         cfg.CacheDir,
 		RecoverPolicy:    rp,
 	})
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to new process manager")
+		return nil, errors.Wrap(err, "create daemons manager")
+	}
+
+	if cfg.APISocket != "" {
+		systemController, _ := system.NewSystemController(manager, cfg.APISocket)
```
imeoer, line 92: Why ignore error here?

cmd/containerd-nydus-grpc/pkg/command/flags.go
changweige: Sure. we can expose a IP based http server later

pkg/system/system.go
changweige: Because the request can't be decoded, the json in request is not leggal.

cmd/containerd-nydus-grpc/pkg/command/flags.go
changweige: fixed

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: marked

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: fixed

snapshot/snapshot.go

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: fixed

pkg/system/system.go
changweige: fixed

pkg/system/system.go
```diff
@@ -0,0 +1,255 @@
...
+	msg, _ := json.Marshal(&m)
+	return string(msg)
+}
+
+func jsonResponse(w http.ResponseWriter, payload interface{}) {
+	respBody, err := json.Marshal(&payload)
+	if err != nil {
+		log.L.Errorf("marshal error, %s", err)
+		m := newErrorMessage(err.Error())
+		http.Error(w, m.encode(), http.StatusInternalServerError)
```
imeoer, line 81: Does any error represent `500`? Or we will adjust in the following PR?

pkg/system/system.go
```diff
@@ -0,0 +1,255 @@
...
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(respBody)
```
imeoer, line 87: The err should be handled.

imeoer (approved):

PR #203, opened 2022-10-09
Validate daemon mode

changweige:
We should give a explicit error if daemon mode argument is not legal.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020. Ant Group. All rights reserved.
+ * Copyright (c) 2022. Nydus Developers. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -122,11 +133,16 @@ func SetStartupParameter(startupFlag *command.Args, cfg *Config) error {
 	cfg.PublicKeyFile = startupFlag.PublicKeyFile
 	cfg.ValidateSignature = startupFlag.ValidateSignature
 
+	if err := validateDaemonMode(startupFlag.DaemonMode); err != nil {
+		return errors.Wrapf(err, "invalidate daemon mode %s", startupFlag.DaemonMode)
```
imeoer, line 137: `invalid`

config/config.go
```diff
@@ -86,22 +87,32 @@ type SnapshotterConfig struct {
...
+		return nil, errors.Wrapf(err, "load snapshotter configuration file %q", snapshotterConfigPath)
 	}
 	if err = tree.Unmarshal(config); err != nil {
-		return nil, errors.Wrapf(err, "failed to unmarshal snapshotter config file %q", snapshotterConfigPath)
+		return nil, errors.Wrapf(err, "unmarshal snapshotter configuration file %q", snapshotterConfigPath)
 	}
 	return &config, nil
 }
 
+func validateDaemonMode(mode string) error {
```
imeoer, line 106: How about

```
type DaemonMode string

func (dm DaemonMode) IsValid() error {
	switch dm {
	case DaemonModeMultiple, DaemonModeShared, DaemonModeNone, DaemonModePrefetch:
			return nil
	}
	return errdefs.ErrInvalidArgument
}
```

changweige: Nice code.

config/config.go
changweige: fixed

config/config.go
changweige: fixed

imeoer (approved):

PR #208, opened 2022-10-12
command builder recognize id parameter

changweige:
Also recognize --id parameter when building the command line

### DIFF
--- a/pkg/daemon/command/command.go
+++ b/pkg/daemon/command/command.go
@@ -25,13 +25,15 @@ type DaemonCommand struct {
 	FscacheThreads string `type:"param" name:"fscache-threads"`
 	Upgrade        bool   `type:"flag" name:"upgrade" default:""`
 	ThreadNum      string `type:"param" name:"thread-num"`
-	Config         string `type:"param" name:"config"`
-	Bootstrap      string `type:"param" name:"bootstrap"`
-	Mountpoint     string `type:"param" name:"mountpoint"`
-	APISock        string `type:"param" name:"apisock"`
-	LogLevel       string `type:"param" name:"log-level"`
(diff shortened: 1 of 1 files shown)


pkg/daemon/command/command.go
```diff
@@ -25,13 +25,15 @@ type DaemonCommand struct {
...
 	Upgrade        bool   `type:"flag" name:"upgrade" default:""`
 	ThreadNum      string `type:"param" name:"thread-num"`
-	Config         string `type:"param" name:"config"`
-	Bootstrap      string `type:"param" name:"bootstrap"`
-	Mountpoint     string `type:"param" name:"mountpoint"`
-	APISock        string `type:"param" name:"apisock"`
-	LogLevel       string `type:"param" name:"log-level"`
-	Supervisor     string `type:"param" name:"supervisor"`
-	LogFile        string `type:"param" name:"log-file"`
+	// `--id` is required by `--supervisor` when starting nydusd
```
imeoer, line 28: It's used in hot upgrade/failover case?

pkg/daemon/command/command.go
```diff
@@ -25,6 +25,7 @@ type DaemonCommand struct {
 	FscacheThreads string `type:"param" name:"fscache-threads"`
 	Upgrade        bool   `type:"flag" name:"upgrade" default:""`
 	ThreadNum      string `type:"param" name:"thread-num"`
+	ID             string `type:"param" name:"id"`
```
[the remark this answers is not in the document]
changweige: Added a comment line to tell why we need the parameter. It is nydusd start arguments restriction

pkg/daemon/command/command.go
changweige: Yes. Because it is required by `--supervisor`

PR #210, opened 2022-10-12
support configure authorization through mirror

sctb512:
Nydusd needs a parameters to determine if it is authorized through mirror. 
So snapshotter should also load auth_through item and provide it to nydusd.

### DIFF
--- a/config/daemonconfig.go
+++ b/config/daemonconfig.go
@@ -61,8 +61,9 @@ type DaemonConfig struct {
 }
 
 type MirrorConfig struct {
-	Host    string            `json:"host,omitempty"`
-	Headers map[string]string `json:"headers,omitempty"`
+	Host        string            `json:"host,omitempty"`
+	Headers     map[string]string `json:"headers,omitempty"`
+	AuthThrough bool              `json:"auth_through,omitempty"`
 }
 type BackendConfig struct {
 	// Localfs backend configs


config/daemonconfig.go
```diff
@@ -61,8 +61,9 @@ type DaemonConfig struct {
 }
 
 type MirrorConfig struct {
-	Host    string            `json:"host,omitempty"`
-	Headers map[string]string `json:"headers,omitempty"`
+	Host        string            `json:"host,omitempty"`
+	Headers     map[string]string `json:"headers,omitempty"`
+	AuthThrough bool              `json:"auth_through"`
```
changweige, line 66: Should also add `omitempty` to json tag. Then the "auth_through" won't appear in json file if it is never set of be false.

imeoer (commented):
Maybe give a doc somewhere?

config/daemonconfig.go
```diff
@@ -61,8 +61,9 @@ type DaemonConfig struct {
 }
 
 type MirrorConfig struct {
-	Host    string            `json:"host,omitempty"`
-	Headers map[string]string `json:"headers,omitempty"`
+	Host        string            `json:"host,omitempty"`
+	Headers     map[string]string `json:"headers,omitempty"`
+	AuthThrough bool              `json:"auth_through,omitempty"`
```
imeoer, line 66: This option should also be available for `proxy` mode, can we move it to `device.backend.config.auth_through`?

PR #211, opened 2022-10-12
support shared domain

userzj:
User could configure the 'share_domain' field in config file to use this feature.

Be cautious! If you use this feature, make sure your kernel version is higher than 6.1.

Signed-off-by: Jia Zhu <zhujia.zj@bytedance.com>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -156,6 +156,9 @@ func SetStartupParameter(startupFlag *command.Args, cfg *Config) error {
 	}
 	cfg.PublicKeyFile = startupFlag.PublicKeyFile
 	cfg.ValidateSignature = startupFlag.ValidateSignature
+	if len(cfg.DaemonCfg.DomainID) != 0 {
+		log.L.Warnf("Linux Kernel Shared Domain feature in use. make sure your kernel version >= 6.1")
+	}
 
 	daemonMode, err := parseDaemonMode(startupFlag.DaemonMode)
 	if err != nil {
--- a/config/daemonconfig.go
+++ b/config/daemonconfig.go
@@ -168,7 +168,9 @@ func NewDaemonConfig(fsDriver string, cfg DaemonConfig, imageID, snapshotID stri
 			backendConfig = &cfg.Config.BackendConfig
 			fscacheID := erofs.FscacheID(snapshotID)
 			cfg.ID = fscacheID
-			cfg.DomainID = fscacheID
+			if len(cfg.DomainID) == 0 {
+				cfg.DomainID = fscacheID
+			}
 			cfg.Config.ID = fscacheID
 		}
 		if keyChain != nil {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -57,6 +57,7 @@ type Daemon struct {
 	// It should only be used to distinguish daemons that needs to be started when restarting nydus-snapshotter
 	Connected bool       `json:"-"`
 	mu        sync.Mutex `json:"-"`
+	domainID  string     `json:"-"`
 }
 
 func (d *Daemon) Lock() {
@@ -250,7 +251,7 @@ func (d *Daemon) sharedErofsMount() error {
 	}
 	fscacheID := erofs.FscacheID(d.SnapshotID)
 
-	if err := erofs.Mount(bootstrapPath, fscacheID, mountPoint); err != nil {
+	if err := erofs.Mount(bootstrapPath, d.domainID, fscacheID, mountPoint); err != nil {
 		if !errdefs.IsErofsMounted(err) {
 			return errors.Wrapf(err, "mount erofs to %s", mountPoint)
 		}
--- a/pkg/utils/erofs/erofs.go
+++ b/pkg/utils/erofs/erofs.go
@@ -12,12 +12,22 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func Mount(bootstrapPath, fsID, mountPoint string) error {
+func Mount(bootstrapPath, domainID, fscacheID, mountPoint string) error {
 	mount := unix.Mount
+	var opts string
 
-	opts := "fsid=" + fsID
+	if domainID != "" {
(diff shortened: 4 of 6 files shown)


pkg/daemon/daemon.go
```diff
@@ -234,7 +234,7 @@ func (d *Daemon) sharedErofsMount() error {
 	}
 	fscacheID := erofs.FscacheID(d.SnapshotID)
 
-	if err := erofs.Mount(bootstrapPath, fscacheID, mountPoint); err != nil {
+	if err := erofs.Mount(bootstrapPath, fscacheID, config.SharedDomainID, mountPoint); err != nil {
```
imeoer, line 237: Why not use `cfg.DomainID` here?

config/config.go
```diff
@@ -156,6 +158,10 @@ func SetStartupParameter(startupFlag *command.Args, cfg *Config) error {
 	}
 	cfg.PublicKeyFile = startupFlag.PublicKeyFile
 	cfg.ValidateSignature = startupFlag.ValidateSignature
+	SharedDomainID = daemonCfg.SharedDomain
```
imeoer, line 161: Better not to use global variable.

pkg/utils/erofs/erofs.go
```diff
@@ -12,12 +12,22 @@ import (
...
+		opts = fmt.Sprintf("domain_id=%s,fsid=%s", shareDomainID, fsID)
+	} else {
+		opts = "fsid=" + fsID
+	}
 	logrus.Infof("Mount erofs to %s with options %s", mountPoint, opts)
+
 	if err := mount("erofs", mountPoint, "erofs", 0, opts); err != nil {
+		if errors.Is(err, unix.EINVAL) && shareDomainID != "" {
+			logrus.Errorf("mount erofs with shared domain failed," +
+				"If using this feature, make sure your Linux kernel version > 6.1")
```
changweige, line 29: > Linux kernel version > 6.1

Should include 6.1 or exclude it?

userzj: include 6.1, will fix it

pkg/utils/erofs/erofs.go
```diff
@@ -12,12 +12,22 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func Mount(bootstrapPath, fsID, mountPoint string) error {
+func Mount(bootstrapPath, fsID, shareDomainID, mountPoint string) error {
```
changweige, line 15: Better to rename it `sharedDomainID`

pkg/utils/erofs/erofs.go
```diff
@@ -12,12 +12,22 @@ import (
...
-	opts := "fsid=" + fsID
+	if shareDomainID != "" {
+		opts = fmt.Sprintf("domain_id=%s,fsid=%s", shareDomainID, fsID)
+	} else {
+		opts = "fsid=" + fsID
+	}
 	logrus.Infof("Mount erofs to %s with options %s", mountPoint, opts)
+
 	if err := mount("erofs", mountPoint, "erofs", 0, opts); err != nil {
+		if errors.Is(err, unix.EINVAL) && shareDomainID != "" {
```
changweige, line 27: Well. Snapshotter won't retry if syscall `mount` returns EINVAL?

config/daemonconfig.go
```diff
@@ -168,7 +168,11 @@ func NewDaemonConfig(fsDriver string, cfg DaemonConfig, imageID, snapshotID stri
 			backendConfig = &cfg.Config.BackendConfig
 			fscacheID := erofs.FscacheID(snapshotID)
 			cfg.ID = fscacheID
-			cfg.DomainID = fscacheID
+			if len(SharedDomainID) == 0 {
+				cfg.DomainID = fscacheID
+			} else {
+				cfg.DomainID = SharedDomainID
+			}
```
changweige, line 175: The logic of getting and parsing `shared_domain_id` is a little dim and tedious. IIUC, the user is wanting the shared domain feature by adding an item to the nydusd configuration like `shared_domain_id=blabla`, so it can be loaded from the json file rather than read from a global variable at compiling time. We don't even need to touch the item but just pass it to the in-memory fscache configuration and use it when perform EROFS mount.  So the assignment `cfg.DomainID = SharedDomainID` here is not necessary

pkg/daemon/daemon.go
```diff
@@ -250,7 +250,7 @@ func (d *Daemon) sharedErofsMount() error {
...
```
changweige, line 253: Looks like only one shared domain is supported?

config/config.go
```diff
@@ -25,6 +25,8 @@ import (
 // Define a policy how to fork nydusd daemon and attach file system instances to serve.
 type DaemonMode string
 
+var SharedDomainID string
```
changweige, line 28: For global variable, should consider race very carefully.

pkg/daemon/daemon.go
userzj: fixed

pkg/utils/erofs/erofs.go
userzj: Yup, if user's system doesn't support shared domain, an error will be returned directly an won't do fallback mount anymore.

pkg/utils/erofs/erofs.go
```diff
@@ -12,12 +12,22 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func Mount(bootstrapPath, fsID, mountPoint string) error {
+func Mount(bootstrapPath, domainID, fscacheID, mountPoint string) error {
 	mount := unix.Mount
+	var opts string
 
-	opts := "fsid=" + fsID
+	if domainID != "" {
```
imeoer, line 19: It seems always `domainID != ""`.

changweige: It can be loaded from rafs configuration json file

changweige (approved):
LGTM

PR #212, opened 2022-10-13
lock daemon client when requesting to daemon

changweige:
There is a race between reseting client during recovering daemons and requesting to daemon.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -144,6 +144,10 @@ func (d *Daemon) State() (types.DaemonState, error) {
 		return types.DaemonStateUnknown, err
 	}
 
+	// Protect daemon client when it's being reset.
+	d.Lock()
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -144,6 +144,10 @@ func (d *Daemon) State() (types.DaemonState, error) {
 		return types.DaemonStateUnknown, err
 	}
 
+	// Protect daemon client when it's being reset.
+	d.Lock()
```
imeoer, line 148: How can we ensure no new race is brought for new operations? In other word, can we simply the `ensureClient(), Lock(), Unlock()` code?

changweige: True. We need to figure out a code restriction framework to close the race once and for all. But now let's fix the race problem first since it occasionally fails e2e test

PR #218, opened 2022-10-19
Fix the annotation error of the converted bootstrap layer

loheagn:
This pr fixes the fs-version annotation error of the converted Nydus bootstrap layer.

By now, the "containerd.io/snapshot/nydus-fs-version" annotation of the converted bootstrap layer is always `5` which is the default value. The reason for this error is that the value of `MergeOption.FsVersion` is not assigned correctly in the `convertManifest` function of `convert_unix.go` when call the `MergeLayers` function and the `MergeLayers` always uses the empty string as the value of `MergeOption.FsVersion`.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -660,6 +660,7 @@ func convertManifest(ctx context.Context, cs content.Store, newDesc *ocispec.Des
 		BuilderPath:   opt.BuilderPath,
 		WorkDir:       opt.WorkDir,
 		ChunkDictPath: opt.ChunkDictPath,
+		FsVersion:     opt.FsVersion,
 		WithTar:       true,
 	})
 	if err != nil {


changweige:
Um. So how the teset case for the converter is passed in snapshotter's repo?

loheagn:
> Um. So how the teset case for the converter is passed in snapshotter's repo?

Because the test code coincidentally use `5` as the value of `FsVersion`.

[https://github.com/containerd/nydus-snapshotter/blob/8abf82f809600c351c2c8041854597b5c936a7d4/tests/converter_test.go#L435](https://github.com/containerd/nydus-snapshotter/blob/8abf82f809600c351c2c8041854597b5c936a7d4/tests/converter_test.go#L435)

changweige:
Should we also port this patch to the stable/v0.3 branch? I am going to release v0.3.0 today

changweige:
> > Should we also port this patch to the stable/v0.3 branch? I am going to release v0.3.0 today
> 
> Yes. I think we need a new release for this patch.

Well. Please help to raise another PR to stable/v0.3. 
Thanks.

PR #221, opened 2022-10-20
Fix the annotation error of the converted bootstrap layer

loheagn:
This pr fixes the fs-version annotation error of the converted Nydus bootstrap layer.

By now, the "containerd.io/snapshot/nydus-fs-version" annotation of the converted bootstrap layer is always `5` which is the default value. The reason for this error is that the value of `MergeOption.FsVersion` is not assigned correctly in the `convertManifest` function of `convert_unix.go` when call the `MergeLayers` function and the `MergeLayers` always uses the empty string as the value of `MergeOption.FsVersion`.


changweige (approved):
LGTM

PR #233, opened 2022-11-03
remove complex blob cache GC code

changweige:
Containerd can contorl if a committed snapshot should be removed. We don't have to consider if snapshot belongs to multiple images.

Just respond to containerd a committed snapshot shoud be removed.


pkg/converter/convert_unix.go
```diff
@@ -832,7 +824,6 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 			// Use this annotation to identify nydus bootstrap layer.
 			LayerAnnotationNydusBootstrap: "true",
 			// Track all blob digests for nydus snapshotter.
```
imeoer, line 826: Should also remove the line comment.

pkg/cache/manager.go
```diff
@@ -136,3 +86,24 @@ func (m *Manager) CacheUsage(ctx context.Context, blobID string) (snapshots.Usag
...
 	return usage, nil
 }
+
+func (m *Manager) RemoveBlobCache(blobID string) error {
+	blobCachePath := path.Join(m.cacheDir, blobID)
+	blobCacheSuffixedPath := path.Join(m.cacheDir, blobID+dataFileSuffix)
+	blobChunkMap := path.Join(m.cacheDir, blobID+chunkMapFileSuffix)
+	blobMeta := path.Join(m.cacheDir, blobID+metaFileSuffix)
+
+	stuffs := []string{blobCachePath, blobChunkMap, blobMeta, blobCacheSuffixedPath}
```
imeoer, line 96: Should remove `blobChunkMap` first, then `blobCachePath` or `blobCacheSuffixedPath`?

snapshot/snapshot.go
```diff
@@ -423,34 +419,24 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
...
-					log.G(ctx).WithError(err).Warn("failed to delete snapshot in cache manager")
-				}
-			} else {
-				log.G(ctx).WithError(err).Warnf("failed to get image ref from snapshot label %#v", snap.Labels)
-
-			}
-		}
+	if info.Kind == snapshots.KindCommitted {
+		blobDigest := info.Labels[label.CRILayerDigest]
+		go o.fs.RemoveBlobCache(blobDigest)
```
imeoer, line 429: Perhaps print a warning if failed here.

pkg/converter/convert_unix.go
changweige: removed

pkg/cache/manager.go
changweige: Yes. Good catch. Fixed

snapshot/snapshot.go

pkg/cache/manager.go
```diff
@@ -136,3 +86,25 @@ func (m *Manager) CacheUsage(ctx context.Context, blobID string) (snapshots.Usag
...
+	blobMeta := path.Join(m.cacheDir, blobID+metaFileSuffix)
+
+	// NOTE: Delete chunk bitmap file before data blob
+	stuffs := []string{blobChunkMap, blobMeta, blobCachePath, blobCacheSuffixedPath}
+
+	for _, f := range stuffs {
+		err := os.Remove(f)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				log.L.Warnf("file %s doest not exist.", f)
```
imeoer, line 103: For the new nydusd version, the `blobCachePath` always not exists, perhaps we just need to skip the error.

snapshot/snapshot.go
```diff
@@ -423,41 +419,35 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
...
-			if imageID, ok := snap.Labels[label.CRIImageRef]; ok {
-				if err := o.fs.DelSnapshot(imageID); err != nil {
-					log.G(ctx).WithError(err).Warn("failed to delete snapshot in cache manager")
-				}
-			} else {
-				log.G(ctx).WithError(err).Warnf("failed to get image ref from snapshot label %#v", snap.Labels)
-
+	if info.Kind == snapshots.KindCommitted {
+		blobDigest := info.Labels[label.CRILayerDigest]
+		go func(digest string) {
```
imeoer, line 429: The parameter seems unnecessary.

pkg/cache/manager.go
changweige: fixed

snapshot/snapshot.go
changweige: fixed

imeoer (approved):
Thanks, LGTM!

PR #236, opened 2022-11-08
decouple Rafs instance from daemon

changweige:
Currently, a virtual daemon represents a rafs instance when invoking API mount against nydusd. It brings a lot of complexity to the filesystem/snapshots control flow. It also makes it harder to implement covering nydusd by restarting or failover since a clear rule is lacking to distinguish a read nydusd daemon and a virtual daemon(rafs instance).

Moreover, there is a deployment case which no nydusd daemon is needed to run a.k.a. `noneDaemonMode`. So rafs instances have their own management structures now in which each rafs instance can refer to a runtime daemon.

In addition, this PR redesign the boltdb bucket records and moves unrelated code from `filesystem` package to different places.


imeoer (approved):
LGTM, let's move forward first.

PR #239, opened 2022-11-08
[Bugfix] Use the correct label to distinguish the lower layer from th…

power-more:
…e upper layer

Previously, the unreliable (may not exist, such as when disable_snapshot_annotations = true) label added by cri was used as the basis for judging the upper layer, then when the OCI image without the label was started, it would mistakenly enter prepare for upper layer logic.

We noticed that "containerd.io/snapshot.ref" is a label added by containerd before calling sn.Prepare(), and only exists in the lower layer snapshot, not in the upper layer, so we can simply put the negative situation of lower layer as the upper layer.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -338,10 +338,8 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 				}
 			}
 		}
-	}
-
-	// Mount image for running container, which has a nydus/stargz image as parent.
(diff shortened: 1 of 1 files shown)


imeoer:
Hi @power-more, did this test in an actual environment?

power-more:
> Hi @power-more, did this test in an actual environment?

not yet

changweige (approved):
LGTM, Thanks

PR #240, opened 2022-11-09
add some comments to prevent users from using return value incorrectly

power-more:

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -532,6 +532,7 @@ func (o *snapshotter) workPath(id string) string {
 	return filepath.Join(o.root, "snapshots", id, "work")
 }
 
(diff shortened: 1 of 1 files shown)


snapshot/snapshot.go
```diff
@@ -532,6 +532,7 @@ func (o *snapshotter) workPath(id string) string {
 	return filepath.Join(o.root, "snapshots", id, "work")
 }
 
+// The first return value is a snapshots.Info containing only the labels, so info.Labels is the only meaningful field
```
imeoer, line 535: Why other fields is not available?

changweige: what field is ever filled is decided by slice []Opt.  We should not make any assumption on what []opt contains

PR #242, opened 2022-11-09
Update golangci-lint to v1.50.1

austinvazquez:
Updates golangci-lint to v1.50.1 and removes a duplicate linter setting.

Signed-off-by: Austin Vazquez <macedonv@amazon.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,7 +39,6 @@ linters:
     - dogsled
     # - nilnil
     - unparam
-    - unused
     - nilerr
     # - goerr113
     - exportloopref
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -39,7 +39,6 @@ linters:
     - dogsled
     # - nilnil
     - unparam
-    - unused
```
austinvazquez, line 42: Duplicate listed above on line 35

changweige: Good catch!

changweige (approved):
LGTM

PR #248, opened 2022-11-15
Running state cache

changweige:
At present, nydus-snapshotter queries daemon's state each
time when it gets daemon object's state. The query comes
from several positions which cause duplicated query.
We can use single-flight mechanism to query daemon'state
to reduce unnecessary comunications between processes.

This changeset will benefit container startup performance.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -135,7 +135,7 @@ func (d *Daemon) AddInstance(r *Rafs) {
 // 1. INIT
 // 2. READY: All needed resources are ready.
 // 3. RUNNING
-func (d *Daemon) State() (types.DaemonState, error) {
+func (d *Daemon) GetState() (types.DaemonState, error) {
 	c, err := d.GetClient()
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -169,23 +163,39 @@ func (d *Daemon) GetState() (types.DaemonState, error) {
...
+		},
+			retry.Attempts(20), // totally wait for 2 seconds, should be enough
+			retry.LastErrorOnly(true),
+			retry.Delay(100*time.Millisecond),
+		)
+
+		return
+	}
+
+	var g singleflight.Group
```
imeoer, line 192: Why new the `singleflight.Group` here, how does it ensures only one is inflight?

changweige: Good catch. 👍

changweige: fixed

imeoer (approved):

PR #249, opened 2022-11-15
Add nydusd live-upgrade control flow

changweige:
Add a control flow of nydusd live-upgrade to the system package.

### DIFF
--- a/pkg/system/system_test.go
+++ b/pkg/system/system_test.go
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2022. Nydus Developers. All rights reserved.
+ *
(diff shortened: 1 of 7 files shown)


pkg/system/system_test.go
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2022. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
hsiangkao, line 4: SPDX should be kept in the first line as a thumb rule.

imeoer (approved):
LGTM

changweige:
Thanks for your review :-)

PR #250, opened 2022-11-15
don't write config file with permissions greater than 0600

changweige:
Don't show config content to other users.


pkg/supervisor/supervisor.go
```diff
@@ -209,6 +209,8 @@ func (su *Supervisor) waitStatesTimeout(to time.Duration) (func() error, error)
 			}
 		}()
 
+		// With non-zero timeout parameter, call should be aware that receiver is nil
```
imeoer, line 212: Can we explain why the nil receiver will be returned if the timeout is zero?

changweige: fixed

imeoer (approved):

PR #253, opened 2022-11-17
fails snapshotter startup if its fsdriver mismatches daemon's

changweige:
I have observed that users misuses snapshotter's fsdirver parameter by changing it between two startups. As the daemon is
persisted to database, snapshotter can't at the same time serves two types of fsdrvier. So we should abort this startup

Otherwise, even snapshotter can restore and recover daemons during this startup, it can't provide correct rootfs.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -91,7 +91,7 @@ type Config struct {
 	NydusdBinaryPath         string        `toml:"nydusd_binary_path"`
 	NydusImageBinaryPath     string        `toml:"nydus_image_binary"`
 	DaemonMode               DaemonMode    `toml:"daemon_mode"`
-	FsDriver                 string        `toml:"daemon_backend"`
+	FsDriver                 string        `toml:"fs_driver"`
 	SyncRemove               bool          `toml:"sync_remove"`
 	EnableMetrics            bool          `toml:"enable_metrics"`
 	MetricsFile              string        `toml:"metrics_file"`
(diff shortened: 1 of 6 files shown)


pkg/manager/manager.go
```diff
@@ -503,6 +509,13 @@ func (m *Manager) Recover(ctx context.Context) (map[string]*daemon.Daemon, map[s
 		var d, _ = daemon.NewDaemon(opt...)
 		d.States = *s
 
+		// It can't change snapshotter's fs driver to a different one from a daemon that ever created in the past.
+		if d.States.FsDriver != m.FsDriver {
+			return errors.Wrapf(errdefs.ErrInvalidArgument,
+				"daemon's fs driver %s mismatches nydus-snapshotter's %s",
```
imeoer, line 515: `can't recover from the last restart, the specified fs-driver=%s option mismatches with the last fs-driver=%s` may be better.

config/config.go
```diff
@@ -91,7 +91,7 @@ type Config struct {
 	NydusdBinaryPath         string        `toml:"nydusd_binary_path"`
 	NydusImageBinaryPath     string        `toml:"nydus_image_binary"`
 	DaemonMode               DaemonMode    `toml:"daemon_mode"`
-	FsDriver                 string        `toml:"daemon_backend"`
+	FsDriver                 string        `toml:"fs_driver"`
```
imeoer, line 94: Does this break compatibility?

changweige: No. We don't have configuration file used in fact.

pkg/manager/manager.go
changweige: fixed

imeoer (approved):

PR #254, opened 2022-11-17
Optimize how nydus meta snapshot is prepared and enable errcheck linter

changweige:
1. Reduce unnecessary database queries
2. enable `errcheck` linter

This PR is very likely to reduce snapshot setup latency.

### DIFF
--- a/pkg/backend/oss.go
+++ b/pkg/backend/oss.go
@@ -141,7 +141,10 @@ func (b *OSSBackend) push(ra content.ReaderAt, blobDigest digest.Digest) error {
 		}
 
 		if err := g.Wait(); err != nil {
-			b.bucket.AbortMultipartUpload(imur)
+			if err := b.bucket.AbortMultipartUpload(imur); err != nil {
+				close(partsChan)
+				return errors.Wrap(err, "aborting upload")
+			}
 			close(partsChan)
 			return errors.Wrap(err, "upload parts")
 		}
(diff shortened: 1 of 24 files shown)


pkg/backend/oss.go
```diff
@@ -141,7 +142,9 @@ func (b *OSSBackend) push(ra content.ReaderAt, blobDigest digest.Digest) error {
 		}
 
 		if err := g.Wait(); err != nil {
-			b.bucket.AbortMultipartUpload(imur)
+			if err := b.bucket.AbortMultipartUpload(imur); err != nil {
+				log.L.WithError(err)
```
imeoer, line 146: Does this print message to stderr? We'd better not do this in a public package.

pkg/blob/blob_manager.go
```diff
@@ -146,7 +146,11 @@ func (b *Manager) PrepareBlobLayer(snapshot storage.Snapshot, labels map[string]
 	if err != nil {
 		return errors.Wrap(err, "rename temp file as blob file")
 	}
-	os.Chmod(blobFile.Name(), 0440)
+
+	err = os.Chmod(blobPath, 0440)
+	if err != nil {
+		return err
```
imeoer, line 152: Better to wrap it.

pkg/backend/oss.go
changweige: No. It prints to logger

changweige: fixed

pkg/blob/blob_manager.go
changweige: fixed

imeoer (approved):

PR #260, opened 2022-11-22
Shutdown nydusd if it serves no filesystem or snapshot

changweige:
In sharedDaemon mode, we should also shut down the shared nydusd if it is not needed by any snapshot.

In addition, we should retain the shared daemon object in fs layer when doing recovery during nydus-snapshotter startup.

### DIFF
--- a/pkg/filesystem/fs/fs.go
+++ b/pkg/filesystem/fs/fs.go
@@ -299,10 +299,6 @@ func (fs *Filesystem) Mount(snapshotID string, labels map[string]string) (err er
 }
 
 func (fs *Filesystem) Umount(ctx context.Context, snapshotID string) error {
-	if !fs.hasDaemon() {
-		return nil
-	}
-
 	instance := daemon.RafsSet.Get(snapshotID)
 	if instance == nil {
 		log.L.Debugf("Not a rafs instance. ID %s", snapshotID)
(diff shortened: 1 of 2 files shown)


pkg/filesystem/fs/fs.go
```diff
@@ -92,26 +91,38 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
...
+		if err := d.WaitUntilState(types.DaemonStateRunning); err != nil {
+			return nil, errors.Wrapf(err, "wait for daemon %s", d.ID())
+		}
+		if err := d.RecoveredMountInstances(); err != nil {
+			return nil, errors.Wrapf(err, "recover mounts for daemon %s", d.ID())
+		}
+
+		// Found shared daemon
+		// Fscache userspace daemon has no host mountpoint.
+		if d.HostMountpoint() == fs.rootMountpoint || fs.fsDriver == config.FsDriverFscache {
```
imeoer, line 114: Maybe put it in a helper func.

pkg/filesystem/fs/fs.go
```diff
@@ -92,26 +91,38 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
...
+			fs.sharedDaemon = d
+			d.IncRef()
+		}
+	}
+
+	for _, d := range liveDaemons {
+		// Found shared daemon
+		if d.HostMountpoint() == fs.rootMountpoint || fs.fsDriver == config.FsDriverFscache {
```
imeoer, line 122: Ditto.

imeoer (approved):
Others LGTM!

pkg/filesystem/fs/fs.go
changweige: extracted a help function tryRetainSharedDaemon

PR #262, opened 2022-11-25
Fscache gc

kevinXYin:
support fscahe gc

### DIFF
--- a/pkg/daemon/client.go
+++ b/pkg/daemon/client.go
@@ -57,7 +57,7 @@ type NydusdClient interface {
 	Umount(mountpoint string) error
 
 	BindBlob(daemonConfig string) error
-	UnbindBlob(daemonConfig string) error
+	UnbindBlob(domainID, blobID string) error
 
 	GetFsMetrics(sid string) (*types.FsMetrics, error)
 	GetCacheMetrics(sid string) (*types.CacheMetrics, error)
@@ -264,9 +264,22 @@ func (c *nydusdClient) BindBlob(daemonConfig string) error {
 	return c.request(http.MethodPut, url, bytes.NewBuffer([]byte(daemonConfig)), nil)
 }
 
-func (c *nydusdClient) UnbindBlob(domainID string) error {
+// Delete /api/v2/blobs implements different functions according to different parameters
+//  1. domainID , delete all blob entries in the domain.
+//  2. domainID + blobID, delete the blob entry, if the blob is bootstrap
+//     also delete blob entries belong to it.
+//  3. blobID, try to find and cull blob cache files by blobID in all domains.
+func (c *nydusdClient) UnbindBlob(domainID, blobID string) error {
 	query := query{}
-	query.Add("domain_id", domainID)
+	if domainID != "" {
+		query.Add("domain_id", domainID)
+		if domainID != blobID {
+			query.Add("blob_id", blobID)
+		}
+	} else {
+		query.Add("blob_id", blobID)
+	}
+
 	url := c.url(endpointBlobs, query)
 
 	return c.request(http.MethodDelete, url, nil, nil)
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -345,6 +345,12 @@ func (d *Daemon) sharedErofsUmount(rafs *Rafs) error {
 		return errors.Wrapf(err, "umount erofs %s mountpoint, %s", err, mountpoint)
 	}
 
+	// delete fscache bootstrap cache file
+	// erofs generate fscache cache file for bootstrap with fscachID
+	if err := c.UnbindBlob("", fscaheID); err != nil {
+		log.L.Warnf("delete bootstrap %s err %s", fscaheID, err)
+	}
+
 	return nil
 }
 
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -312,6 +312,7 @@ func (d *Daemon) sharedErofsMount(rafs *Rafs) error {
 
 	cfg := c.(*daemonconfig.FscacheDaemonConfig)
 	rafs.AddAnnotation(AnnoFsCacheDomainID, cfg.DomainID)
+	rafs.AddAnnotation(AnnoFsCacheID, fscacheID)
 
 	if err := erofs.Mount(bootstrapPath, cfg.DomainID, fscacheID, mountPoint); err != nil {
 		if !errdefs.IsErofsMounted(err) {
@@ -333,8 +334,9 @@ func (d *Daemon) sharedErofsUmount(rafs *Rafs) error {
 		return errors.Wrapf(err, "unbind blob %s", d.ID())
 	}
 	domainID := rafs.Annotations[AnnoFsCacheDomainID]
+	fscaheID := rafs.Annotations[AnnoFsCacheID]
 
-	if err := c.UnbindBlob(domainID); err != nil {
+	if err := c.UnbindBlob(domainID, fscaheID); err != nil {
 		return errors.Wrapf(err, "request to unbind fscache blob")
 	}
 
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -193,11 +193,17 @@ func NewSnapshotter(ctx context.Context, cfg *config.Config) (snapshots.Snapshot
 		return nil, err
 	}
 
+	syncRemove := cfg.SyncRemove
+	if cfg.FsDriver == config.FsDriverFscache {
+		log.L.Infof("for fscache mode enable syncRemove")
+		syncRemove = true
+	}
+
 	return &snapshotter{
 		root:                 cfg.RootDir,
 		nydusdPath:           cfg.NydusdBinaryPath,
 		ms:                   ms,
-		syncRemove:           cfg.SyncRemove,
+		syncRemove:           syncRemove,
 		fs:                   nydusFs,
 		manager:              manager,
 		hasDaemon:            hasDaemon,
--- a/pkg/daemon/rafs.go
+++ b/pkg/daemon/rafs.go
@@ -24,6 +24,7 @@ import (
 
 const (
 	AnnoFsCacheDomainID string = "fscache.domainid"
+	AnnoFsCacheID       string = "fscache.id"
 )
 
 type NewRafsOpt func(r *Rafs) error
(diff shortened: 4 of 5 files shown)


pkg/daemon/daemon.go
```diff
@@ -333,6 +335,11 @@ func (d *Daemon) sharedErofsUmount(rafs *Rafs) error {
 		return errors.Wrapf(err, "umount erofs %s mountpoint, %s", err, mountpoint)
 	}
 
+	// delete fscache bootstrap cache file
+	if err := c.UnbindBlob("", fscaheID); err != nil {
```
changweige, line 339: Does this PR depend on changes in nydusd?

pkg/daemon/client.go
```diff
@@ -262,9 +262,17 @@ func (c *nydusdClient) BindBlob(daemonConfig string) error {
...
+func (c *nydusdClient) UnbindBlob(domainID, blobID string) error {
 	query := query{}
-	query.Add("domain_id", domainID)
+	if domainID != "" {
+		query.Add("domain_id", domainID)
+		if domainID != blobID {
+			query.Add("blob_id", blobID)
+		}
+	} else {
+		query.Add("blob_id", blobID)
```
changweige, line 273: Can we have a comment describing the blob id and domain id combination here or a UT test? Otherwise, it is hard to maintain

pkg/filesystem/fs/fs.go
```diff
@@ -336,6 +336,16 @@ func (fs *Filesystem) RemoveCache(blobDigest string) error {
 		return errors.Wrapf(err, "invalid blob digest from label %s", label.CRILayerDigest)
 	}
 	blobID := digest.Hex()
+	if fs.fsDriver == config.FsDriverFscache {
+		c, err := fs.getSharedDaemon().GetClient()
+		if err != nil {
+			log.L.Warnf("GetClient err %s", err)
```
changweige, line 342: Without a successful client `c`, we can't call `UnbindBlob`, should return error here to containerd

pkg/filesystem/fs/fs.go
```diff
@@ -336,6 +336,16 @@ func (fs *Filesystem) RemoveCache(blobDigest string) error {
...
+		}
+		// delete fscache blob cache file
+		if err := c.UnbindBlob("", blobID); err != nil {
+			log.L.Warnf("delete blob %s err %s", blobID, err)
```
changweige, line 346: Should return error?

pkg/filesystem/fs/fs.go
```diff
@@ -336,6 +336,16 @@ func (fs *Filesystem) RemoveCache(blobDigest string) error {
...
+		}
+	}
 	return fs.cacheMgr.RemoveBlobCache(blobID)
```
changweige, line 349: For fscache driver, skip `RemoveBlobCache` ?

snapshot/snapshot.go
```diff
@@ -203,11 +203,17 @@ func NewSnapshotter(ctx context.Context, cfg *config.Config) (snapshots.Snapshot
 		return nil, err
 	}
 
+	syncRemove := cfg.SyncRemove
+	if cfg.FsDriver == config.FsDriverFscache {
+		log.L.Infof("for fscache mode enable syncRemove")
```
changweige, line 208: We can not use syncRemove flag to control whether to remove fscache files. Just directly remove it in a goroutine

pkg/daemon/daemon.go
kevinXYin: Yes , this depend on changes for nydusd API , so maybe we should review the PR on nydusd first? and make API fixed.
https://github.com/dragonflyoss/image-service/pull/894

pkg/daemon/client.go
kevinXYin: Yeah , will add comment after the API fixed

pkg/filesystem/fs/fs.go
kevinXYin: We also need to delete meta and chunkmap files in RemoveBlobCache. so do we need let nydusd do this for fscache?

snapshot/snapshot.go
kevinXYin: AKAIK, the syncRemove is used for making removing snapshot in order , cause we need remove bootstrap snapshot first and then we can umount erofs. So should we force enable syncRemove for fscache ?

pkg/filesystem/fs/fs.go
kevinXYin: yes , we better return error , but this will cause snapshotter can not forward compatible nydusd?

changweige: You are right, perhaps we should let snapshotter nydusd client is aware or nydusd version to workaround

pkg/daemon/daemon.go
changweige: Sure. Let's wait for that dependency and delay this PR for snapshotter v0.5

pkg/filesystem/fs/fs.go
changweige: It has no effect. Fuse driver and fscache driver have different places to put blog meta files which won't be deleted by the function RemoveBlobCache now...

snapshot/snapshot.go
changweige: We can simply make fscache files deletion out of the syncRemove==true if arm

pkg/filesystem/fs/fs.go
changweige: Nydus-snapshotter cache manager is somewhat conflicting with fscache driver

kevinXYin: Understood , will fix this.

pkg/filesystem/fs/fs.go
kevinXYin: Actually, we call RemoveCache in a goroutine , the error will not pass to containerd.

pkg/daemon/client.go
changweige: As the nydusd API was fixed and merged, shall we have a description comments now?

pkg/daemon/daemon.go
```diff
@@ -323,8 +324,9 @@ func (d *Daemon) sharedErofsUmount(rafs *Rafs) error {
 		return errors.Wrapf(err, "unbind blob %s", d.ID())
 	}
 	domainID := rafs.Annotations[AnnoFsCacheDomainID]
+	fscaheID := rafs.Annotations[AnnoFsCacheID]
```
changweige, line 327: It's better to handle the absence of fscacheid. Moreover, should the name be consistent? I saw it is named blobID in function `UnbindBlob` body

pkg/daemon/daemon.go
changweige, line 339: Please also comment the reason why bootstrap is the first stuff to be deleted/unbind from nydusd

pkg/daemon/daemon.go
```diff
@@ -333,6 +335,11 @@ func (d *Daemon) sharedErofsUmount(rafs *Rafs) error {
 		return errors.Wrapf(err, "umount erofs %s mountpoint, %s", err, mountpoint)
 	}
 
+	// delete fscache bootstrap cache file
+	if err := c.UnbindBlob("", fscaheID); err != nil {
+		log.L.Warnf("delete bootstrap %s err %s", fscaheID, err)
+	}
```
changweige, line 341: Why ignoring the unbind error here?

pkg/filesystem/fs/fs.go
changweige: But we still have to return once GetClient() fails, otherwise, the whole process panics when performing L345

changweige: Please don't ignore errors and assume that a function must run in a separate goroutine.

pkg/daemon/client.go
changweige: We have to consider the API compatibility here. Old nydusd won't extract "blob_id' from the request, right? So old nydusd still can work with this nydus-snapshotter?

changweige:
Can you please this PR to latest main branch? The base code has changed. It will be easier to view those changes.

pkg/filesystem/fs/fs.go
kevinXYin: sorry for confusion, actually my last reply is for the below comment~ . Yes, we can not ignore the err for getClinet. I mean we delete cache in a goroutine , then if get error and make cache residue, the error will not pass to containerd. so maybe we can just ignore the error if can not delete cache. I will update the code firstly , then we can move on.

pkg/daemon/client.go
kevinXYin: Already add some comment for the API , and this new semantic is just used to delete fscache cache files, if the old nydusd did not support this then snapshotter will just ignore the error and print error info, which is as expectation. so I think we do not need compatible with an old nydusd.

pkg/daemon/daemon.go
kevinXYin: fscacheID will be set with domainID together, it can not be absent. fscacheID will used for generate fscache cookie key for bootstrap, and we need it when delete fscache cache file for bootstrap. Already add comment for this.

pkg/daemon/daemon.go
kevinXYin: two reasons for this:
1. make  compatible with old nydusd.
2. Is this need to return err to containerd when delete bootstrap cache failed? seems we do not do this for blob cache

pkg/daemon/client.go
```diff
@@ -264,9 +264,24 @@ func (c *nydusdClient) BindBlob(daemonConfig string) error {
...
 }
 
-func (c *nydusdClient) UnbindBlob(domainID string) error {
+/*
+ * Delete /api/v2/blobs implements different functions according to different parameters
+ * 1. domainID , delete all blob entries in the domain.
+ * 2. domainID + blobID, delete the blob entry, if the blob is bootstrap
+ *    also delete blob entries belong to it.
+ * 3. blobID, try to find and cull blob cache files by blobID in all domains.
+ */
```
changweige, line 273: Just a nit:  `//`  comment style might match the project comment style

pkg/daemon/daemon.go
changweige: Yes. It's a way to handle compatibility, however, it also hides the error if the nydusd supports this new API. Let's do it this way right now.

changweige (approved):
LGTM

PR #265, opened 2022-11-28
converter: enhance nydus tar unpack

imeoer:
Unpacking file from nydus blob should seek until encountering the head
of blob stream, this patch ensures this to improve unpack compatibility
and refine the test codes.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -113,14 +113,14 @@ func unpackOciTar(ctx context.Context, dst string, reader io.Reader) error {
 }
 
 // Unpack a Nydus formatted tar stream into a directory.
-func unpackNydusTar(bootDst, blobDst string, ra content.ReaderAt) error {
+func unpackNydusBlob(bootDst, blobDst string, ra content.ReaderAt) error {
 	boot, err := os.OpenFile(bootDst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
 	if err != nil {
 		return errors.Wrapf(err, "write to bootstrap %s", bootDst)
 	}
 	defer boot.Close()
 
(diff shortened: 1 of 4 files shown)


pkg/converter/convert_unix.go
```diff
@@ -130,134 +130,76 @@ func unpackNydusTar(bootDst, blobDst string, ra content.ReaderAt) error {
...
-			break
-		}
-
 		// Try to seek to the part of tar header.
-		var err error
-		cur, err = reader.Seek(cur-headerSize, io.SeekStart)
+		_, err := reader.Seek(cur, io.SeekStart)
 		if err != nil {
-			return errors.Wrapf(err, "seek to %d for tar header", cur-headerSize)
+			return errors.Wrapf(err, "seek to %d for nydus tar header", cur)
```
changweige, line 161: seek %d

pkg/converter/convert_unix.go
```diff
@@ -130,134 +130,76 @@ func unpackNydusTar(bootDst, blobDst string, ra content.ReaderAt) error {
 	}
 	defer blob.Close()
 
-	if err = unpackBlobFromNydusTar(ra, blob); err != nil {
+	if err = unpackFileFromNydusTar(ra, blobNameInTar, blob); err != nil {
```
changweige, line 133: Can we rename NydusTar to NydusArchive which archives both blob and bootstrap borrowing the tar header structure describing them

changweige (commented):
Otherwise, looks good

pkg/converter/convert_unix.go
imeoer: Fixed.

pkg/converter/convert_unix.go
imeoer: Fixed.

changweige (approved):

changweige:
Smoke failed

PR #268, opened 2022-11-30
remove scheme field in nydusd config

imeoer:
remove scheme field in nydusd config

We have implemented automatically scheme detect for nydusd in this PR:

https://github.com/dragonflyoss/image-service/pull/893#merged-event

So we can remove the default scheme field in nydusd config file.

### DIFF
--- a/cmd/containerd-nydus-grpc/app/snapshotter/testdata/happypath/config.json
+++ b/cmd/containerd-nydus-grpc/app/snapshotter/testdata/happypath/config.json
@@ -3,7 +3,6 @@
     "backend": {
       "type": "registry",
       "config": {
-        "scheme": "https",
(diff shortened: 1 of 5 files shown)


changweige:
Seems no logic changes in this change set?

liubogithub:
Looks good for the 1st commit.

changweige:
We need to discuss the arrangement of currently scattered documents. I am inclining to merge this change until the nydusd (v2.2) including automatically changing HTTP scheme is released which means it can actually be used.

changweige:
Needs rebase. Can we only submit the changes for reference nydusd config file in this PR and clean up the docs in the next step?

changweige (approved):

PR #272, opened 2022-12-05
feat: add some nydusd-config for metrics work

Desiki-high:
Fix the issue#271 , just add the two fileds to the nudusd config,local test the containerd-nydus-grpc works well.
image
Also, this is my first PR, so I am not sure the work is right, It look like so simple, but I can't find any other correlation that needs to be modified.


changweige (approved):
LGTM~ Thank you.

PR #280, opened 2022-12-12
maintenance api server's default socket should be under root dir

changweige:
We should put all snapshotter's stuff under root dir which can be customized.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -110,7 +110,7 @@ type Config struct {
 	RotateLogMaxAge          int           `toml:"log_rotate_max_age"`
 	RotateLogLocalTime       bool          `toml:"log_rotate_local_time"`
 	RotateLogCompress        bool          `toml:"log_rotate_compress"`
-	APISocket                string        `toml:"api_socket"`
+	EnableSystemController   bool          `toml:"enable_system_controller"`
 	RecoverPolicy            string        `toml:"recover_policy"`
 }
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -205,7 +205,11 @@ func SetStartupParameter(startupFlag *command.Args, cfg *Config) error {
 	cfg.GCPeriod = d
 
 	cfg.Address = startupFlag.Address
-	cfg.APISocket = startupFlag.APISocket
+
+	if startupFlag.APISocket == "" {
```
imeoer, line 209: What will happen when `startupFlag.APISocket != ""` ?

config/config.go
```diff
@@ -205,7 +205,11 @@ func SetStartupParameter(startupFlag *command.Args, cfg *Config) error {
...
+		cfg.APISocket = path.Join(cfg.RootDir, "api.sock")
```
imeoer, line 210: It seems we should allow setting the `APISocket` to empty.

changweige: Do you mean completely shutdown api server?

imeoer: I mean how do users set it to empty?

imeoer:
It seems the default grpc address is `/run/containerd-nydus/containerd-nydus-grpc.sock`, should we also put the system management api into `/run/containerd-nydus` dir ?

changweige:
> It seems the default grpc address is `/run/containerd-nydus/containerd-nydus-grpc.sock`, should we also put the system management api into `/run/containerd-nydus` dir ?

Yes. It sounds a good idea

snapshot/snapshot.go
```diff
@@ -116,8 +116,8 @@ func NewSnapshotter(ctx context.Context, cfg *config.Config) (snapshots.Snapshot
 		}
 	}()
 
-	if cfg.APISocket != "" {
-		systemController, err := system.NewSystemController(manager, cfg.APISocket)
+	if cfg.EnableSystemController {
+		systemController, err := system.NewSystemController(manager, path.Join(cfg.RootDir, "system_controller.sock"))
```
imeoer, line 120: Maybe name it `supervisor.sock` ?

changweige: Um. We already have supervisors for nydusd daemon.  Or just rename it to system.sock ?snapshotter.sock ?

imeoer: `system.sock` is ok.

changweige: updated

imeoer (approved):

PR #284, opened 2022-12-13
Check KubeSecretListener start error

liubin:
Check if KubeSecretListener has start error, then return the error to prevent snapshotter to start.

Also this commit reduce a goroutine that is not required.


changweige (approved):
LGTM

PR #286, opened 2022-12-13
treewide: tidy up --config-path

hsiangkao:
1. --config-path "/etc/nydus/nydusd-config.fusedev.json" or
                 "/etc/nydus/nydusd-config.fscache.json"

2. Add a symlink /etc/nysud/nydusd-config.json to match the default configuration.

In order to match
https://github.com/dragonflyoss/image-service/pull/955

Signed-off-by: Gao Xiang \<hsiangkao@linux.alibaba.com\>

### DIFF
--- a/misc/snapshotter/nydus-snapshotter.fscache.service
+++ b/misc/snapshotter/nydus-snapshotter.fscache.service
@@ -6,7 +6,7 @@ Before=containerd.service
 [Service]
 Type=simple
 Environment=HOME=/root
-ExecStart=/usr/local/bin/containerd-nydus-grpc --fs-driver fscache --daemon-mode shared --config-path /etc/nydus/config.json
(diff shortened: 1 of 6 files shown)


misc/snapshotter/nydus-snapshotter.fscache.service
```diff
@@ -6,7 +6,7 @@ Before=containerd.service
 [Service]
 Type=simple
 Environment=HOME=/root
-ExecStart=/usr/local/bin/containerd-nydus-grpc --fs-driver fscache --daemon-mode shared --config-path /etc/nydus/config.json
+ExecStart=/usr/local/bin/containerd-nydus-grpc --fs-driver fscache --daemon-mode shared --config-path /etc/nydus/nydusd-config.fscache.json
```
changweige, line 9: The nydusd config file is better to match what was installed by `make install`. So better use `nydusd-config.json` here?

hsiangkao: I think /etc/nydus/nydusd-config.json is just used for the default use.

imeoer (approved):

PR #289, opened 2022-12-14
Don't return error if test pod is not created or has been deleted

liubin:

### DIFF
--- a/.github/workflows/k8s-e2e.yml
+++ b/.github/workflows/k8s-e2e.yml
@@ -143,7 +143,7 @@ jobs:
           docker exec kind-control-plane journalctl --no-pager -u containerd >> $log_dir/containerd.log
           docker exec kind-control-plane journalctl --no-pager -u kubelet >> $log_dir/kubelet.log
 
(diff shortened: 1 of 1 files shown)


.github/workflows/k8s-e2e.yml
```diff
@@ -143,7 +143,7 @@ jobs:
           docker exec kind-control-plane journalctl --no-pager -u containerd >> $log_dir/containerd.log
           docker exec kind-control-plane journalctl --no-pager -u kubelet >> $log_dir/kubelet.log
 
-          kubectl get pod test-pod -o yaml >> $log_dir/test-pod.log
+          kubectl get pod test-pod -o yaml >> $log_dir/test-pod.log || echo "test-pod may be deleted or not created"
```
changweige, line 146: Should exit from the script once any command fails by `set -e` ?

liubin: This step is for getting more logs, so it should continue even with errors.

imeoer (approved):

changweige (approved):

PR #282, opened 2022-12-12
Support cri image service proxy mode for getting registry auth

darfux:
Support cri based authentication like stargz-snapshotter (https://github.com/containerd/stargz-snapshotter/blob/main/docs/overview.md#cri-based-authentication). The proxy service uses the implementation by stargz directly.

Fixes: #47

Signed-off-by: Li Yuxuan <liyuxuan.darfux@bytedance.com>

PR #290, opened 2022-12-15
test:e2e: Add cri auth test

darfux:
Make the original k8s-e2e workflow to be a reusable workflow template, which is used by cri case and kubeconf case with different `auth-type` input.

Differences between cri case and kubeconf case:
- CRI snapshotter spec needs extra containerd socket mount
- CRI snapshotter spec has no cluster role for secrets
- CRI snapshotter spec uses `--enable-cri-keychain` instead of `--enable-kubeconfig-keychain`
- In the `Run E2E test` step, cri case needs to change the image service endpoint of kubelet to nydus-snapshotter

Other changes:
- Add `imagePullSecrets` to test pod spec and change it's namespace to `nydus-system`. So that kubelet can get the image secret under both cri&kubeconf cases. The auth info in the containerd config file is no longer needed and has been removed.
- Add a `ps -ef` info log in the `Dump logs` step

Signed-off-by: Li Yuxuan <liyuxuan.darfux@bytedance.com>


darfux:
Add e2e test for PR https://github.com/containerd/nydus-snapshotter/pull/282 /cc @changweige

changweige:
@liubin Do you have time to take a look at this PR too?

tests/e2e/k8s/test-pod.yaml.tpl
```diff
@@ -10,3 +11,5 @@ spec:
       command: ["sh", "-c"]
       args:
         - tail -f /dev/null
+  imagePullSecrets:
+    - name: regcred
```
liubin, line 15: Add a new line at the end, otherwise, GitHub will show a `No newline at end of file` message.

tests/e2e/k8s/snapshotter-cri.yaml
```diff
@@ -0,0 +1,121 @@
...
+    args:
+      - containerd-nydus-grpc
+      - --nydusd-path /usr/local/bin/nydusd
+      - --config-path /etc/nydus/config.json
+      - --root /var/lib/containerd-nydus
+      - --address /run/containerd-nydus/containerd-nydus-grpc.sock
+      - --log-level debug
+      - --daemon-mode shared
+      - --log-to-stdout
+      - --enable-cri-keychain
```
liubin, line 37: Is this `--enable-cri-keychain` needed? this flag will enable Kubernetes secrets listener to watch secrects. For `cri` type, is it needed?

liubin (approved):

.github/workflows/k8s-e2e.yml
```diff
@@ -91,10 +89,8 @@ jobs:
 
           [plugins."io.containerd.grpc.v1.cri".registry.mirrors."${registry_ip}:5000"]
               endpoint = ["http://${registry_ip}:5000"]
-          [plugins."io.containerd.grpc.v1.cri".registry.configs."${registry_ip}:5000".auth]
-            username = "${{ env.DOCKER_USER }}"
-            password = "${{ env.DOCKER_PASSWORD }}"
```
liubin, line 96: Is this not required?
I see the CI passed, but it seems both `cri` and `kubeconf` are useing the same node to run test. 
You added `image-service-endpoint` to kubelet, but not delete it when running `kubeconfg` test.

darfux: From [using-jobs-in-a-workflow](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#overview) and [using-a-github-hosted-runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#using-a-github-hosted-runner):

> Each job runs in a runner environment specified by runs-on.
> When the job begins, GitHub automatically provisions a new VM for that job.

I think each job will run in a standalone runner environment(VM) and don't share any data 🤔

> Is this not required?

Under kubeconf case, kubelet can get the image secret from `imagePullSecrets` as well and pass it to containerd now.

darfux: I‘ve tried to run a single kubeconf case: https://github.com/darfux/nydus-snapshotter/actions/runs/3701944536/jobs/6271709388 and it seems still OK. @liubin FYI😀

changweige (approved):

PR #293, opened 2022-12-22
converter: remove pkg/errdefs dependence

imeoer:
To reduce a external package for the users.


changweige (approved):
LGTM

PR #294, opened 2022-12-23
converter: add stream mode to `Unpack`

loheagn:
This patch adds stream mode to `Unpack`, which avoids to unpack the blob data to the local disk but sets up a local http server to proxy the blob data reading.

### DIFF
--- a/pkg/converter/cs_proxy_unix.go
+++ b/pkg/converter/cs_proxy_unix.go
@@ -1,0 +1,164 @@
+//go:build !windows
+// +build !windows
+
+/*
+ * Copyright (c) 2023. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package converter
+
+import (
+	"archive/tar"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/containerd/containerd/content"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type contentStoreProxy struct {
+	socketPath string
+	server     *http.Server
+}
+
+func setupContentStoreProxy(workDir string, ra content.ReaderAt) (*contentStoreProxy, error) {
+	sockP, err := os.CreateTemp(workDir, "nydus-cs-proxy-*.sock")
+	if err != nil {
+		return nil, errors.Wrap(err, "create unix socket file")
+	}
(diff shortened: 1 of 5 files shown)


pkg/converter/cs_proxy_unix.go
```diff
@@ -0,0 +1,159 @@
...
+
+type contentStoreProxy struct {
+	socketPath string
+	server     *http.Server
+}
+
+func setupContentStoreProxy(workDir string, ra content.ReaderAt) (*contentStoreProxy, error) {
+	sockP, err := os.CreateTemp(workDir, "nydus-cs-proxy-*.sock")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create unix socket file")
```
imeoer, line 34: The wrap message only mentions the action generally, just saying `create unix socket file` be better.

pkg/converter/cs_proxy_unix.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to listen unix socket when setup content store proxy")
+	}
+
+	server := &http.Server{
+		Handler: contentProxyHandler(ra),
+	}
+
+	go func() {
+		_ = server.Serve(listener)
```
imeoer, line 49: The error should be handled, at least printing an error message?

pkg/converter/cs_proxy_unix.go
```diff
@@ -0,0 +1,159 @@
...
+		socketPath: sockP.Name(),
+		server:     server,
+	}, nil
+}
+
+func (p *contentStoreProxy) close() error {
+	if err := p.server.Close(); err != nil {
+		return err
+	}
+	return os.Remove(p.socketPath)
```
imeoer, line 62: The `Remove` should be put into `defer`?

pkg/converter/cs_proxy_unix.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func parseRangeHeader(rangeStr string, totalLen int64) (start, wantedLen int64, err error) {
+	rangeList := strings.Split(rangeStr, "-")
+	start, err = strconv.ParseInt(rangeList[0], 10, 64)
+	if err != nil {
+		errors.Wrap(err, "failed to parse range header")
+		return
```
imeoer, line 70: Should be `err = errors.Wrap(err, "parse range header")` ?

imeoer (approved):
LGTM!

imeoer:
cc @changweige @sctb512

PR #295, opened 2022-12-26
converter: support more options

imeoer:
The converter package will be imported in nydusify.


pkg/backend/oss.go
```diff
@@ -63,114 +61,104 @@ func newOSSBackend(rawConfig []byte) (*OSSBackend, error) {
...
-	if chunkNum <= 0 || chunkNum > 10000 {
-		return nil, errors.New("chunkNum invalid")
+func splitFileByPartSize(blobSize, chunkSize int64) ([]oss.FileChunk, error) {
+	if chunkSize <= 0 {
+		return nil, errors.New("invalid chunk size")
 	}
 
-	if int64(chunkNum) > blobSize {
-		return nil, errors.New("oss: chunkNum invalid")
+	var chunkN = blobSize / chunkSize
```
changweige, line 73: It will leak the last chunk

pkg/backend/oss.go
```diff
@@ -63,114 +61,104 @@ func newOSSBackend(rawConfig []byte) (*OSSBackend, error) {
...
+	if chunkN >= 10000 {
+		return nil, errors.New("too many parts, please increase chunk size")
```
changweige, line 75: Is there any principle why 10000 is chosen as the threshold?

changweige, line 75: How user can increase chunks size, it seems hard-coded as 500MB?

pkg/backend/oss.go
```diff
@@ -63,114 +61,104 @@ func newOSSBackend(rawConfig []byte) (*OSSBackend, error) {
...
 	}
 
 	var chunks []oss.FileChunk
 	var chunk = oss.FileChunk{}
-	var chunkN = (int64)(chunkNum)
 	for i := int64(0); i < chunkN; i++ {
 		chunk.Number = int(i + 1)
```
changweige, line 81: Why is there a offset `1` between chunk number and the index?

pkg/converter/types.go
```diff
@@ -66,6 +66,10 @@ type PackOption struct {
 	Compressor string
 	// OCIRef enables converting OCI tar(.gz) blob to nydus referenced blob.
 	OCIRef bool
+	// AlignedChunk aligns uncompressed data chunks to 4K, only for RAFS V5.
```
changweige, line 69: Why do we need chunks to be aligned by 4K for rafs v5?

pkg/backend/oss.go
imeoer: It will be appended later.

pkg/backend/oss.go
imeoer: The code is copied from aliyun-oss-go-sdk, I will add a port comment.

pkg/backend/oss.go
imeoer: It's defined by `oss.FileChunk`.

pkg/converter/types.go
imeoer: It's compatible for `nydusify convert --aligned-chunk`.

pkg/backend/oss.go
changweige: Can we have a comment telling the trick at least?

pkg/converter/types.go
changweige: I think we don't have to expose the parameter `aligned-chunk`. Rafs v5 does not have to be aligned and rafs v6 must be aligned. Perhaps, the parameter can be removed both from nydusify and nydus-image

pkg/backend/oss.go
imeoer: Updated, users can set `backend.MultipartChunkSize = X` to change the chunk size.

pkg/backend/oss.go
imeoer: It's ported from https://github.com/aliyun/aliyun-oss-go-sdk/blob/v2.2.6/oss/utils.go#L259, leave the port code as unchanged as possible.

pkg/converter/types.go
imeoer: We still need to keep the compatibility, at least it doesn't have any broken changes for minor release version, right?

changweige: Um. there is no compatibility problem no matter whether aligned-chunk is exposed or not.

imeoer: The converter package will be used in nydusify v2.2.

changweige (approved):
LGTM

PR #296, opened 2022-12-29
snapshot:remoteMounts: Load snapshot config under shared mode

darfux:
When shared daemon is enabled, snapshot config uses `<daemon id>/<snapshot id>/config.json` now.

Signed-off-by: Li Yuxuan <liyuxuan.darfux@bytedance.com>


changweige (approved):
LGTM, thanks

PR #297, opened 2022-12-29
docs: add docs for running nydus snapshotter in Kubernetes

liubin:
This new document will describe how to create a local Kubernetes cluster using kind and then deploy nydus snapshotter in the Kubernetes cluster.

### DIFF
--- a/docs/run_nydus_in_kubernetes.md
+++ b/docs/run_nydus_in_kubernetes.md
@@ -1,0 +1,128 @@
+# Run Nydus snapshotter in Kubernetes
+
+This document will introduce how to run Nydus snapshotter in Kubernetes cluster, you can use helm to deploy Nydus snapshotter container.
+
+**NOTE:** This document is mainly to allow everyone to quickly deploy and experience Nydus snapshotter in the Kubernetes cluster. You cannot use it as a deployment and configuration solution for the production environment.
+
+## Setup Kubernetes using kind
+
+[kind](https://kind.sigs.k8s.io/) is a tool for running local Kubernetes clusters using Docker container “nodes”.
+kind was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
+
+First, we need to prepare a configuration file(`kind-config.yaml`) for kind to specify the devices and files we need to mount to the kind node.
+
+```yaml
+kind: Cluster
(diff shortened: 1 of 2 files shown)


docs/run_nydus_in_kubernetes.md
```diff
@@ -0,0 +1,128 @@
...
+
+```toml
+version = 2
+[debug]
+  level = "debug"
+
+[plugins."io.containerd.grpc.v1.cri".containerd]
+  discard_unpacked_layers = false
+  disable_snapshot_annotations = false
+  snapshotter = "overlayfs"
```
changweige, line 39: Should we change the CRI snapshotter to `nydus`?

docs/run_nydus_in_kubernetes.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+Use `kubectl` to create the Pod.
+
+```bash
+$ kubectl create -f nydus-pod.yaml
+$ kubectl get pods -w
+```
+
+The `-w` options will blcok the console and wait for the changes of the status of the pod. If everything is normal, you can see that the pod will become `Running` after a while.
```
changweige, line 123: typo `block`

liubin: fixed

docs/run_nydus_in_kubernetes.md
liubin: Can we do that? First, we should pull the nydus-snapshotter image and start the container, then the nydus snapshotter can be used.

changweige: I see. Nydus-snapshotter is deployed as DaemonSet and packed in an OCI container image.
Is it the same containerd pulling  OCI image packing nydus-snapshotter and nydus image?
If yes, I don't understand how the containerd can choose nydus as snapshotter when pulling the nydus image `ghcr.io/dragonflyoss/image-service/nginx:nydus-latest` since containerd can only configure statically the snapshotter type here

liubin: In the current helm chart, the snapshotter will be pulled and unpacked by overlayfs, and then change the global snapshotter to nydus and restart the containerd after the snapshotter started.

changweige (approved):
LGTM

PR #298, opened 2022-12-29
metrics: support HTTP metrics server

sctb512:
The common way to display prometheus format data is through the HTTP server. For instance, [Grafana](https://grafana.com/grafana) does not support collecting metrics from `unix domain socket`. Therefore, we should change the `unix domain socket` server to a HTTP server.

Here is an example:
```shell
containerd-nydus-grpc --config-path /etc/nydus/config.json -metrics-address :9010 --log-to-stdout
curl http://localhost:9010/metrics
```
Then, you will see the prometheus metrics.

### DIFF
--- a/cmd/containerd-nydus-grpc/pkg/command/flags.go
+++ b/cmd/containerd-nydus-grpc/pkg/command/flags.go
@@ -39,7 +39,7 @@ type Args struct {
 	DaemonMode               string `toml:"daemon_mode"`
 	FsDriver                 string `toml:"fs_driver"`
 	SyncRemove               bool   `toml:"sync_remove"`
-	EnableMetrics            bool   `toml:"enable_metrics"`
+	MetricsAddress           string `toml:"metrics_address"`
 	MetricsFile              string `toml:"metrics_file"`
 	EnableStargz             bool   `toml:"enable_stargz"`
 	DisableCacheManager      bool   `toml:"disable_cache_manager"`
@@ -111,10 +111,11 @@ func buildFlags(args *Args) []cli.Flag {
 			Usage:       "whether to disable blob cache manager",
 			Destination: &args.DisableCacheManager,
 		},
-		&cli.BoolFlag{
-			Name:        "enable-metrics",
-			Usage:       "whether to collect metrics",
-			Destination: &args.EnableMetrics,
+		&cli.StringFlag{
+			Name:        "metrics-address",
+			Value:       "",
+			Usage:       "Enable metrics server by setting to an `ADDRESS` such as \"localhost:8080\", \":8080\"",
+			Destination: &args.MetricsAddress,
 		},
 		&cli.BoolFlag{
 			Name:        "enable-nydus-overlayfs",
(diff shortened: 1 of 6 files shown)


changweige:
Please fix linting warnings.

pkg/metrics/listener.go
```diff
@@ -9,21 +9,35 @@ package metrics
...
 	}
-	ln, err := net.Listen("unix", addr)
+
+	http.Handle(endpointPromMetrics, promhttp.HandlerFor(registry.Registry, promhttp.HandlerOpts{
+		ErrorHandling: promhttp.HTTPErrorOnError,
+	}))
+
+	log.L.Infof("Start metrics HTTP server on %s", addr)
+
+	listener, err := net.Listen("tcp4", addr)
```
changweige, line 34: No need to restrict it to "tcp4"

sctb512: Yes, fixed.

cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -111,10 +111,11 @@ func buildFlags(args *Args) []cli.Flag {
...
 			Destination: &args.DisableCacheManager,
 		},
-		&cli.BoolFlag{
-			Name:        "enable-metrics",
-			Usage:       "whether to collect metrics",
-			Destination: &args.EnableMetrics,
+		&cli.StringFlag{
+			Name:        "metrics-address",
+			Value:       "",
+			Usage:       "metrics `ENDPOINT` does not listen by default",
```
imeoer, line 117: Maybe `Enable metrics server by setting to an address such as "localhost:8080", ":8080"`.

sctb512: Yes, It is more reasonable. Fixed.

pkg/metrics/listener.go
```diff
@@ -8,22 +8,31 @@ package metrics
...
+	"github.com/containerd/containerd/log"
+	"github.com/containerd/nydus-snapshotter/pkg/metrics/registry"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-// DefaultBindAddress sets the default bind address for the metrics
-// listener.
-var DefaultBindAddress = ":8080"
+// Endpoint for prometheus metrics
+var endpointPromMetrics = "/metrics"
```
changweige, line 19: Should we change it to "/v1/metrics" like containerd ?
```
### Metrics API

The metrics API that outputs prometheus style metrics will be versioned independently,
prefixed with the API version. i.e. `/v1/metrics`, `/v2/metrics`.

The metrics API version will be incremented when breaking changes are made to the prometheus
output. New metrics can be added to the output in a backwards compatible manner without
bumping the API version.

```

changweige:
Otherwise, LGTM

pkg/metrics/listener.go
sctb512: Yes, fixed it.

changweige (approved):
LGTM

PR #301, opened 2022-12-30
config: add warmup option for nydus daemon

sctb512:
This is an experimental feature. With the warmup option, the nydus daemon could cache more data on the local disk to increase hit ratio.


config/daemonconfig/fuse.go
```diff
@@ -27,6 +27,7 @@ type FuseDaemonConfig struct {
 	EnableXattr     bool          `json:"enable_xattr,omitempty"`
 	AccessPattern   bool          `json:"access_pattern,omitempty"`
 	LatestReadFiles bool          `json:"latest_read_files,omitempty"`
+	Warmup          uint64        `json:"warmup"`
```
changweige, line 30: Please also add a comment line here to tell it's experimental.

sctb512: Got it. Fixed.

changweige (approved):
LGTM

PR #302, opened 2022-12-30
metrics: add nydus daemon version

sctb512:
Add nydus daemon version to prometheus format metrics.

```=
# HELP nydusd_version_string The version of nydus daemon.
# TYPE nydusd_version_string counter
nydusd_version_string{version="v2.1.0-rc.3.1-473-gf4dd9e1c"} 1
```

Moreover, remove some metrics that have been deleted in nydus and improve descriptions for daemon FS metrics.

### DIFF
--- a/pkg/metrics/collector/collector.go
+++ b/pkg/metrics/collector/collector.go
@@ -30,6 +30,10 @@ func NewFsMetricsCollector(m *types.FsMetrics, imageRef string) *FsMetricsCollec
 	return &FsMetricsCollector{m, imageRef}
 }
 
+func NewDaemonInfoCollector(version *types.BuildTimeInfo, value float64) *DaemonInfoCollector {
+	return &DaemonInfoCollector{version, value}
+}
+
 func NewSnapshotterMetricsCollector(ctx context.Context, cacheDir string, pid int) (*SnapshotterMetricsCollector, error) {
 	currentStat, err := tool.GetProcessStat(pid)
 	if err != nil {
--- a/pkg/metrics/collector/daemon.go
+++ b/pkg/metrics/collector/daemon.go
@@ -7,6 +7,8 @@
 package collector
 
 import (
+	"github.com/containerd/containerd/log"
+	"github.com/containerd/nydus-snapshotter/pkg/daemon/types"
 	"github.com/containerd/nydus-snapshotter/pkg/metrics/data"
 )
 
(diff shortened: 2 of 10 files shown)


pkg/metrics/data/daemon.go
```diff
@@ -20,4 +21,11 @@ var (
 		},
 		[]string{nydusdEventLabel},
 	)
+	NydusdVersion = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "nydusd_version_string",
```
changweige, line 26: It would be better to rename the metric name to "nydusd_count", version is just is label of the metircs.

pkg/metrics/data/fs.go
```diff
@@ -74,8 +55,8 @@ var MetricHists = []*mtypes.MetricHistogram{
 
 	{
 		Desc: prometheus.NewDesc(
-			"nydusd_fop_hit_hist",
-			"File operations histogram",
+			"nydusd_file_operation_hit_hist",
```
changweige, line 58: Could we also remove the suffix `hist`?

pkg/metrics/collector/daemon.go
```diff
@@ -7,13 +7,22 @@
...
+type DaemonInfoCollector struct {
+	Metrics *types.DaemonInfo
+}
+
 func (d *DaemonEventCollector) Collect() {
 	data.NydusdEventCount.WithLabelValues(d.event).Inc()
 }
+
+func (d *DaemonInfoCollector) Collect() {
+	data.NydusdVersion.WithLabelValues(d.Metrics.Version.PackageVer).Inc()
```
changweige, line 27: Perhaps, we should use Gauge instead. When snapshots are removed, nydusd stops, and the number of running nydusd decreases. In addition, we should also consider the gauge should change when a nydusd dies in an accident and when nydus-snapshotter restarts, it re-manage the running nydusd daemons.

pkg/metrics/collector/collector.go
```diff
@@ -30,6 +30,10 @@ func NewFsMetricsCollector(m *types.FsMetrics, imageRef string) *FsMetricsCollec
 	return &FsMetricsCollector{m, imageRef}
 }
 
+func NewDaemonInfoCollector(version *types.BuildTimeInfo, countType NydusDaemonCountType) *DaemonInfoCollector {
```
changweige, line 33: How about redesigning the interface to `func NewDaemonInfoCollector(version *types.BuildTimeInfo, value float64) *DaemonInfoCollector`, where value can be negative

Then we can directly call `Add()` of gauge type

sctb512: It is more reasonable. I have finished it.

changweige (approved):
LGTM

PR #311, opened 2023-01-11
command: remove unused flag metrics-file

sctb512:
There is enough to export metrics via HTTP port. Therefore, The flag to set metrics file path, `metrics-file`, is unused. We could remove it.


changweige (approved):
LGTM

PR #312, opened 2023-01-12
misc: fix 404 for snapshotter dockerfile

sctb512:
The github workflow frequently causes errors to break tests:
```
Step 4/18 : RUN export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")') &&     wget https://github.com/dragonflyoss/image-service/releases/download/$NYDUS_VERSION/nydus-static-$NYDUS_VERSION-linux-amd64.tgz &&     echo $NYDUS_VERSION > /.nydus_version &&     tar xzf nydus-static-$NYDUS_VERSION-linux-amd64.tgz &&     rm nydus-static-$NYDUS_VERSION-linux-amd64.tgz
 ---> Running in 39df2748e06d
Connecting to github.com (140.82.112.4:443)
wget: server returned error: HTTP/1.1 404 Not Found
The command '/bin/sh -c export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")') &&     wget https://github.com/dragonflyoss/image-service/releases/download/$NYDUS_VERSION/nydus-static-$NYDUS_VERSION-linux-amd64.tgz &&     echo $NYDUS_VERSION > /.nydus_version &&     tar xzf nydus-static-$NYDUS_VERSION-linux-amd64.tgz &&     rm nydus-static-$NYDUS_VERSION-linux-amd64.tgz' returned a non-zero code: 1
```

I guess the reason for this situation is the `$NYDUS_VERSION` is an invalid value.
This PR tries to verify it by adding an `echo` command.

### DIFF
--- a/misc/snapshotter/Dockerfile
+++ b/misc/snapshotter/Dockerfile
@@ -2,7 +2,10 @@ FROM alpine:3.17.0 AS sourcer
 
 RUN apk add --no-cache curl
 RUN apk add --no-cache --upgrade grep
-RUN export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")') && \
+RUN unset NYDUS_VERSION && RETRY_COUNT=0 && \
+    while [ -z $NYDUS_VERSION ] && [ $RETRY_COUNT != 600 ]; do \
+    export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")'); RETRY_COUNT=$(($RETRY_COUNT+1)); sleep 1; done && \
+    echo nydus version: $NYDUS_VERSION after $RETRY_COUNT times && if [ -z $NYDUS_VERSION ]; then echo Invalid NYDUS_VERSION && exit -1; fi && \
     wget https://github.com/dragonflyoss/image-service/releases/download/$NYDUS_VERSION/nydus-static-$NYDUS_VERSION-linux-amd64.tgz && \
     echo $NYDUS_VERSION > /.nydus_version && \
     tar xzf nydus-static-$NYDUS_VERSION-linux-amd64.tgz && \


sctb512:
Yes, that is right. The version in a null value.

```
version:
Connecting to github.com (140.82.114.4:443)
wget: server returned error: HTTP/1.1 404 Not Found
```

fengshunli (commented):
This needs to support arm64

sctb512:
> This needs to support arm64

Do you mean that we should add a variable to control the architecture?
```
nydus-static-$NYDUS_VERSION-linux-$NYDUS_ARCHITECTURE.tgz
```

fengshunli:
> 

yes, add ARG TARGETARCH

fengshunli:
> > This needs to support arm64
> 
> Do you mean that we should add a variable to control the architecture?
> 
> ```
> nydus-static-$NYDUS_VERSION-linux-$NYDUS_ARCHITECTURE.tgz
> ```

Yes, add add ARG TARGETARCH environment variable

sctb512:
Done. PTAL @fengshunli

fengshunli:
> Done. PTAL @fengshunli
Refer to ARG TARGETARCH way to achieve

sctb512:
> refer to ARG TARGETARCH way to achieve

Fixed.

fengshunli:
LGTM

liubin:
It may be better to add a `sleep 1` in one loop, because GitHub may limit the request frequency, retry immediately will continue to get a 403 error.

changweige (commented):
I am wondering why `NYDUS_VERSION` is empty.

changweige:
> It may be better to add a `sleep 1` in one loop, because GitHub may limit the request frequency, retry immediately will continue to get a 403 error.

Agree. It might because nydus-snapshotter is under Containerd org where a lot tests are running at the same time

liubin:
> NYDUS_VERSION



> > It may be better to add a `sleep 1` in one loop, because GitHub may limit the request frequency, retry immediately will continue to get a 403 error.
> 
> Agree. It might be because nydus-snapshotter is under Containerd org where a lot of tests are running at the same time

Yes, if the user exceeded the limit, will get a 403 error, so the NYDUS_VERSION will be empty.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit

sctb512:
It seems that the `ARG TARGETARCH` not work. I will fix it later.

changweige:
> It seems that the `ARG TARGETARCH` not work. I will fix it later.

I didn't catch on specifying the target architecture in Dockerfile. Do we want a cross-platform build here? Who is specifying the docker build argument? I guess amd64 should meet the test requirements for now. In fact, nydus-snapshotter never releases platforms rather than amd64

fengshunli:
> > It seems that the `ARG TARGETARCH` not work. I will fix it later.
> 
> I didn't catch on specifying the target architecture in Dockerfile. Do we want a cross-platform build here? Who is specifying the docker build argument? I guess amd64 should meet the test requirements for now. In fact, nydus-snapshotter never releases platforms rather than amd64

no need to specify, the base image supports multiple architectures and is directly compatible

changweige:
> It seems that the `ARG TARGETARCH` not work. I will fix it later.



> > > It seems that the `ARG TARGETARCH` not work. I will fix it later.
> > 
> > 
> > I didn't catch on specifying the target architecture in Dockerfile. Do we want a cross-platform build here? Who is specifying the docker build argument? I guess amd64 should meet the test requirements for now. In fact, nydus-snapshotter never releases platforms rather than amd64
> 
> no need to specify, the base image supports multiple architectures and is directly compatible

Seems we need to enable `buildkit` to leverage the automatically defined build arguments.

fengshunli:
> > It seems that the `ARG TARGETARCH` not work. I will fix it later.
> 
> > > > It seems that the `ARG TARGETARCH` not work. I will fix it later.
> > > 
> > > 
> > > I didn't catch on specifying the target architecture in Dockerfile. Do we want a cross-platform build here? Who is specifying the docker build argument? I guess amd64 should meet the test requirements for now. In fact, nydus-snapshotter never releases platforms rather than amd64
> > 
> > 
> > no need to specify, the base image supports multiple architectures and is directly compatible
> 
> Seems we need to enable `buildkit` to leverage the automatically defined build arguments.

Just change docker build to docker buildx build @sctb512 @changweige

sctb512:
> Just change docker build to docker buildx build

It works. But maybe there is no need to use `TARGETARCH`. This will cause errors when the `nydus-static-$NYDUS_VERSION-linux-$TARGETARCH.tgz` is not exist. `amd64` is enough for testing.

fengshunli:
> > Just change docker build to docker buildx build
> 
> It works. But maybe there is no need to use `TARGETARCH`. This will cause errors when the `nydus-static-$NYDUS_VERSION-linux-$TARGETARCH.tgz` is not exist. `amd64` is enough for testing.

You have to consider the situation of arm. If you develop on an arm machine, the amd architecture will report an error

sctb512:
> > > Just change docker build to docker buildx build
> > 
> > 
> > It works. But maybe there is no need to use `TARGETARCH`. This will cause errors when the `nydus-static-$NYDUS_VERSION-linux-$TARGETARCH.tgz` is not exist. `amd64` is enough for testing.
> 
> You have to consider the situation of arm. If you develop on an arm machine, the amd architecture will report an error

Yes, It is more reasonable to consider the situation of arm.  PTAL @fengshunli cc @liubin @changweige

changweige:
> > > Just change docker build to docker buildx build
> > 
> > 
> > It works. But maybe there is no need to use `TARGETARCH`. This will cause errors when the `nydus-static-$NYDUS_VERSION-linux-$TARGETARCH.tgz` is not exist. `amd64` is enough for testing.
> 
> You have to consider the situation of arm. If you develop on an arm machine, the amd architecture will report an error

At present, k8s e2e test can only run on Github action, so we don't have to worry about it. Perhaps, we can raise another task to support multi-platform for snapshotter

misc/snapshotter/Dockerfile
```diff
@@ -1,12 +1,21 @@
...
 
+ARG TARGETARCH
+
 RUN apk add --no-cache curl
 RUN apk add --no-cache --upgrade grep
-RUN export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")') && \
-    wget https://github.com/dragonflyoss/image-service/releases/download/$NYDUS_VERSION/nydus-static-$NYDUS_VERSION-linux-amd64.tgz && \
+RUN unset NYDUS_VERSION && \
+    while [ -z $NYDUS_VERSION ]; do \
+        export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")'); sleep 1; \
```
changweige, line 9: It's better to limit the total retry times.

misc/snapshotter/Dockerfile
```diff
@@ -1,12 +1,21 @@
...
+    done && \
+    echo NYDUS_VERSION: $NYDUS_VERSION && \
+    if [ -z $NYDUS_VERSION ]; then echo Invalid NYDUS_VERSION && exit -1; fi && \
```
changweige, line 12: This condition judgment conflicts with line 8, which means the loop should not end.

misc/snapshotter/Dockerfile
```diff
@@ -1,12 +1,21 @@
...
+    echo TARGETARCH: $TARGETARCH && \
+    if [ "$TARGETARCH" != "amd64" -a "$TARGETARCH" != "arm64" -a "$TARGETARCH" != "ppc64le" -a "$TARGETARCH" != "riscv64" ]; then echo Unsupported TARGETARCH for nydus && exit -1; fi && \
```
changweige, line 14: I still incline to simplify the test by assuming that only amd64 is the base nydus artifacts platform since the test target platform is decided and can't changed

misc/snapshotter/Dockerfile
sctb512: Done.

misc/snapshotter/Dockerfile
sctb512: Yes. Now, we limit the total retry times, so we still need this.

misc/snapshotter/Dockerfile
sctb512: It still failed when i retry 20 times. Maybe we could assume it will work eventually.

sctb512:
Dockerfile with `ARG TARGETARCH` is more adaptable in multiple architectures. However, this Dockerfile is only for the E2E test. It is better to keep things simple. 😊
Maybe we could open a new PR to solve this requirement?

misc/snapshotter/Dockerfile
sctb512: It works now. 10 minutes are enough.

changweige:
> Dockerfile with `ARG TARGETARCH` is more adaptable in multiple architectures. However, this Dockerfile is only for the E2E test. It is better to keep things simple. 😊 Maybe we could open a new PR to solve this requirement?

Agree.

changweige (approved):

changweige:
Thanks for investigating this 👍

PR #313, opened 2023-01-13
metrics: avoid collect FS metrics for fscache driver

sctb512:
The metrics for `/api/v1/metrics` are in the RAFS of nydus daemon. When the driver is fscache, there are no metrics in RAFS.


changweige (approved):
LGTM

PR #314, opened 2023-01-13
misc: use fixed nydus version for E2E test

sctb512:
The value of NYDUS_VERDION is null frequently if we request it from github API. We cannot solve this issue by sleeping 1 second, because the next available time is in the response headers and longer than 1 second. The NYDUS_VERSION is not changed often, so we can fix it.

More information: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limit-headers


misc/snapshotter/Dockerfile
```diff
@@ -2,10 +2,7 @@ FROM alpine:3.17.0 AS sourcer
 
 RUN apk add --no-cache curl
 RUN apk add --no-cache --upgrade grep
-RUN unset NYDUS_VERSION && RETRY_COUNT=0 && \
-    while [ -z $NYDUS_VERSION ] && [ $RETRY_COUNT != 600 ]; do \
-    export NYDUS_VERSION=$(curl --silent "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | grep -Po '"tag_name": "\K.*?(?=")'); RETRY_COUNT=$(($RETRY_COUNT+1)); sleep 1; done && \
-    echo nydus version: $NYDUS_VERSION after $RETRY_COUNT times && if [ -z $NYDUS_VERSION ]; then echo Invalid NYDUS_VERSION && exit -1; fi && \
+RUN export NYDUS_VERSION=v2.1.2 && \
```
changweige, line 5: Can we use docker file build argument with default value `v2.1.2` here

.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Build
         run: |
           # Download nydus components
-          NYDUS_VER=v$(curl -s "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | jq -r .tag_name | sed 's/^v//')
+          NYDUS_VER=v2.1.2
```
changweige, line 59: Strange, the e2e test never got empty `NYDUS_VER` before. Is it really a github rate limitation problem? cc @liubin

changweige (approved):
Thanks for investigating the unreliable smoke test.

PR #317, opened 2023-01-16
metrics: avoid to collect inflight metrics for fscache driver

sctb512:
Similar to FS metrics. The metrics for /api/v1/metrics/inflight are in the RAFS of nydus daemon.
When the driver is fscache, there are no metrics in RAFS.


changweige (approved):
LGTM

PR #320, opened 2023-01-16
Remove some cli parameters

changweige:
Nydus-snapshotter now has its own toml based configuration file whose entries are duplicated with CLI startup parameters.
In the first phase, we will remove those not frequently used cli parameters and suggest users to migrate the configuration to toml based configuration file.

### DIFF
--- a/cmd/containerd-nydus-grpc/pkg/command/flags.go
+++ b/cmd/containerd-nydus-grpc/pkg/command/flags.go
@@ -25,28 +25,17 @@ const (
 type Args struct {
 	Address                  string
 	LogLevel                 string
-	LogDir                   string
 	ConfigPath               string
 	SnapshotterConfigPath    string
 	RootDir                  string
-	CacheDir                 string
-	GCPeriod                 string
-	ValidateSignature        bool
-	PublicKeyFile            string
-	ConvertVpcRegistry       bool
 	NydusdPath               string
 	NydusImagePath           string
 	SharedDaemon             bool
 	DaemonMode               string
 	FsDriver                 string
-	SyncRemove               bool
 	MetricsAddress           string
-	EnableStargz             bool
-	DisableCacheManager      bool
 	LogToStdout              bool
 	EnableNydusOverlayFS     bool
-	NydusdThreadsNumber      int
-	CleanupOnClose           bool
 	KubeconfigPath           string
 	EnableKubeconfigKeychain bool
 	RecoverPolicy            string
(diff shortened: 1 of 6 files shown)


cmd/containerd-nydus-grpc/pkg/command/flags.go
```diff
@@ -127,31 +93,13 @@ func buildFlags(args *Args) []cli.Flag {
 			Usage:       "whether to enable nydus-overlayfs",
 			Destination: &args.EnableNydusOverlayFS,
 		},
-		&cli.BoolFlag{
-			Name:        "enable-stargz",
```
imeoer, line 131: Does this impact compatibility?

misc/snapshotter/entrypoint.sh
```diff
@@ -17,10 +17,7 @@ if [ "$#" -eq 0 ]; then
 		--root ${NYDUS_LIB} \
 		--address ${NYDUS_RUN}/containerd-nydus-grpc.sock \
 		--log-level ${LEVEL} \
-		--enable-metrics=${ENABLE_METRICS} \
-		--enable-nydus-overlayfs=${ENABLE_NYDUS_OVERLAY} \
 		--daemon-mode ${NYDUSD_DAEMON_MODE} \
```
imeoer, line 20: Can we also remove `--daemon-mode` ? It seems using `shared`.

cmd/containerd-nydus-grpc/pkg/command/flags.go
changweige: I suppose it doesn't. Stargz compatibility does not work for a long time.

misc/snapshotter/entrypoint.sh
changweige: `multiple` mode is the default one, so we should keep it?

cmd/containerd-nydus-grpc/pkg/command/flags.go
imeoer: Should other flags still be reserved to ensure that snapshotter can be started if users use old flags?

misc/snapshotter/entrypoint.sh
imeoer: Ok, just keep `--daemon-mode` flag and use `multiple` by default.

cmd/containerd-nydus-grpc/pkg/command/flags.go
changweige: I am inclined not to reserve other parameters. We have to migrate to the configuration file pattern in the end.
This PR only removes the parameters that are not likely to be used.

misc/snapshotter/entrypoint.sh
changweige: Let's change the shell variable default value to `multiple`

changweige: done

cmd/containerd-nydus-grpc/pkg/command/flags.go
imeoer: Ok, we need notice users that this is a breaking change in the release note.

imeoer (approved):

PR #321, opened 2023-01-16
fix k8s e2e test docker file

changweige:
Fix the nydus-snapshotter CLI parameters in k8s e2e test docker file where several CLI parameters are removed

### DIFF
--- a/misc/snapshotter/entrypoint.sh
+++ b/misc/snapshotter/entrypoint.sh
@@ -2,8 +2,6 @@
 
 NYDUS_LIB="${NYDUS_LIB:-/var/lib/containerd-nydus}"
 NYDUS_RUN="${NYDUS_RUN:-/run/containerd-nydus}"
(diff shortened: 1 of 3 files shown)


misc/snapshotter/entrypoint.sh
```diff
@@ -17,10 +17,7 @@ if [ "$#" -eq 0 ]; then
 		--root ${NYDUS_LIB} \
 		--address ${NYDUS_RUN}/containerd-nydus-grpc.sock \
 		--log-level ${LEVEL} \
-		--enable-metrics=${ENABLE_METRICS} \
```
sctb512, line 20: Shall we remove the related variables as well?
```shell
ENABLE_NYDUS_OVERLAY="${ENABLE_NYDUS_OVERLAY:-true}"
ENABLE_METRICS="${ENABLE_METRICS:-false}"
```

changweige: Good point, already removed

imeoer (approved):

PR #328, opened 2023-01-18
assign root dir in the first place when filling up defaults

changweige:
As other default value may depend on root dir.


sctb512 (approved):
LGTM

imeoer (approved):

PR #329, opened 2023-01-19
Back port some stable patches for v0.5.x

changweige:
Clearly pick some stable patches for v0.5 and prepare to release v0.5.1

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -518,7 +518,7 @@ func (m *Manager) DestroyDaemon(d *daemon.Daemon) error {
 		log.L.Warnf("Failed to wait for daemon, %v", err)
 	}
 
-	collector.NewDaemonEventCollector(string(types.DaemonStateDestroyed)).Collect()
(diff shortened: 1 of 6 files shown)


pkg/manager/manager.go
```diff
@@ -311,8 +311,9 @@ func (m *Manager) handleDaemonDeathEvent() {
 			log.L.Warnf("Daemon %s was not found", ev.daemonID)
 			return
 		}
-		collector.NewDaemonInfoCollector(&d.Version, -1).Collect()
+
 		d.Lock()
+		collector.NewDaemonInfoCollector(&d.Version, -1).Collect()
```
sctb512, line 316: Whether is a lock needed before another call to this method?

sctb512: In a few cases, I get the extra values that `d.version` is incorrect. However, the value of the expected version is eventually correct.

changweige: What do you mean by "incorrect"?  The version string is shorter than expected or polluted?

sctb512: The `version` string is null. It appeared by chance.

PR #336, opened 2023-01-30
Remove the other cli parameters

changweige:
In this stage, we're going the remove other not frequently used CLI parameters and let e2e test adapt it.

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -61,7 +61,7 @@ func (c *SnapshotterConfig) FillUpWithDefaults() error {
 	if daemonConfig.NydusdConfigPath == "" {
 		daemonConfig.NydusdConfigPath = defaultNydusDaemonConfigPath
 	}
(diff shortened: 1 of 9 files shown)


config/default.go
```diff
@@ -61,6 +61,7 @@ func (c *SnapshotterConfig) FillUpWithDefaults() error {
 	if daemonConfig.NydusdConfigPath == "" {
 		daemonConfig.NydusdConfigPath = defaultNydusDaemonConfigPath
 	}
+	daemonConfig.RecoverPolicy = "restart"
```
imeoer, line 64: Maybe make `restart`, `failover` as constant?

imeoer:
others LGTM.

config/default.go
changweige: We already have the ReoverPolicy custom type definition which is an alias of int. So let's use the parse String() method here rather literal string

PR #343, opened 2023-02-03
enable more golang linters

changweige:
In addition, fix linter warnings reported by new linter.

### DIFF
--- a/pkg/auth/image_proxy_test.go
+++ b/pkg/auth/image_proxy_test.go
@@ -21,12 +21,20 @@ import (
 	runtime "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
-type MockImageService struct {
+type MockAlphaImageService struct {
 	runtime_alpha.UnimplementedImageServiceServer
 }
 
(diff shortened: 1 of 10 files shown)


pkg/auth/image_proxy_test.go
```diff
@@ -55,14 +63,26 @@ func TestFromImagePull(t *testing.T) {
...
+		assert.NoError(err)
+	}()
+	defer func() {
+		mockRPC.Stop()
+	}()
 
 	AddImageProxy(ctx, proxyRPC, mockSocket)
-	go proxyRPC.Serve(lp)
-	defer lp.Close()
+	go func() {
```
imeoer, line 79: Maybe `go assert.NoError(proxyRPC.Serve(lp))`, `defer proxyRPC.Stop()` ?

changweige: I tried, but it seems that `proxyRPC.Serve(lp)` be first evaluated before the new goroutine starts. So the test blocks on `proxyRPC.Serve(lp)`

PR #346, opened 2023-02-06
mark stargz configuration as experimental

changweige:
As the stargz compatiblity feature is not production ready
 and reliable enought

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -93,6 +93,10 @@ const (
 	FsDriverFscache string = "fscache"
 )
 
+type Experimental struct {
+	EnableStargz bool `toml:"enable_stargz"`
+}
+
 // Configure how to start and recover nydusd daemons
 type DaemonConfig struct {
 	NydusdPath       string `toml:"nydusd_path"`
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -175,6 +178,7 @@ type SnapshotterConfig struct {
 	ImageConfig        ImageConfig        `toml:"image"`
 	CacheManagerConfig CacheManagerConfig `toml:"cache_manager"`
 	LoggingConfig      LoggingConfig      `toml:"log"`
+	Experimental       Experimental       `toml:"experimental"`
```
imeoer, line 181: Will it move out once the feature becomes ready?

changweige: Yes. We will move it out when it's ready. In fact, the stargz compibility feature has malfunctioned for quite a time.

imeoer: How do we ensure the CLI flags compatibility when moving it out?

changweige: CLI does not has --enable-stargz flag now. and stargz feature has malfunctioned for a long time with stable nydusd release. So I suppose we don't have to worry about compatiblity.

config/config.go
```diff
@@ -93,6 +93,10 @@ const (
 	FsDriverFscache string = "fscache"
 )
 
+type Experimental struct {
```
sctb512, line 96: Could we add an example in `toml` and related document?

changweige: Good point, already changed the example toml configuraion file

sctb512 (approved):
LGTM

imeoer:
Need to be rebased.

changweige:
> Need to be rebased.

Rebased. Thanks.

PR #349, opened 2023-02-08
generate accessed files list for container with NRI plugin

sctb512:
Generate an in-order accessed files list for container workload. The accessed files list can be set as `prefetch-patterns` when converting OCI images to nydus format images.

Related to https://github.com/dragonflyoss/image-service/issues/807

# Requirements

- NRI 2.0: Which has been integrated into containerd since [v1.7.0-beta.1](https://github.com/containerd/containerd/tree/v1.7.0-beta.1).

# Workflow
optmizer_workflow

1. Run the optimizer as an NRI plugin to run optimizer server when the StartContainer event occurs.
2. Optimizer server joins container mount namespace and starts fanotify server.
3. Optimizer server detects fanotify event and sends the accessed file descriptor to client.
4. Optimizer client converts file descriptor to path.
5. Generate list of accessed files on local disk.
6. Convert OCI images with accessed files list to nydus image.

# Generate an accessed files list

To install the optimizer and optimizer-server, all you need is the command:

```
make optimizer && make install-optimizer

```

Install `containerd` that supports NRI 2.0：
```
VERSION=1.7.0-beta.3

CONTAINERD_DIR=$(cat /lib/systemd/system/containerd.service | grep "ExecStart=" | awk -F= '{gsub("/containerd","",$2); print $2}')

wget https://github.com/containerd/containerd/releases/download/$VERSION/containerd-$VERSION-linux-amd64.tar.gz
tar -zxvf containerd-$VERSION-linux-amd64.tar.gz

sudo install -D -m 755 bin/* $CONTAINERD_DIR
rm -rf bin
$CONTAINERD_DIR/containerd --version
```

Modify containerd's toml configuration file to enable NRI.

```
sudo tee -a /etc/containerd/config.toml <<- EOF
[plugins."io.containerd.nri.v1.nri"]
  config_file = "/etc/nri/nri.conf"
  disable = false
  plugin_path = "/opt/nri/plugins"
  socket_path = "/var/run/nri.sock"
EOF

sudo tee /etc/nri/nri.conf <<- EOF
disableConnections: false
EOF

sudo systemctl restart containerd

```

Now, just start a container workload and you will get the list of accessed files in `persistDir`.

```
sudo tee nginx.yaml <<- EOF
metadata:
  name: nginx
image:
  image: nginx:latest
log_path: nginx.0.log
linux: {}
EOF

sudo tee pod.yaml <<- EOF
metadata:
  name: nginx-sandbox
  namespace: default
  attempt: 1
  uid: hdishd83djaidwnduwk28bcsb
log_directory: /tmp
linux: {}
EOF

crictl run nginx.yaml pod.yaml

```

The result file for the nginx image is `/opt/nri/optimiser/results/nginx:latest`.

# Optimize a nydus image

Nydus provides a nydusify CLI tool to convert OCI images from the source registry or local file system to nydus format and push them to the target registry.

We can install the nydusify cli tool from the nydus package.

```
VERSION=v2.1.3

wget https://github.com/dragonflyoss/image-service/releases/download/$VERSION/nydus-static-$VERSION-linux-amd64.tgz
tar -zxvf nydus-static-$VERSION-linux-amd64.tgz
sudo install -D -m 755 nydus-static/nydusify /usr/local/bin/nydusify
sudo install -D -m 755 nydus-static/nydus-image /usr/local/bin/nydus-image

```

With the `--prefetch-patterns` argument, we can specify the list of files to be written in the front of the nydus image and be prefetched in order when starting a container.

```
sudo nydusify convert --nydus-image $(which nydus-image) --source nginx --target sctb512/nginx:optimized-nydusv6 --fs-version 6 --prefetch-patterns < /opt/nri/optimizer/results/nginx:latest

```

Make sure the nydusd and nydus-snapshotter environments are installed correctly. Then, start a container with an optimized nydus image.

```
sudo nerdctl --snapshotter nydus run --rm --net host -it sctb512/nginx:optimized-nydusv6

```


imeoer:
Hi @sctb512 , any doc for this? Does it work on nydus image?

sctb512:
> Hi @sctb512 , any doc for this? Does it work on nydus image?

Work in progress. This NRI plugin is image format independent. And it works on nydus image.

Makefile
```diff
@@ -36,9 +37,16 @@ endif
...
 
+OPTIMIZER_SERVER = tools/optimizer-server
+OPTIMIZER_SERVER_TOML = ${OPTIMIZER_SERVER}/Cargo.toml
+OPTIMIZER_SERVER_BIN = ${OPTIMIZER_SERVER}/target/release/optimizer-server
+
 .PHONY: build
 build:
 	GOOS=${GOOS} GOARCH=${GOARCH} ${PROXY} go build -ldflags "$(LDFLAGS)" -v -o bin/containerd-nydus-grpc ./cmd/containerd-nydus-grpc
+	${CARGO} fmt --manifest-path ${OPTIMIZER_SERVER_TOML} -- --check
+	${CARGO} build --release --manifest-path ${OPTIMIZER_SERVER_TOML} && cp ${OPTIMIZER_SERVER_BIN} ./bin
```
changweige, line 48: Could we have a new Makefile target for building the `optimizer` ? It will even better if we have its own Makefile

changweige: By default, we should not build the optimizer as developers may not have rust environment

cmd/optimizer/main.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
+
+	"github.com/containerd/nri/pkg/api"
+	"github.com/containerd/nri/pkg/stub"
+	"github.com/containerd/nydus-snapshotter/pkg/fanotify"
+)
+
+type config struct {
+	LogFile string   `yaml:"logFile"`
```
changweige, line 28: Could we use the Linux naming style like `log_file`

cmd/optimizer/main.go
```diff
@@ -0,0 +1,215 @@
...
+	Events  []string `yaml:"events"`
+
+	ServerPath string `yaml:"serverPath"`
+	PersistDir string `yaml:"persistDir"`
```
changweige, line 32: ditto

cmd/optimizer/main.go
```diff
@@ -0,0 +1,215 @@
...
+	Timeout    int    `yaml:"timeout"`
+	Overwrite  bool   `yaml:"overwrite"`
```
changweige, line 34: Moreover, toml is preferred

cmd/optimizer/main.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse provided configuration: %w", err)
+	}
+
+	p.mask, err = api.ParseEventMask(cfg.Events...)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse events in configuration: %w", err)
+	}
+
+	if cfg.LogFile != oldCfg.LogFile {
```
changweige, line 70: When to close the file `f`

cmd/optimizer/main.go
sctb512: I followed this style in NRI's example code: https://github.com/containerd/nri/blob/main/plugins/logger/nri-logger.go#L33. Maybe we could keep the same style with them?

Makefile
sctb512: Yes, this is reasonable. Fixed it by adding `build-optimizer`. Is it acceptable?

cmd/optimizer/main.go
sctb512: Now the optimizer server will close the file descriptor after sending the `path` to the client.

cmd/optimizer/main.go
sctb512: Anyway. It makes sense to use the same parameter style in this project. Already fixed.

cmd/optimizer/main.go
sctb512: Fixed.

cmd/optimizer/main.go
sctb512: Emm... It is not the same issue. The file descriptor mentioned in last comment is created by optimizer server. So we close it after getting the path.

The file `f` is closed before exiting this NRI plugin, i.e. in `onClose` event.

tools/optimizer-server/src/main.rs
```diff
@@ -0,0 +1,128 @@
+use fanotify::low_level::*;
```
changweige, line 1: Please add license claim here

tools/optimizer-server/src/main.rs
```diff
@@ -0,0 +1,128 @@
...
+use nix::poll::{poll, PollFd, PollFlags};
+use nix::sched::{setns, CloneFlags};
+use nix::unistd::fork;
+use nix::unistd::ForkResult::{Child, Parent};
+use std::env;
+use std::fs;
+use std::io;
+use std::io::Write;
+use std::os::unix::io::AsRawFd;
+use std::path::Path;
```
changweige, line 11: We usually put `std` dependencies to the top position and then the third-party dependencies and then then project internal dependenecies

sctb512: Fixed it.

tools/optimizer-server/src/main.rs
sctb512: Done.

Makefile
```diff
@@ -36,10 +37,21 @@ endif
...
+OPTIMIZER_SERVER_BIN = ${OPTIMIZER_SERVER}/target/release/optimizer-server
+
 .PHONY: build
 build:
 	GOOS=${GOOS} GOARCH=${GOARCH} ${PROXY} go build -ldflags "$(LDFLAGS)" -v -o bin/containerd-nydus-grpc ./cmd/containerd-nydus-grpc
 
+.PHONY: build-optimizer
+build-optimizer:
+	${CARGO} fmt --manifest-path ${OPTIMIZER_SERVER_TOML} -- --check
+	${CARGO} build --release --manifest-path ${OPTIMIZER_SERVER_TOML} && cp ${OPTIMIZER_SERVER_BIN} ./bin
```
changweige, line 52: Could you please also add `cargo clippy` to lint the tool, which will give some wise suggestions.

cmd/optimizer-nri-plugin/main.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/containerd/nri/pkg/api"
+	"github.com/containerd/nri/pkg/stub"
+	"github.com/containerd/nydus-snapshotter/pkg/errdefs"
+	"github.com/containerd/nydus-snapshotter/pkg/fanotify"
+	"github.com/containerd/nydus-snapshotter/version"
+	"github.com/pelletier/go-toml"
+)
+
+const (
+	defaultLogFile    = "/opt/nri/optimizer/optimizer.log"
```
changweige, line 31: Logs should be stored in `/var/log` or just stream it to syslog service rather than set a file as default on at `/opt`

cmd/optimizer-nri-plugin/main.go
```diff
@@ -0,0 +1,327 @@
...
+func (p *plugin) Configure(config, runtime, version string) (stub.EventMask, error) {
+	log.Infof("got configuration data: %q from runtime %s %s", config, runtime, version)
+	if config == "" {
+		return p.mask, nil
+	}
+
+	oldCfg := cfg
+	tree, err := toml.Load(config)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse TOML: %w", err)
```
changweige, line 142: Can you use errors.Errorf  and errors.Wrap ? `%w` is deprecated now in Golang

cmd/optimizer-nri-plugin/main.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	log.Infof("configuration: %#v", cfg)
+
+	return p.mask, nil
+}
+
+func (p *plugin) StartContainer(pod *api.PodSandbox, container *api.Container) error {
+	dump("StartContainer", "pod", pod, "container", container)
+
+	imageName := GetImageName(container.Annotations)
```
changweige, line 174: We should validate the image. Containerd already provides a package to validate image name.

cmd/optimizer-nri-plugin/main.go
```diff
@@ -0,0 +1,327 @@
...
+	if fanotifyServer, ok := globalFanotifyServer[imageName]; ok {
+		fanotifyServer.StopServer()
+	} else {
+		return nil, errors.New("can not find fanotify server for container image " + imageName)
+	}
+
+	return update, nil
+}
+
+func GetImageName(annotations map[string]string) string {
```
changweige, line 203: There exists a standard method comes from contanerd package to extract the tagged image name. Can you try to use it.

tools/optimizer-server/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+    writer.write_all(path.as_bytes())?;
+    writer.flush()
+}
+
+fn handle_fanotify_event(fd: i32) {
+    let mut writer = io::stdout();
+
+    let mut fds = [PollFd::new(fd.as_raw_fd(), PollFlags::POLLIN)];
+    loop {
+        let poll_num = poll(&mut fds, -1).unwrap();
```
changweige, line 88: Try to avoid directly `unwrap` unless we be sure that it's safe

pkg/fanotify/fanotify.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	fserver.Client = &conn.Client{
+		Scanner: bufio.NewScanner(notifyR),
+	}
+
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	f, err := os.OpenFile(fserver.PersistFile, os.O_CREATE|os.O_WRONLY, 0644)
```
changweige, line 91: Should close the file `f` ?

Makefile
sctb512: Good suggestion. I have added this.

cmd/optimizer-nri-plugin/main.go
sctb512: Now, we redirect it to the syslog service.

cmd/optimizer-nri-plugin/main.go
sctb512: Fixed.

cmd/optimizer-nri-plugin/main.go
sctb512: Done.

tools/optimizer-server/src/main.rs
sctb512: Fixed it with `match`.

pkg/fanotify/fanotify.go
sctb512: Yes, now we close it before exiting the server.

tools/optimizer-server/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+    )
+}
+
+fn mark_fanotify(fd: i32, path: &str) -> Result<(), io::Error> {
+    fanotify_mark(
+        fd,
+        FAN_MARK_ADD | FAN_MARK_MOUNT,
+        FAN_OPEN | FAN_CLOSE_WRITE,
+        AT_FDCWD,
+        path,
```
changweige, line 71: FAN_OPEN_EXEC
No need to monitor event FAN_CLOSE_WRITE

sctb512: Fixed it.

changweige (approved):
Thanks. This is indeed a useful tool to help us build more efficient nydus images

PR #351, opened 2023-02-12
export label and handler defined in containerd

imeoer:
See https://github.com/containerd/containerd/pull/8054


pkg/label/label.go
```diff
@@ -6,12 +6,22 @@
...
 package label
 
+import (
+	snpkg "github.com/containerd/containerd/pkg/snapshotters"
+)
+
+// For package compatibility, we still keep the old exported name here.
+var AppendLabelsHandlerWrapper = snpkg.AppendInfoHandlerWrapper
+
+// For package compatibility, we still keep the old exported name here.
```
changweige, line 16: This compatibility trick looks good to me.  Hopefully, we can update this package's consumer to let them directly import `containerd/snapshotters` packge.

changweige (approved):

PR #355, opened 2023-02-16
don't overwrite existed user's configuration file

changweige:
We should not overwrite the user's  existed configuration file


Makefile
```diff
@@ -54,15 +58,14 @@ clear:
 
 .PHONY: install
 install:
-	sudo install -D -m 755 bin/containerd-nydus-grpc /usr/local/bin/containerd-nydus-grpc
-	sudo install -D -m 755 misc/snapshotter/nydusd-config.fusedev.json /etc/nydus/nydusd-config.fusedev.json
-	sudo install -D -m 755 misc/snapshotter/nydusd-config.fscache.json /etc/nydus/nydusd-config.fscache.json
-	sudo install -D -m 755 misc/snapshotter/config.toml /etc/nydus/config.toml
-	sudo ln -f -s /etc/nydus/nydusd-config.${FS_DRIVER}.json /etc/nydus/nydusd-config.json
-	sudo install -D -m 644 misc/snapshotter/nydus-snapshotter.${FS_DRIVER}.service /etc/systemd/system/nydus-snapshotter.service
+	@sudo install -D -m 755 bin/containerd-nydus-grpc /usr/local/bin/containerd-nydus-grpc
```
sctb512, line 61: Should we keep the output of the installed binary path for users?

imeoer (approved):

Makefile
changweige: Good point. We should remind users that artifacts are being installed.

```
+ install bin/containerd-nydus-grpc
+ install misc/snapshotter/nydus-snapshotter.fusedev.service
```

PR #363, opened 2023-02-17
fail the snapshotter immediately if metrics server can't be bound

changweige:
We should fail snapshotter's startup if the metrics address weren't bound.

### DIFF
--- a/pkg/metrics/listener.go
+++ b/pkg/metrics/listener.go
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021. Ant Group. All rights reserved.
+ * Copyright (c) 2023. Nydus Developers. All rights reserved.
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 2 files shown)


pkg/metrics/listener.go
```diff
@@ -30,9 +30,11 @@ func NewHTTPListener(addr string) error {
 
 	log.L.Infof("Start metrics HTTP server on %s", addr)
 
-	if err := http.ListenAndServe(addr, nil); err != nil {
-		return fmt.Errorf("error serve on %s: %v", addr, err)
-	}
+	go func() {
+		if err := http.ListenAndServe(addr, nil); err != nil {
+			log.L.Errorf("error serve on %s: %v", addr, err)
```
imeoer, line 35: Can we make sure to exit snapshotter once `http.ListenAndServe` fails?

changweige: Better not to, we should try to maintain the snapshots service , I suppose.

sctb512, line 35: Could we output information such as "Failed to serve metrics HTTP endpoint on %s: %v" as this error is in the go routine.

changweige: Good point. We'd better give more information about error context.

imeoer: So the user can only determine if the metrics server is working by checking the error log? I think it's a not good design.

changweige: Um. If we indeed need to ensure that the metrics server is successfully started, perhaps we can try to connect to the listener in the main goroutine

changweige: I think we just need to move the connection `listen` out from `ListenAndServe`. `listen` is likely to fail when the port is  already bound when serve can only fail when the connection is closed by accident

sctb512: Agree. It is sufficient to quit `snapshotter` if the port is already bound.

sctb512 (approved):
LGTM

pkg/metrics/listener.go
```diff
@@ -28,11 +38,17 @@ func NewHTTPListener(addr string) error {
...
-	if err := http.ListenAndServe(addr, nil); err != nil {
-		return fmt.Errorf("error serve on %s: %v", addr, err)
+	l, err := net.Listen("tcp", addr)
+	if err != nil {
+		return errors.Wrapf(err, "metrics server listener, addr=%s", addr)
 	}
 
+	go func() {
+		log.L.Infof("Start metrics HTTP server on %s", addr)
+		if err := http.Serve(l, nil); trapClosedConnErr(err) != nil {
```
sctb512, line 48: Perhaps `err != nil && !errors.Is(err, net.ErrClosed)` will also work?

imeoer (approved):

PR #364, opened 2023-02-17
Fscache e2e

changweige:
1. Fix the test status check that cache dir must be empty
2. Add failover test case for fscache driver

### DIFF
--- a/integration/entrypoint.sh
+++ b/integration/entrypoint.sh
@@ -192,7 +192,9 @@ function umount_global_shared_mnt {
 }
 
 function is_cache_cleared {
-    if [[ $(ls -A "${SNAPSHOTTER_CACHE_DIR}") == "" ]]; then
+    # With fscache driver, 2.1 nydusd don't have API to release the cache files.
+    # Thy locate at directory ${SNAPSHOTTER_CACHE_DIR}/cache
+    if [[ $(ls -A -p "${SNAPSHOTTER_CACHE_DIR}" | grep -v /) == "" ]]; then
         true
     else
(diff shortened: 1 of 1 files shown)


integration/entrypoint.sh
```diff
@@ -571,17 +596,19 @@ only_restart_snapshotter multiple
...
-# kill_multiple_nydusd_recover_failover shared
 
 ctr_snapshot_usage multiple
 ctr_snapshot_usage shared
 
 blob_manager_pull_preheat multiple fusedev
 
 if [[ $(can_erofs_ondemand_read) == 0 ]]; then
+    kill_multiple_nydusd_recover_failover multiple
+    kill_multiple_nydusd_recover_failover shared
```
sctb512, line 607: It seems that `can_erofs_ondemand_read` is used to judge if erofs is available. Why do we have `kill_multiple_nydusd_recover_failover` and  `kill_multiple_nydusd_recover_failover` which use fuse driver in this case?

changweige: Because github action's host's kernel can't support EROFS on-demand read. So we have an internal kernel supporting both nydusd failover and erofs.

sctb512: Got it.

sctb512: But it seems that `kill_multiple_nydusd_recover_failover` can also run without EROFS?
```shell
function kill_multiple_nydusd_recover_failover {
    ...
    reboot_containerd "${daemon_mode}" fusedev failover
    ...
}
```

sctb512 (approved):
LGTM

PR #366, opened 2023-02-22
print instance image id from system diagnose API

changweige:
In order to print more info about image instance to help analyze


sctb512 (approved):
LGTM

imeoer (approved):

PR #367, opened 2023-02-22
fix system controller default socket address

changweige:
The default address should locate at
/var/run/containerd-nydus/system.sock


sctb512 (commented):
LGTM

imeoer (approved):

PR #370, opened 2023-02-22
upgrade all the dependencies

changweige:
It's time to upgrade all the dependent packages now


snapshot/utils.go
```diff
@@ -19,3 +19,7 @@ func isNydusMetaLayer(labels map[string]string) bool {
 	_, ok := labels[label.NydusMetaLayer]
 	return ok
 }
+
+func isNydusLayer(labels map[string]string) bool {
```
imeoer, line 23: Should be removed?

changweige: Let's close this PR. #387  is upgrading the deps in a elegant way

PR #371, opened 2023-02-22
upgrade project's golang version to 1.19.6

changweige:
Thus to catch up with containerd

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: "1.19.1"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: "1.19.1"
+          go-version: "1.20.1"
```
imeoer, line 18: Is this an action bug?

changweige: No. For `golangci-lint`, we'd like to use a higher version of golang thus to allow it to report more warnings

PR #377, opened 2023-02-24
only count nydusd local cache if it a nydus layer

changweige:
Sandbox container image may not be downloaded through CRI flow. So  cri related annotations won't be appended.


imeoer (approved):
LGTM

PR #378, opened 2023-02-24
config: validate mirrors directory when starting snapshotter

sctb512:


config/tools/tools.go
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright (c) 2023. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package tools
+
+import "os"
+
+func IsDirExisted(path string) (bool, error) {
```
changweige, line 11: Can we put such helper functions into files named as utils.go ?

changweige (commented):
Otherwise, looks good

config/tools/tools.go
sctb512: Done.

Issue #379, opened 2023-02-24
toml configuration file does not work


sctb512:
It seems that the `toml` configuration file does not work. Because it is overwritten even the flags.Args is null in the `ParseParameters` function.
Here is a test:

```shell
diff --git a/config/config_test.go b/config/config_test.go
index ad47909..d123f45 100644
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -10,6 +10,7 @@ import (
        "testing"
        "time"
 
+       "github.com/containerd/nydus-snapshotter/cmd/containerd-nydus-grpc/pkg/command"
        "github.com/stretchr/testify/assert"
 )
 
@@ -81,6 +82,13 @@ func TestLoadSnapshotterTOMLConfig(t *testing.T) {
 
        A.EqualValues(cfg, &exampleConfig)
 
+       var args = command.Args{}
+       args.RootDir = "/var/lib/containerd/nydus"
+       exampleConfig.Root = "/var/lib/containerd/nydus"
+       err = ParseParameters(&args, cfg)
+       A.NoError(err)
+       A.EqualValues(cfg, &exampleConfig)
+
        err = ProcessConfigurations(cfg)
        A.NoError(err)
```

```
--- FAIL: TestLoadSnapshotterTOMLConfig (0.00s)
    /data00/home/tangbin.bin/test_code/nydus-snapshotter/config/config_test.go:90: 
        	Error Trace:	/data00/home/tangbin.bin/test_code/nydus-snapshotter/config/config_test.go:90
        	Error:      	Not equal: 
[...]

PR #380, opened 2023-02-24
Fix the problem of snapshotter's toml configuration file not working

sctb512:
Remove default value of cmd arguments to avoid overwriting configuration loaded from `toml` file.
This PR is addressing #379


changweige (approved):
Overall looks good to me, please try to squash the patches with the same title

PR #385, opened 2023-02-27
Fix some small problems in the live upgrade procedure

mofishzz:
This is to prepare for the upcoming nydusd live upgrade function.

### DIFF
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -347,9 +347,11 @@ func (sc *Controller) upgradeNydusDaemon(d *daemon.Daemon, c upgradeRequest) err
 
 	log.L.Infof("Started service of upgraded daemon on socket %s", new.GetAPISock())
 
-	if err := manager.UpdateDaemon(&new); err != nil {
+	if err := manager.UpdateDaemonNoLock(&new); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/system/system.go
```diff
@@ -316,6 +318,11 @@ func (sc *Controller) upgradeNydusDaemon(d *daemon.Daemon, c upgradeRequest) err
 		return err
 	}
 
+	su := manager.SupervisorSet.GetSupervisor(d.ID())
+	if err := su.SendStatesTimeout(time.Second * 10); err != nil {
```
changweige, line 322: Why do we have to receive the old nydusd's states before a takeover request to new nydusd ?

changweige (commented):
Thanks for the patches. Only a few questions are left. Overall looks good and tidy.

pkg/system/system.go
mofishzz: It's a similar procedure to doDaemonFailover function. The snapshotter needs to send the old nydusd's state to new one when takeover. 

Here send refers to from snapshotter to nydus, as opposed to sendfd process, which may be a little confusing.

changweige (approved):

PR #390, opened 2023-03-02
action: add trivy ci for image vulnerability scan

imeoer:


.github/workflows/trivy.yml
```diff
@@ -0,0 +1,20 @@
+name: CI
```
changweige, line 1: Should rename the action name. something like scan vulnerability ?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Trivy Scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Trivy
+        run: |
+          wget https://github.com/aquasecurity/trivy/releases/download/v0.38.0/trivy_0.38.0_Linux-64bit.deb
+          sudo dpkg -i trivy_0.38.0_Linux-64bit.deb
+      - name: Scan Image
+        run: |
+          trivy image --timeout 60m ghcr.io/containerd/nydus-snapshotter:latest
```
changweige, line 20: How is the image `hcr.io/containerd/nydus-snapshotter:latest` built and uploaded?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,20 @@
+name: CI
+
+on:
+  push:
+    branches: ["*"]
+  pull_request:
+    branches: [main]
```
changweige, line 7: I suppose we don't have to check it on events push and pull_request, a corn job should be ok

PR #391, opened 2023-03-02
Fix override for log-to-stdout cmdline option

PKizzle:
Updates github.com/urfave/cli/v2 to v.2.24.4 which allows the usage of the `Count` property for `BoolFlag`. This enables the detection of the cmdline flag being set or not and allows for correctly overriding the option from the toml config.
The corresponding tests have been added as well.

Signed-off-by: Philipp Kolberg <philipp.kolberg@t-online.de>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -304,7 +304,9 @@ func ParseParameters(args *command.Args, cfg *SnapshotterConfig) error {
 	if args.LogLevel != "" {
 		logConfig.LogLevel = args.LogLevel
 	}
-	logConfig.LogToStdout = args.LogToStdout
+	if args.LogToStdoutCount > 0 {
+		logConfig.LogToStdout = args.LogToStdout
+	}
 
 	// --- remote storage configuration
 	// empty
--- a/config/config.go
+++ b/config/config.go
@@ -109,7 +109,7 @@ type DaemonConfig struct {
 }
 
 type LoggingConfig struct {
-	LogToStdout         bool
+	LogToStdout         bool   `toml:"log_to_stdout"`
 	LogLevel            string `toml:"level"`
 	LogDir              string `toml:"dir"`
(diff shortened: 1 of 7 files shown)


PKizzle:
I re-added the log_to_stdout option where it was removed by the mentioned commits

sctb512 (approved):
LGTM

sctb512:
There is a legacy issue. We should also restore the configuration item `log_to_stdout` in the toml file. https://github.com/containerd/nydus-snapshotter/pull/380/commits/c07eaa48131870282c4408a0b0ccf8dc1f0adae1#diff-f4126e327c22a9a152759a185ad32dcb8ae69c1b785429a354815b0618754788L39

changweige (approved):
Many thanks for bringing this trick into nydus-snapshotter !

changweige:
> There is a legacy issue. We should also restore the configuration item `log_to_stdout` in the toml file. [c07eaa4#diff-f4126e327c22a9a152759a185ad32dcb8ae69c1b785429a354815b0618754788L39](https://github.com/containerd/nydus-snapshotter/commit/c07eaa48131870282c4408a0b0ccf8dc1f0adae1#diff-f4126e327c22a9a152759a185ad32dcb8ae69c1b785429a354815b0618754788L39)

It is already restored in this PR?

config/config.go
```diff
@@ -109,7 +109,7 @@ type DaemonConfig struct {
 }
 
 type LoggingConfig struct {
-	LogToStdout         bool
+	LogToStdout         bool   `toml:"log_to_stdout"`
```
changweige, line 112: @sctb512 The `log_to_stdout` config is already restored

sctb512: Yes, it does. But the `log_to_stdout` entry in `misc/snapshotter/config.toml` is lost.

changweige: Yes. another PR should be raised to fix the example config toml

PR #393, opened 2023-03-03
config: avoid to use default dir if root is valid

sctb512:
In the default configuration for snapshotter we fill the cacheDir and LogDir:
```go
    logConfig := &c.LoggingConfig
    if len(logConfig.LogDir) == 0 {
		logConfig.LogDir = filepath.Join(c.Root, logging.DefaultLogDirName)
	}
    cacheConfig := &c.CacheManagerConfig
	if len(cacheConfig.CacheDir) == 0 {
		cacheConfig.CacheDir = filepath.Join(c.Root, "cache")
	}
```
But we do not fill these elements after loading the configuration from the `toml' file and parsing the arguments. This results in cacheDir and LogDir not being in snapshotter's RootDir.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -245,10 +245,6 @@ func ValidateConfig(c *SnapshotterConfig) error {
 			"\"enable_cri_keychain\" and \"enable_kubeconfig_keychain\" can't be set at the same time")
 	}
 
-	if !c.CacheManagerConfig.Disable && c.CacheManagerConfig.CacheDir == "" {
-		return errors.Wrapf(errdefs.ErrInvalidArgument, "cache directory is empty")
-	}
-
 	if c.RemoteConfig.MirrorsConfig.Dir != "" {
 		dirExisted, err := file.IsDirExisted(c.RemoteConfig.MirrorsConfig.Dir)
 		if err != nil {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -7,6 +7,7 @@
 package config
 
 import (
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


cmd/containerd-nydus-grpc/main.go
```diff
@@ -65,6 +66,17 @@ func main() {
 				}
 			}
 
+			if snapshotterConfig.Root != "" {
+				logConfig := &snapshotterConfig.LoggingConfig
```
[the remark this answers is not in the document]
sctb512: I think `FillUpWithDefaults` might not work. But putting the implementation in `config.MergeConfig` makes more sense.

changweige (commented):
In fact, cache dir and log dir do not have to locate under snapshotter's root. It can be anywhere as per the user's configurations.

sctb512:
> In fact, cache dir and log dir do not have to locate under snapshotter's root. It can be anywhere as per the user's configurations.

Got it. Should we keep the cache dir and log dir under the root directory if the user does not specify these values.
The problems I encountered was that the cache dir and log dir are always in `/var/lib/containerd-nydus` even I specify a root directory.

changweige:
> > In fact, cache dir and log dir do not have to locate under snapshotter's root. It can be anywhere as per the user's configurations.
> 
> Got it. Should we keep the cache dir and log dir under the root directory if the user does not specify these values. 

Yes. We should always put the log dir and cache dir under custom root dir

config/config.go
```diff
@@ -220,6 +222,19 @@ func MergeConfig(to, from *SnapshotterConfig) error {
...
+		logConfig := &to.LoggingConfig
+		if logConfig.LogDir == "" {
+			logConfig.LogDir = filepath.Join(to.Root, logging.DefaultLogDirName)
+		}
+		cacheConfig := &to.CacheManagerConfig
+		if cacheConfig.CacheDir == "" {
+			cacheConfig.CacheDir = filepath.Join(to.Root, "cache")
+		}
+	}
+
```
changweige, line 237: Could we move the above code into `ProcessConfigurations`? I think `MergeConfig` is only a thin layer to merge default configurations and custom configurations. But we can tweak the merged configurations in `ProcessConfigurations`

changweige:
Need rebase

config/config.go
sctb512: Done. PTAL

changweige (approved):
LGMT, thanks

Issue #392, opened 2023-03-03
the nydusd count metric is incorrect


sctb512:
The number of running nydus daemons  is incorrect. Here is an example:

```shell
# curl -s http://127.0.0.1:9110/v1/metrics | grep nydusd_count

# HELP nydusd_counts The counts of nydus daemon.
# TYPE nydusd_counts gauge
nydusd_counts{version="v2.1.0"} 60

# ps -ef | grep "bin/nydusd" | wc -l
74
```

PR #394, opened 2023-03-03
daemon: increase retry times to avoid premature returns

sctb512:
Currently, snapshotter retries for 2s while waiting for the nydus daemon's unix domain socket file to exist. This probably causes the nydusd_count metric to be incorrect because it returns prematurely.

This PR is addressing #392

### DIFF
--- a/pkg/daemon/client.go
+++ b/pkg/daemon/client.go
@@ -173,7 +173,7 @@ func WaitUntilSocketExisted(sock string) error {
 
 		return nil
 	},
-		retry.Attempts(20), // totally wait for 2 seconds, should be enough
+		retry.Attempts(100), // totally wait for 10 seconds, should be enough
 		retry.LastErrorOnly(true),
 		retry.Delay(100*time.Millisecond))
 }


pkg/daemon/client.go
```diff
@@ -173,7 +173,7 @@ func WaitUntilSocketExisted(sock string) error {
 
 		return nil
 	},
-		retry.Attempts(20), // totally wait for 2 seconds, should be enough
+		retry.Attempts(200), // totally wait for 20 seconds, should be enough
```
imeoer, line 176: Do we have to wait so long?

sctb512: How about 5s or 10s? In any case, 2s is too short and may not get the `nydusd' startup status right.

changweige: What worries me is why nydusd does not start its HTTP server after 2 seconds. It should be long enough. Extend the wait time is not a solution to the inconsistency problem. We should allow nydusd's startup failure and handle the inconsistency of metrics counting

changweige: But it's fine to expand the wait time for the existence of the UNIX socket. But we still have to handle the inconsistency condition. Let's merge this PR in the first place and raise another PR to handle the inconsistency once and for all.
Let's keep the issue https://github.com/containerd/nydus-snapshotter/issues/392 open to trace it. 10 seconds sounds good to me  🤗

sctb512: Well, I updated it.

changweige (approved):
LGTM

PR #399, opened 2023-03-06
allow nydusd and nydus-image absence in PATH

changweige:
Nydusd and nydus-image's path be customized.
It should throw error if they can't be found.


sctb512 (approved):
LGTM

Issue #398, opened 2023-03-06
Nydus image optimizer can be configured to print event timestamp


changweige:
Currently, only a sorted file list is generated by the image optimizer. With extra information like event access timestamp, it helps to analyze what files are accessed at the very beginning while others do not

PR #400, opened 2023-03-07
add size and timestamp for monitored files and dump to json file

sctb512:
This PR tries to add more information to the optimizer. The file size and event timestamp are useful to understand the workload's IO pattern.

Related to #398

### DIFF
--- a/pkg/fanotify/conn/conn.go
+++ b/pkg/fanotify/conn/conn.go
@@ -8,16 +8,32 @@ package conn
 
 import (
 	"bufio"
+	"encoding/json"
 	"io"
 )
 
 type Client struct {
-	Scanner *bufio.Scanner
+	Reader *bufio.Reader
 }
 
-func (c *Client) GetPath() (string, error) {
-	if !c.Scanner.Scan() { // NOTE: no timeout
-		return "", io.EOF
+type EventInfo struct {
+	Path      string `json:"path"`
+	Size      uint32 `json:"size"`
+	Timestamp int64  `json:"timestamp"`
+}
+
+func (c *Client) GetEventInfo() ([]EventInfo, error) {
+	// Before reaching '\n', the reader has successfully read
+	// the event information from optimizer server.
+	data, err := c.Reader.ReadBytes('\n')
+	if err != nil && err != io.EOF {
+		return nil, err
 	}
-	return c.Scanner.Text(), nil
+
+	eventInfo := []EventInfo{}
+	if err := json.Unmarshal(data, &eventInfo); err != nil {
+		return nil, err
+	}
+
+	return eventInfo, nil
 }
--- a/pkg/fanotify/fanotify.go
+++ b/pkg/fanotify/fanotify.go
@@ -8,8 +8,8 @@ package fanotify
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
-	"io"
 	"log/syslog"
 	"os"
 	"os/exec"
@@ -20,30 +20,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
-
-func StartFanotifier(client *conn.Client, persistFile string) error {
-	f, err := os.OpenFile(persistFile, os.O_CREATE|os.O_WRONLY, 0644)
-	if err != nil {
-		return errors.Wrapf(err, "failed to open persist file %q", persistFile)
-	}
-	var existedFiles = make(map[string]struct{}, 1)
-	for {
-		path, err := client.GetPath()
-		if err != nil {
-			f.Close()
-			if err == io.EOF {
-				break
-			}
-			return fmt.Errorf("failed to get notified path: %v", err)
-		}
-
-		if _, ok := existedFiles[path]; !ok {
-			existedFiles[path] = struct{}{}
-			fmt.Fprintln(f, path)
-		}
-	}
-	return nil
-}
 
 type Server struct {
 	BinaryPath   string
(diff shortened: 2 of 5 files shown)


pkg/fanotify/conn/conn.go
```diff
@@ -8,16 +8,30 @@ package conn
...
+		return nil, io.EOF
 	}
-	return c.Scanner.Text(), nil
+
+	eventInfo := &EventInfo{}
+	err := json.Unmarshal([]byte(c.Scanner.Text()), eventInfo)
+	if err != nil {
+		return nil, err
+	}
+
```
changweige, line 35:
```go
func (c *Client) GetEventInfo() (*EventInfo, error) {
	if !c.Scanner.Scan() { // NOTE: no timeout
		return nil, c.Scanner.Err()
	}

	eventInfo := &EventInfo{}
	err := json.Unmarshal(c.Scanner.Bytes(), eventInfo)
	if err != nil {
		return nil, err
	}

	return eventInfo, nil
}

```

pkg/fanotify/fanotify.go
```diff
@@ -22,13 +23,15 @@ import (
...
 		return errors.Wrapf(err, "failed to open persist file %q", persistFile)
 	}
-	var existedFiles = make(map[string]struct{}, 1)
+	var eventInfoList = []*conn.EventInfo{}
+	var existedFiles = make(map[string]struct{}, 10)
 	for {
-		path, err := client.GetPath()
+		eventInfo, err := client.GetEventInfo()
 		if err != nil {
 			f.Close()
```
changweige, line 36: Why not call close the file by defer statement to clean up when returning from the function.

pkg/fanotify/fanotify.go
```diff
@@ -22,13 +23,15 @@ import (
...
 	f, err := os.OpenFile(persistFile, os.O_CREATE|os.O_WRONLY, 0644)
 	if err != nil {
 		return errors.Wrapf(err, "failed to open persist file %q", persistFile)
 	}
-	var existedFiles = make(map[string]struct{}, 1)
+	var eventInfoList = []*conn.EventInfo{}
+	var existedFiles = make(map[string]struct{}, 10)
 	for {
-		path, err := client.GetPath()
+		eventInfo, err := client.GetEventInfo()
```
changweige, line 34: Just name it to  GetEvent?

pkg/fanotify/fanotify.go
```diff
@@ -37,11 +40,22 @@ func StartFanotifier(client *conn.Client, persistFile string) error {
...
-		if _, ok := existedFiles[path]; !ok {
-			existedFiles[path] = struct{}{}
-			fmt.Fprintln(f, path)
+		if _, ok := existedFiles[eventInfo.Path]; !ok {
+			existedFiles[eventInfo.Path] = struct{}{}
+			fmt.Fprintln(f, eventInfo.Path)
 		}
+		eventInfoList = append(eventInfoList, eventInfo)
 	}
+	fJSON, err := os.OpenFile(persistJSONFile, os.O_CREATE|os.O_WRONLY, 0644)
```
changweige, line 49: Forget to close fJSON?

changweige (commented):
It's better to command optimizer server to gracefully exit rather than send a KILL signal.

pkg/fanotify/conn/conn.go
sctb512: Replaced by `bufio.Reader.ReadString`. Please take a look.

pkg/fanotify/fanotify.go
sctb512: Fixed.

pkg/fanotify/fanotify.go
sctb512: Fixed.

pkg/fanotify/fanotify.go
sctb512: Personally, I prefer `GetEventInfo` because it is not just the `fanotify` event. 😄

sctb512:
> It's better to command optimizer server to gracefully exit rather than send a KILL signal.

Done. Please take a look. Thanks. 😊

pkg/fanotify/conn/conn.go
```diff
@@ -8,16 +8,29 @@ package conn
...
-	if !c.Scanner.Scan() { // NOTE: no timeout
-		return "", io.EOF
+type EventInfo struct {
+	Path      string `json:"path"`
+	Size      uint32 `json:"size"`
+	Timestamp int64  `json:"timestamp"`
+}
+
+func (c *Client) GetEventInfo() ([]EventInfo, error) {
+	data, err := c.Reader.ReadString('\n')
```
changweige, line 25: Why must wait for `\n` ?

changweige: It's better to wait for EOF here

sctb512: Fixed. 

The `bufio.Reader.ReadBytes` is not limited by the buffer size. But it needs to take a `delim` argument, so we pass the '\n'.
Before reaching '\n', the reader has successfully read the event information from the optimizer server because of `io.EOF`.

pkg/fanotify/conn/conn.go
```diff
@@ -8,16 +8,32 @@ package conn
...
+type EventInfo struct {
+	Path      string `json:"path"`
+	Size      uint32 `json:"size"`
+	Timestamp int64  `json:"timestamp"`
+}
+
+func (c *Client) GetEventInfo() ([]EventInfo, error) {
+	// Before reaching '\n', the reader has successfully read
+	// the event information from optimizer server.
+	data, err := c.Reader.ReadBytes('\n')
```
changweige, line 28: Generally, we should not assume what is the special token in the stream. Let's merge this PR now and I will try to improve it later.

changweige (approved):

pkg/fanotify/conn/conn.go
sctb512: Got it.

PR #403, opened 2023-03-08
converter: support parent bootstrap for merge

imeoer:
Add `ParentBootstrapPath` option for `converter.Merge` method, the
option holds the bootstrap path of a parent image.

This option allows merging multiple bootstraps of upper layer with
the bootstrap of a parent image, so that we can implement container
commit operation for nydus image.


changweige (approved):
lgtm

PR #404, opened 2023-03-09
change oom score adj for nydus daemon

sctb512:
Nydus-snapshotter does not have a maximum memory limit by default. This can result in host invalidity. We can fix this problem by adding MemoryMax to nydus-snapshotter.service. We can set OOMScoreAdjust=-1000 to prevent killing nydus-snapshotter on OOM.


pkg/manager/daemon_adaptor.go
```diff
@@ -40,6 +41,10 @@ func (m *Manager) StartDaemon(d *daemon.Daemon) error {
 	defer d.Unlock()
 
 	d.States.ProcessID = cmd.Process.Pid
+	err = oom.ChangeDaemonOOMScoreAdj(d.States.ProcessID, -999)
```
changweige, line 44: Is it possible to make it work as a Daemon method?

sctb512:
A similar solution is #415

PR #407, opened 2023-03-09
fix containerd mount failed when root is a relative path

wllenyj:
When the Root path is set as a relative path, Prepare return a relative path, the containerd will fail to mount.

### DIFF
--- a/config/global.go
+++ b/config/global.go
@@ -10,11 +10,13 @@
 package config
 
 import (
+	"os"
 	"path/filepath"
 	"time"
 
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/nydus-snapshotter/internal/logging"
+	"github.com/containerd/nydus-snapshotter/pkg/utils/mount"
 	"github.com/pkg/errors"
 )
 
@@ -132,5 +134,18 @@ func ProcessConfigurations(c *SnapshotterConfig) error {
 
 	globalConfig.DaemonMode = m
 
+	return nil
+}
+
+func SetupEnvironment(c *SnapshotterConfig) error {
+	if err := os.MkdirAll(c.Root, 0700); err != nil {
+		return errors.Wrapf(err, "create root dir %s", c.Root)
+	}
+
+	realPath, err := mount.NormalizePath(c.Root)
+	if err != nil {
+		return errors.Wrapf(err, "root path invalid")
+	}
+	c.Root = realPath
 	return nil
 }
--- a/cmd/containerd-nydus-grpc/main.go
+++ b/cmd/containerd-nydus-grpc/main.go
@@ -76,6 +76,10 @@ func main() {
 				return errors.Wrap(err, "process configurations")
 			}
 
+			if err := config.SetupEnvironment(&snapshotterConfig); err != nil {
+				return errors.Wrap(err, "setup environment failed")
+			}
+
 			ctx := logging.WithContext()
 			logConfig := &snapshotterConfig.LoggingConfig
 			logRotateArgs := &logging.RotateLogArgs{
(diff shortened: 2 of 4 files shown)


liubogithub (approved):

changweige:
> The patch looks good.
> 
> But I don't understand why this makes the CI fail.

Because of the below error
```
time="2023-03-09T18:46:34Z" level=fatal msg="failed to start nydus-snapshotter" error="validate configurations: Root path invalid.: canonicalise path for /var/lib/containerd-nydus: lstat /var/lib/containerd-nydus: no such file or directory"
```

config/config.go
```diff
@@ -239,6 +240,11 @@ func ValidateConfig(c *SnapshotterConfig) error {
 	if len(c.Root) == 0 {
 		return errors.New("empty root directory")
 	}
+	if realPath, err := mount.NormalizePath(c.Root); err != nil {
+		return errors.Wrapf(err, "Root path invalid.")
+	} else {
+		c.Root = realPath
+	}
```
changweige, line 247: Shall we move the path normalization to `ProcessConfigurations()` where we do all the stuff for tweaking configurations?

changweige:
@imeoer Any chance to take look at the `converter` smoke test? It looks not stable 🥹

config/config.go
wllenyj: Fixed

imeoer:
@changweige @wllenyj Fixed the latest release on image-servcie repo, just help to retry the failed CI.

Makefile
```diff
@@ -109,7 +109,7 @@ check: vet
 
 .PHONY: test
 test:
-	go test -race -v -mod=mod -cover ${PACKAGES}
+	$(SUDO) -E go test -race -v -mod=mod -cover ${PACKAGES}
```
changweige, line 112: Why do we need privilege when performing the unit test? At least for me, I don't have to go toolchain installed for `root` user 😂

config/global.go
```diff
@@ -97,6 +99,19 @@ func GetDaemonProfileCPUDuration() int64 {
...
+		if err != nil {
+			if !os.IsNotExist(errors.Cause(err)) {
+				return errors.Wrapf(err, "Root path invalid.")
+			}
+			if err := os.MkdirAll(c.Root, 0755); err != nil {
+				return errors.Wrapf(err, "create root dir %s", c.Root)
+			}
+		}
+		c.Root = realPath
+		break
```
changweige, line 113: Can we restrict the permissions to `0700`? 
A more straightforward way would be to create the root dir (relative)  and then normalize the relative path. Then reassign it to `c.Root`. It is safe to remove the current code of creating root dir then. Furthermore, we don't have to check the target dir's existence before `MkdirAll`

Moreover, why a relative path is set for a long-lived background service, which makes it a little vague?

Makefile
wllenyj: There are ProcessConfigurations tests in Unit test. And used `/var/lib/` as root path.

changweige: Um. We'd better not pollute the developer's local file system. Directories base UT is better to be performed within a local temporary dir :-)

config/global.go
wllenyj: For users it doesn't matter, but for developers, I just want to simply use my local directory.

Makefile
wllenyj: So, I rewrote a SetupEnvironment function, the single test will not test this function. what do you think?

config/global.go
wllenyj: Or if the user configures a relative path, it will directly return the startup failure, otherwise it will be very difficult to debug when the containerd reports an error.

Makefile
changweige: SGTM

changweige (approved):
LGTM

PR #408, opened 2023-03-10
converter: fix handle of manifest index

imeoer:
- Skip to convert the nydus image manifest/index again;
- Handle the situation when the conversion did not occur;

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -690,7 +690,7 @@ func IsNydusBlobAndExists(ctx context.Context, cs content.Store, desc ocispec.De
 	return IsNydusBlob(ctx, desc)
 }
 
-// IsNydusBlob returns true when the specified descriptor is nydus blob format.
+// IsNydusBlob returns true when the specified descriptor is nydus blob layer.
 func IsNydusBlob(ctx context.Context, desc ocispec.Descriptor) bool {
 	if desc.Annotations == nil {
 		return false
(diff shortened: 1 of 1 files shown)


pkg/converter/convert_unix.go
```diff
@@ -700,6 +700,16 @@ func IsNydusBlob(ctx context.Context, desc ocispec.Descriptor) bool {
 	return hasAnno
 }
 
+// IsNydusBootstrap returns true when the specified descriptor is nydus bootstrap layer.
+func IsNydusBootstrap(ctx context.Context, desc ocispec.Descriptor) bool {
```
sctb512, line 704: The `ctx` is not used, should we remove it?

pkg/converter/convert_unix.go
```diff
@@ -858,6 +877,22 @@ func convertIndex(ctx context.Context, cs content.Store, orgDesc ocispec.Descrip
...
+func isNydusImage(manifest *ocispec.Manifest) bool {
+	layers := manifest.Layers
+	if len(layers) != 0 {
+		desc := &layers[len(layers)-1]
+		if desc.Annotations == nil {
+			return false
+		}
+		if _, ok := desc.Annotations[LayerAnnotationNydusBootstrap]; ok {
+			return true
+		}
```
sctb512, line 891: It seems that we can replace this with the `IsNydusBootstrap` method?

imeoer: Thanks! Fixed.

pkg/converter/convert_unix.go
imeoer: Fixed.

changweige (approved):

PR #413, opened 2023-03-15
fix bugs in image optimizer and improve the code a bit

changweige:
We should always close event.fd even with error returned.
It's clear to show elapsed time than epoch time as the timestamp
Try to get rid of crate `chrono`, just use Rust std::time

### DIFF
--- a/tools/optimizer-server/src/main.rs
+++ b/tools/optimizer-server/src/main.rs
@@ -192,13 +192,18 @@ fn handle_fanotify_event(fd: i32) {
     }
 
     loop {
-        if let Ok(poll_num) = poll(&mut fds, -1) {
-            if poll_num > 0 {
+        match poll(&mut fds, -1) {
+            Ok(polled_num) => {
+                if polled_num <= 0 {
+                    eprintln!("polled_num <= 0!");
+                    break;
+                }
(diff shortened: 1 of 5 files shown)


tools/optimizer-server/src/main.rs
```diff
@@ -45,7 +45,7 @@ struct fanotify_event {
 struct EventInfo {
     path: String,
     size: u64,
-    timestamp: i64,
+    timestamp: u128,
```
sctb512, line 48: Keep the timestamp type in 'pkg/fanotify/conn/conn.go' the same as 'u128'?

sctb512: Also, can the 'timestamp' be changed to 'elapsed'?

sctb512 (commented):
Others, LGTM

tools/optimizer-server/src/main.rs
changweige: Unfortunately, golang does not have uint128 type. uint64 should be large enough

changweige: renamed

sctb512 (approved):

PR #415, opened 2023-03-16
Limit memory resource usage for nydus daemon via cgroup

sctb512:
Nydus-snapshotter does not have a maximum memory limit by default. This can lead to host invalidity. Now, we put snapshotter and daemon in different cgroup to limit the maximum available memory.

This PR adds two entries to the configuration file.
```toml
[cgroup]
# Whether to use separate cgroup for nydusd.
enable = true
# The memory limit for nydusd.slice, which contains all nydusd processes.
# Percentage is supported as well, please ensure it is end with "%".
# The default unit is bytes. Acceptable values include "209715200", "200MiB", "200Mi" and "10%".
memory_limit = ""
```


pkg/daemon/daemon.go
```diff
@@ -559,6 +562,38 @@ func (d *Daemon) ClearVestige() {
 	d.ResetClient()
 }
 
+func (d *Daemon) AddCgroup(pid int) error {
```
imeoer, line 565: I think we should implement a cgroup manager and add the daemon to it, instead add the method `AddCgroup`, `DeleteCgroup` on `*Daemon`. One reason is that maybe we need to support containerized snapshotter/nydusd in the future.

changweige (commented):
Is it based on cgroup v1 or v2?

pkg/daemon/daemon.go
sctb512: Got it. I will refactor this later.

sctb512:
> Is it based on cgroup v1 or v2?

It is based on cgroup v1.

changweige:
> > Is it based on cgroup v1 or v2?
> 
> It is based on cgroup v1.

Is it possible to depend cgroup v2?

sctb512:
> Is it possible to depend cgroup v2?

May we implement both v1 and v2? Group v2 depends on the Linux kernel. On my machine, it mounts the cgroup v1 file system by default.

pkg/daemon/daemon.go
changweige: Agree. We should support resources limitation based on both raw cgroup controller and containers

snapshot/snapshot.go
```diff
@@ -82,6 +84,19 @@ func NewSnapshotter(ctx context.Context, cfg *config.SnapshotterConfig) (snapsho
 		return nil, errors.Wrap(err, "parse recover policy")
 	}
 
+	cgroupConfig, err := config.ParseCgroupConfig(cfg.DaemonConfig.CgroupConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "parse cgroup configuration")
+	}
+	log.L.Infof("parsed cgroup config: %#v", cgroupConfig)
+	cgroupMgr, err := cgroup.NewManager(cgroup.Opt{
+		Name:   "nydusd.slice",
```
changweige, line 93: `.slice` is a systemd style that I don't think we have to follow

sctb512: Fixed.

snapshot/snapshot.go
```diff
@@ -82,6 +84,19 @@ func NewSnapshotter(ctx context.Context, cfg *config.SnapshotterConfig) (snapsho
...
+	cgroupConfig, err := config.ParseCgroupConfig(cfg.CgroupConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "parse cgroup configuration")
+	}
+	log.L.Infof("parsed cgroup config: %#v", cgroupConfig)
+	cgroupMgr, err := cgroup.NewManager(cgroup.Opt{
+		Name:   "nydusd",
+		Config: cgroupConfig,
+	})
+	if err != nil && (err != cgroup.ErrCgroupNotSupported || err != v2.ErrRootMemorySubtreeControllerNotEnableed) {
```
changweige, line 96: typo “Enableed”  or this can be named to xxxDisabled

pkg/utils/sysinfo/sysinfo.go
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package sysinfo
+
+import "syscall"
+
+var (
+	sysinfo *syscall.Sysinfo_t
+)
```
changweige, line 13: Can you try sync.Once to initialize the global sysinfo?

changweige: It somewhat not threaded safe

pkg/cgroup/v2/v2.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (cg Cgroup) Delete() error {
+	if cg.manager != nil {
+		return cg.manager.Delete()
+	}
+	return nil
+}
+func (cg Cgroup) AddProc(pid int) error {
+	if cg.manager != nil {
+		err := cg.manager.AddProc(uint64(pid))
```
changweige, line 87: Need move the process out from controller if nydusd dies or exits normally?

sctb512: The process will be automatically removed if nydusd dies. The cgroups library in containerd does not have this method either.

snapshot/snapshot.go
sctb512: Good point. Fixed.

changweige (approved):

PR #416, opened 2023-03-16
remove useless blob manager

imeoer:
This feature is used to preheat the download of nydus blob data.
It is no longer used by any users and should be removed to
reduce the dead code.


changweige (commented):
Overall looks good. We should migrate to containerd transfer service if any special artifact is needed

changweige:
Could you also remove the e2e case `blob_manager_pull_preheat` ?

changweige (approved):
LGTM

PR #419, opened 2023-03-16
misc: update stable nydusd version v2.1.5 in Dockerfile and docs

adamqqqplay:
We should upgrade the version of nydus to v2.1.5 to fix known issues.


changweige (approved):

adamqqqplay:
CI fails with network timeout and needs to be re-run.

changweige (approved):

PR #423, opened 2023-03-20
Simplify the logics preparing snapshots for containerd

changweige:
Right now, apart from making room for snapshot, snapshotter is also processing the blobs by converting or inspecting it. In the future, we may have more processing logic like them. So simplify and refactor related code for future work


imeoer (approved):
LGTM!

PR #426, opened 2023-03-20
experimental referrers support

imeoer:
Add the new `enable_referrers` option in config.toml, to enable trying
to fetch the nydus image associated with the OCI image and run it.

This feature relies on the support of [OCI reference type](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) in OCI
distribution-spec 1.1. Currently, only Harbor (>=v2.8) registry and
ORAS registry supports this. Additionally, we need to rely on containerd's
referrer fetcher interface, related links:

https://github.com/containerd/nydus-snapshotter/pull/414
https://github.com/containerd/containerd/pull/7644
https://oras.land/cli/6_reference_types/


go.mod
```diff
@@ -153,3 +153,8 @@ retract (
 	v0.11.0
 	v0.3.0 // retagged: cd604c1b597558ea045a79c4f80a8c780909801b -> 85653575c7dafb6b06548478ee1dc61ac5905d00
 )
+
+// We are using the forked containerd package version for enabling
+// OCI reference type feature. Once containerd officially releases
+// this feature, we will remove the replacement here.
+replace github.com/containerd/containerd => github.com/nydusaccelerator/containerd v0.0.0-20230331034304-d4d4bdc00ef9
```
changweige, line 160: Maybe we'd better merge this PR into a feature branch with the go.mod `replace` directive. When containerd has released the support, we can safely merge it back to the main branch. Otherwise, we can hardly get rid the experimental feature and use normal containerd

pkg/filesystem/config.go
```diff
@@ -68,3 +80,10 @@ func WithEnableStargz(enable bool) NewFSOpt {
 		return nil
 	}
 }
+
+func WithEnableReferrers(enable bool) NewFSOpt {
+	return func(fs *Filesystem) error {
+		fs.enableReferrers = enable
```
changweige, line 86: Can we just use `fs.referrerMgr` to judge if the referrer is enabled?

pkg/referrer/referrer.go
```diff
@@ -0,0 +1,163 @@
...
+			return "", "", nil
+		}
+		return keyChain.Username, keyChain.Password, nil
+	}
+
+	newClient := func(insecure bool) *http.Client {
+		client := http.DefaultClient
+		transport := http.DefaultTransport.(*http.Transport)
+		transport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: insecure,
```
changweige, line 52: Should default to secured transport.

pkg/referrer/referrer.go
```diff
@@ -0,0 +1,163 @@
...
+					docker.WithAuthCreds(credFunc),
+				),
+			),
+			docker.WithClient(newClient(insecure)),
+			docker.WithPlainHTTP(func(host string) (bool, error) {
+				return plainHTTP, nil
+			}),
+		)
+
+		return docker.NewResolver(docker.ResolverOptions{
```
changweige, line 72: Can we extract out the logic fetching OCI artifacts into a separated package which can be reused?

pkg/referrer/unpack.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/containerd/containerd/archive/compression"
+)
+
+// unpackFromTar unpacks the file named `source` in tar stream
+// and write into `target` path.
+func unpackFromTar(reader io.Reader, source, target string) error {
```
changweige, line 20: It's better to move this functionality to a package handling OCI artifacts.

snapshot/process.go
```diff
@@ -61,13 +68,10 @@ func chooseProcessor(ctx context.Context, logger *logrus.Entry,
...
-		// Hope to find bootstrap layer and prepares to start nydusd
-		// TODO: Trying find nydus meta layer will slow down setting up rootfs to OCI images
-		if id, info, err := sn.findMetaLayer(ctx, key); err == nil {
-			logger.Infof("Prepares active snapshot %s, nydusd should start afterwards", key)
-			handler = func() (bool, []mount.Mount, error) {
+		remoteHandler := func(id string, lables map[string]string) func() (bool, []mount.Mount, error) {
+			return func() (bool, []mount.Mount, error) {
 				logger.Debugf("Found nydus meta layer id %s", id)
-				if err := sn.prepareRemoteSnapshot(id, info.Labels); err != nil {
+				if err := sn.prepareRemoteSnapshot(id, lables); err != nil {
```
changweige, line 74: The referenced nydus bootstrap can be downloaded the most bottom layer is prepared

config/config.go
```diff
@@ -95,7 +95,8 @@ const (
 )
 
 type Experimental struct {
-	EnableStargz bool `toml:"enable_stargz"`
+	EnableStargz    bool `toml:"enable_stargz"`
+	EnableReferrers bool `toml:"enable_referrers"`
```
changweige, line 99: Can we hide the details that we use OCI referer API? How about renaming it to `oci_subject`, which means oci image manifest has a subject field associated with nydus image

go.mod
changweige: Then we can create the feature branch "oci_subject"

pkg/referrer/referrer.go
```diff
@@ -0,0 +1,163 @@
...
+	return &referrer{
+		resolverFunc:  resolverFunc,
+		WithPlainHTTP: true,
+	}
+}
+
+// checkReferrer fetches the referrers and parse out the nydus
+// image by specified manifest digest.
+// it's using distribution list referrers API.
+func (r *referrer) checkReferrer(ctx context.Context, ref string, manifestDigest digest.Digest) (*ocispec.Descriptor, error) {
```
changweige, line 86: It's better to make this a common OCI artifact downloading capability, not only for downloading referenced artifacts.

changweige:
Is ORAS only a CLI client? Or it provides artifacts storage and distribution service either?

pkg/filesystem/config.go
imeoer: Fixed

pkg/referrer/referrer.go
imeoer: Updated, should follow with the `SkipVerify` option.

pkg/referrer/referrer.go
imeoer: Fixed, moved to `pkg/remote`.

pkg/referrer/unpack.go
imeoer: Fixed, moved to `pkg/remote`.

config/config.go
imeoer: Updated to `enable_referrer_detect`.

pkg/referrer/referrer.go
imeoer: Fixed, moved to `pkg/remote`.

imeoer:
> Is ORAS only a CLI client? Or it provides artifacts storage and distribution service either?

It also provides [distribution](https://github.com/oras-project/distribution) implementation.

go.mod
imeoer: Fixed, moved the ported codes to `pkg/remote/remotes`.

pkg/filesystem/referer_adaptor.go
```diff
@@ -0,0 +1,56 @@
...
+		return false
+	}
+
+	return true
+}
+
+func (fs *Filesystem) TryFetchMetadata(ctx context.Context, labels map[string]string, metadataPath string) error {
+	ref, ok := labels[snpkg.TargetRefLabel]
+	if !ok {
+		return fmt.Errorf("invalid label %s=%s", snpkg.TargetRefLabel, ref)
```
changweige, line 43: `ref` must be its zero value. The error message seems meaningless

pkg/referrer/manager.go
```diff
@@ -0,0 +1,88 @@
...
+	return &manager
+}
+
+// CheckReferrer attempts to fetch the referrers and parse out
+// the nydus image by specified manifest digest.
+func (manager *Manager) CheckReferrer(ctx context.Context, ref string, manifestDigest digest.Digest) (*ocispec.Descriptor, error) {
+	metaLayer, err, _ := manager.sg.Do(manifestDigest.String(), func() (interface{}, error) {
+		// Try to get nydus metadata layer descriptor from LRU cache.
+		if metaLayer, ok := manager.cache.Get(manifestDigest); ok {
+			desc := metaLayer.(ocispec.Descriptor)
```
changweige, line 43: Should check if type assertion is failed?

pkg/referrer/referrer.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// Unpack nydus metadata file to specified path.
+		rc, err := fetcher.Fetch(ctx, desc)
+		if err != nil {
+			return errors.Wrap(err, "fetch nydus metadata")
+		}
+		defer rc.Close()
+
+		if err := remote.UnpackFromTar(rc, metadataNameInLayer, metadataPath); err != nil {
```
changweige, line 120: Can we rename `UpanckFromTar` to `Unpack`? It also decompess the fetched content

pkg/filesystem/stargz_adaptor.go
```diff
@@ -36,6 +36,10 @@ func (fs *Filesystem) StargzEnabled() bool {
 	return fs.stargzResolver != nil
 }
 
+func (fs *Filesystem) ReferrerDetectEnabled() bool {
```
changweige, line 39: This function should not locate at stargz adopter.

pkg/filesystem/referer_adaptor.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"fmt"
+
+	snpkg "github.com/containerd/containerd/pkg/snapshotters"
+	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
+)
+
+func (fs *Filesystem) CheckReferrer(ctx context.Context, labels map[string]string) bool {
+	if !fs.ReferrerDetectEnabled() {
```
changweige, line 19: It's better not to bind Referrer manger  with Filesystem since filesystem is a very thin layer to adapt to snapshots layer

changweige (commented):
Otherwise, looks good to me. A note would be that we have to drop the copied remotes package from containerd once containerd releases the OCI artifact referer support.

pkg/filesystem/referer_adaptor.go
imeoer: Fixed

pkg/referrer/manager.go
imeoer: No need, it must be `ocispec.Descriptor` if `ok == true`.

pkg/referrer/referrer.go
imeoer: Changed `UpanckFromTar` to `Unpack`.
The downloaded artifact is put into snapshot directory, I think it will be GCed by `Cleanup` interface.

pkg/filesystem/stargz_adaptor.go
imeoer: Fixed.

pkg/filesystem/referer_adaptor.go
imeoer: It seems `cacheMgr` and `stargzResolver` still also exist here..

changweige: Fine, I am planning to refactor them. Let's keep it the same way now.

PR #431, opened 2023-03-21
[stable/v0.7] converter: support OCI reference type

imeoer:
Associate a reference to the original OCI manifest, see the
`subject` field description in the doc:
https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions

And if the converted manifest list contains only one manifest,
then convert it directly to manifest.


changweige (approved):
LGTM

PR #434, opened 2023-03-28
Misc improvements about enhancing nydus-snapshotter's stability and add more debug message

changweige:
Before querying nydusd's state, nydus-snapshotter is checking periodically nydus'd s API socket's existence. It should abort  immediately if nydud dies before creating such Unix socket file.

We should only collect metrics that come from normally serving nydusd.


adamqqqplay:
Maybe we could squash these commits? 10 commits in 1 PR seems like a lot.

pkg/daemon/client.go
```diff
@@ -175,7 +177,20 @@ func WaitUntilSocketExisted(sock string) error {
 	},
 		retry.Attempts(100), // totally wait for 10 seconds, should be enough
 		retry.LastErrorOnly(true),
-		retry.Delay(100*time.Millisecond))
+		retry.Delay(100*time.Millisecond),
+		retry.OnlyRetryIf(func(_ error) bool {
```
imeoer, line 181: Just `func (error) bool`

imeoer (approved):

PR #446, opened 2023-04-11
workflow: fix incorrect files number for optimizer test

sctb512:
We use `nginx:1.23.3` to test the optimizer, but sometimes the number of files accessed is not equal to 44. Normally, the optimizer works fine if the number of files accessed is greater than 0.

### DIFF
--- a/.github/workflows/optimizer.yml
+++ b/.github/workflows/optimizer.yml
@@ -76,15 +76,16 @@ jobs:
           sudo systemctl restart containerd
       - name: Generate accessed files list
         run: |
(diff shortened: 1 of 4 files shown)


sctb512:
This PR could solve the above problem. But I think the more sensible solution is to check why the number of files accessed is incorrect.
Just wait.

sctb512:
The reason why the test sometimes fails is that `nginx` generates a temporary file that may not be captured by the `fanotify` server.
```
/etc/nginx/conf.d/sed5PCVAq
```

It replaced by a simple workload `echo hello`.

changweige (approved):
LGMT, thanks

PR #447, opened 2023-04-12
support nydus image topmost layer view

changweige:
Without inline layer fs meta blob, we can still support the topmost View for nydus image. It is useful for Buildkit to use nydus image as the base image.

Next PR will implement View on any layer of the nydus image.

```
➜  ~ ctr snapshot view --mounts cccddd sha256:62b1c52c224c7907ca34194f2c7e819b8057251b6d69623ff3bf8a7c9393894a
[
    {
        "Type": "overlay",
        "Source": "overlay",
        "Target": "",
        "Options": [
            "lowerdir=/var/lib/containerd-nydus/snapshots/510/mnt:/var/lib/containerd-nydus/snapshots/536/fs"
        ]
    }
]

```

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -394,7 +394,7 @@ func (o *snapshotter) View(ctx context.Context, key, parent string, opts ...snap
 		metaSnapshotID   string
 	)
 
-	if isNydusMetaLayer(pInfo.Labels) {
+	if label.IsNydusMetaLayer(pInfo.Labels) {
 		// Nydusd might not be running. We should run nydusd to reflect the rootfs.
 		if err = o.fs.WaitUntilReady(pID); err != nil {
 			if errors.Is(err, errdefs.ErrNotFound) {
@@ -412,7 +412,7 @@ func (o *snapshotter) View(ctx context.Context, key, parent string, opts ...snap
 
 		needRemoteMounts = true
 		metaSnapshotID = pID
-	} else if isNydusDataLayer(pInfo.Labels) {
+	} else if label.IsNydusDataLayer(pInfo.Labels) {
 		return nil, errors.New("only can view nydus topmost layer")
 	}
 	// Otherwise, it is OCI snapshots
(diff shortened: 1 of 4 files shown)


sctb512:
Fantastic! This PR is useful for mounting a nydus image into a directory with `ctr i mount`. 
I also recommend improving the `Prepare` API. Because the command with `--rw` (`ctr i mount --rw IMAGE TARGET` ) does not work now.
The `ctr` source code about it:
```go
if context.Bool("rw") {
	mounts, err = s.Prepare(ctx, target, chainID)
} else {
	mounts, err = s.View(ctx, target, chainID)
}
```

Here is a comparison with stargz:
```shell
mkdir mkdir test.bootstrap.stargz
nerdctl --snapshotter stargz pull ghcr.io/stargz-containers/alpine:3.15.3-esgz
ctr --debug i mount --rw --snapshotter stargz ghcr.io/stargz-containers/alpine:3.15.3-esgz test.bootstrap.stargz
ls test.bootstrap.stargz
```
Output:
```
bin dev etc home lib media mnt opt proc root run sbin srv sys tmp usr var
```

Unfortunately this does not work for nydus:
```shell
nerdctl --snapshotter nydus pull sctb512/golang:nydusv6
ctr --debug i mount --rw --snapshotter nydus docker.io/sctb512/golang:nydusv6 test.bootstrap.nydus
ls test.bootstrap.nydus
```
Output: None

snapshot/snapshot.go
```diff
@@ -353,14 +354,55 @@ func (o *snapshotter) findMetaLayer(ctx context.Context, key string) (string, sn
...
+
+	var (
+		needRemoteMounts = false
+		metaSnapshotID   string
+	)
+
+	if isNydusMetaLayer(pInfo.Labels) {
+		// Nydusd might not be running. We should run nydusd to reflect the rootfs.
+		err = o.fs.WaitUntilReady(pID)
+		if err != nil {
```
imeoer, line 374: `if err := o.fs.WaitUntilReady(pID); err != nil {`

changweige:
> o.fs.WaitUntilReady(pID)

`Prepare` is now implemented

snapshot/snapshot.go

imeoer (approved):

PR #450, opened 2023-04-18
Avoid optimizer to use too much memory and improve error handling processes

sctb512:
1. To avoid using too much memory in the `containerd` cgroup, we should send event information one by one.
2. Add an option to make the result human-readable.
3. Refine the code to better handle errors on the optimizer server.

### DIFF
--- a/cmd/optimizer-nri-plugin/main.go
+++ b/cmd/optimizer-nri-plugin/main.go
@@ -40,6 +40,7 @@ type PluginConfig struct {
 
 	ServerPath string `toml:"server_path"`
 	PersistDir string `toml:"persist_dir"`
+	Readable   bool   `toml:"readable"`
 	Timeout    int    `toml:"timeout"`
 	Overwrite  bool   `toml:"overwrite"`
 }
@@ -85,6 +86,12 @@ func buildFlags(args *PluginArgs) []cli.Flag {
 			Value:       defaultPersistDir,
 			Usage:       "the directory to persist accessed files list for container",
 			Destination: &args.Config.PersistDir,
+		},
+		&cli.BoolFlag{
+			Name:        "readable",
+			Value:       false,
+			Usage:       "whether to make the csv file human readable",
+			Destination: &args.Config.Readable,
 		},
 		&cli.IntFlag{
 			Name:        "timeout",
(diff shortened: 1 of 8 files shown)


tools/optimizer-server/src/main.rs
```diff
@@ -163,36 +164,41 @@ fn generate_event_info(path: &Path) -> Result<EventInfo, io::Error> {
...
+    let mut event_duplicate = Vec::new();
+
+    let term = Arc::new(AtomicBool::new(false));
+    if let Err(e) = signal_hook::flag::register(signal_hook::consts::SIGTERM, Arc::clone(&term)) {
         eprintln!("failed to set SIGTERM signal handler {e:?}");
         return;
     }
 
     loop {
+        if term.load(Ordering::Relaxed) {
```
changweige, line 197: `Ordering::Relaxed` is not reliable in the case breaking this loop

changweige: Should also wake up the Poller when receiving a TERM signal

cmd/optimizer-nri-plugin/main.go
```diff
@@ -86,6 +87,12 @@ func buildFlags(args *PluginArgs) []cli.Flag {
 			Usage:       "the directory to persist accessed files list for container",
 			Destination: &args.Config.PersistDir,
 		},
+		&cli.BoolFlag{
+			Name:        "readable",
```
changweige, line 91: Is CSV the only output format now?

pkg/utils/display/display.go
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) 2023. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package display
+
+import "fmt"
+
+func ByteToReadableIEC(b uint32) string {
```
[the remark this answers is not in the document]
sctb512: International Electrotechnical Commission. A [standard](https://en.wikipedia.org/wiki/Byte) for converting units.

For example, when converting `KiB` to `B`, it should be divided by 1024 instead of 1000.

cmd/optimizer-nri-plugin/main.go
sctb512: In addition to the line-by-line file path list, Yes.

tools/optimizer-server/src/main.rs
sctb512: Fixed.

changweige (approved):
LGTM

changweige:
It would be better to rearrange the project layout in the future

PR #462, opened 2023-05-03
manager: simplify implementation of DaemonStates

jiangliu:
Simplify implementation of DaemonStates by removing field `daemons`.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -29,16 +29,13 @@ import (
 )
 
 type DaemonStates struct {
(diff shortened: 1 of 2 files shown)


changweige (approved):
LGTM. 
Looks that `idxBySnapshotID` is not needed anymore too

jiangliu:
> LGTM. Looks like that `idxBySnapshotID` is not needed anymore too

removed too:)

changweige (approved):
LGTM

PR #465, opened 2023-05-06
Prepare 0.8.1

changweige:
Downgrade some packages in order to avoid introducing too many vendor files to Buildkit


adamqqqplay:
Maybe we need cherry-pick this to stable/v0.8 branch?

changweige:
> Maybe we need cherry-pick this to stable/v0.8 branch?

Yes. But we are still investigating why nydus-snapshotter bumps so many deps of Buildkit. Perhaps we don't nee to merge this PR if we figure out a way to avoid introduce deps to Buildkit

changweige:
This is not needed anymore

PR #466, opened 2023-05-06
Maintain daemon mode information for each daemon instance

jiangliu:
Currently daemon mode is a global configuration, all daemon instances must be with the same daemon mode, which limits some possible usage scenarios. So add `DaemonMode` to daemon object, so daemons may have different daemon modes.

Aslo add an alias for daemon mode "multiple" as "dedicated".


changweige:
need rebase 🥺

misc/snapshotter/config.toml
```diff
@@ -78,6 +78,7 @@ sync_remove = false
 [cache_manager]
 disable = false
 gc_period = "24h"
+# Directory to host cached files for fusedev
```
changweige, line 81: fscache also stores cache here?

internal/flags/flags.go
```diff
@@ -70,7 +70,7 @@ func buildFlags(args *Args) []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:        "daemon-mode",
-			Usage:       "nydusd daemon working mode, possible values: \"multiple\", \"shared\" or \"none\"",
+			Usage:       "nydusd daemon working mode, possible values: \"dedicated\"/\"multiple\", \"shared\" or \"none\"",
```
changweige, line 73: Should we remind user that `dedicated` is a alias of `multiple` while `multiple` has been deprecated?

pkg/daemon/daemon.go
```diff
@@ -37,13 +37,14 @@ const (
...
 type States struct {
-	// Generated by daemons manager as a unique to identify a nydusd
+	// A unique ID generated by daemon manager to identify the nydusd instance.
 	ID          string
 	ProcessID   int
 	APISocket   string
 	LogDir      string
 	LogLevel    string
 	LogToStdout bool
+	DaemonMode  config.DaemonMode
```
changweige, line 47: Do we need to handle the compatibility if the newer nydus-snapshotter is loading old records in DB without `DamonMode`  ?

internal/flags/flags.go
jiangliu: How about this?
```
"nydusd daemon working mode, possible values: \"dedicated\", \"multiple\", \"shared\" or \"none\". \"multiple\" is an alias of \"dedicated\" and will be deprecated in v1.0"
```

pkg/daemon/daemon.go
jiangliu: Yes, we need to migrate the database, that's on the todo list.

jiangliu: Currently it works, but may get broken in future easily. So the safe way is to convert records in databases. `Rafs` will face the same situation in coming PRs, so let's delay support of data migration.

misc/snapshotter/config.toml
jiangliu: Fixed

pkg/daemon/daemon.go
changweige: Um. I tend to tolerate the DB compatibility within this PR rather than make it a future work.
For the compatibility, I suppose it wouldn't be very complicated. Nydus-sanpshotter already has an experience on how to migrate different record formats in function `tryTranslateRecords()`. We can simply assign a DaemonMode by comparing daemon `Mountpoint` and `HostRootMount` and rewrite the daemon's record of state

changweige (commented):
Otherwise LGTM

internal/flags/flags.go
changweige: SGTM

pkg/daemon/daemon.go
jiangliu: Added a patch to migrate the database.

imeoer (approved):

bergwolf:
> Aslo add an alias for daemon mode "multiple" as "dedicated".

Why do we need to introduce the alias?

jiangliu:
> > Aslo add an alias for daemon mode "multiple" as "dedicated".
> 
> Why do we need to introduce the alias?

We are planning to deprecated `multiple`.

changweige (approved):

changweige:
> > Aslo add an alias for daemon mode "multiple" as "dedicated".
> 
> Why do we need to introduce the alias?

Feel like `dedicated` is more suggestive and descriptive 😀

PR #467, opened 2023-05-06
metrics: prepare to support multiple instances of Manager

jiangliu:
Enhance metrics server to support multiple instances of Manager.


changweige (commented):
Why do we need multiple daemon managers?

jiangliu:
> Why do we need multiple daemon managers?

It's a preparation for this work: https://github.com/containerd/nydus-snapshotter/pull/472/commits/2a8da1226cb6e64f1369f1553edc00e0baeff2b9

imeoer (approved):

PR #468, opened 2023-05-06
cache: simplify cache manager by removing FsDriver

jiangliu:
The FsDriver field is never used, so remove it.

### DIFF
--- a/pkg/cache/manager.go
+++ b/pkg/cache/manager.go
@@ -27,18 +27,17 @@ const (
 	dataFileSuffix = ".blob.data"
 )
 
+// Disk cache manager for fusedev.
 type Manager struct {
 	cacheDir string
 	period   time.Duration
 	eventCh  chan struct{}
-	fsDriver string
 }
 
 type Opt struct {
 	CacheDir string
 	Period   time.Duration
 	Database *store.Database
-	FsDriver string
 }
 
 func NewManager(opt Opt) (*Manager, error) {
@@ -52,7 +51,6 @@ func NewManager(opt Opt) (*Manager, error) {
 		cacheDir: opt.CacheDir,
 		period:   opt.Period,
 		eventCh:  eventCh,
-		fsDriver: opt.FsDriver,
 	}
 
 	return m, nil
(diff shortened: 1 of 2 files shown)


changweige:
FsDriver is associated with cache manager because fuse driver and fscache driver organize local cache files in different manner. So method `CacheUsage` has to distinguish what fs driver it is working on. Currently we don't implement how to count disk usage for fscache driver  which means some client tools like nerdctl can't report disk usage under fscache driver.

I think rather than remove the field, we can go further to implement it for fscache driver 

See the SIZE is zero:

```console
REPOSITORY                                   TAG                IMAGE ID        CREATED           PLATFORM       SIZE     BLOB SIZE
cr-cn-guilin-boe.volces.com/vci_qa/golang    latest-nydus-v6    08fe692dc892    33 seconds ago    linux/amd64    0.0 B    346.8 MiB
```

changweige:
> My plan is to use the same cache directory for all of tarfs/fusedev/fscache, and enhance CacheManager to understand the directory structure of fscache/cachefiles. It won't depend on FsDriver anymore.

At present, fusedev and fsache actually use the same cache directory. I am curious how the cache manager can distinguish how to count disk usage for different drivers. Ideally, nydus-snapshotter can query the cache size of fscache via API to nydusd as such how cache is deleted for fscache

jiangliu:
> > My plan is to use the same cache directory for all of tarfs/fusedev/fscache, and enhance CacheManager to understand the directory structure of fscache/cachefiles. It won't depend on FsDriver anymore.
> 
> At present, fusedev and fsache actually use the same cache directory. I am curious how the cache manager can distinguish how to count disk usage for different drivers. Ideally, nydus-snapshotter can query the cache size of fscache via API to nydusd as such how cache is deleted for fscache

The fscache/cachefiles driver saves actual data file under /var/lib/containerd-nydus-grpc/cache/cache/__with_magic__directory_and_file_name__.
We only need to decode the magic directory and file, then we can stat disk space usage.

changweige (approved):

PR #469, opened 2023-05-07
referrer: tune implementation details of referrer

jiangliu:
Refine implementation of referrer by:
- avoid retry if error is not network related
- limit maximum size of manifest index blob
- remove partially downloaded meta blob

### DIFF
--- a/pkg/referrer/referrer.go
+++ b/pkg/referrer/referrer.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/containerd/nydus-snapshotter/pkg/auth"
 	"github.com/containerd/nydus-snapshotter/pkg/label"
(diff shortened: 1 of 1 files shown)


pkg/referrer/referrer.go
```diff
@@ -54,7 +59,7 @@ func (r *referrer) checkReferrer(ctx context.Context, ref string, manifestDigest
 
 		// Parse image manifest list from referrers.
 		var index ocispec.Index
-		bytes, err := io.ReadAll(rc)
+		bytes, err := io.ReadAll(io.LimitReader(rc, 0x100000))
```
imeoer, line 62: It seems we can't suppose the manifest index size.

pkg/referrer/referrer.go
```diff
@@ -93,7 +98,7 @@ func (r *referrer) checkReferrer(ctx context.Context, ref string, manifestDigest
 	}
 
 	desc, err := handle()
-	if err != nil && r.remote.RetryWithPlainHTTP(ref, err) {
+	if err != nil && retry && r.remote.RetryWithPlainHTTP(ref, err) {
```
imeoer, line 101: the `RetryWithPlainHTTP` func will avoid retry if the error is not network related.

pkg/referrer/referrer.go
jiangliu: I would prefer a reasonable limit here, otherwise it may be an attacker point for DoS.

imeoer (approved):

PR #470, opened 2023-05-07
converter: restrict file permission to 0640

jiangliu:
Restrict permission of files generated by converter from 0644 to 0640.


changweige (approved):
Nit: with open flag O_RDONLY, creation file mode 0440 should be enough

imeoer (approved):

PR #475, opened 2023-05-10
pkg: minor code style improvements

jiangliu:
Minor code style improvements, should be no functional changes.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -10,6 +10,7 @@ package daemon
 import (
 	"os"
 	"path/filepath"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"syscall"
@@ -46,8 +47,7 @@ type States struct {
 	LogToStdout bool
 	DaemonMode  config.DaemonMode
 	FsDriver    string
-	// Host kernel mountpoint, only applies to fuse fs driver. The fscache fs driver
-	// doesn't need a host kernel mountpoint.
+	// Fusedev mountpoint on host kernel, the fscache fs driver doesn't need a host kernel mountpoint.
 	Mountpoint string
 	ThreadNum  int
 	// Where the configuration file resides, all rafs instances share the same configuration template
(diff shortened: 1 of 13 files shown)


pkg/daemon/daemon.go
```diff
@@ -60,16 +60,17 @@ type Daemon struct {
 	States States
 
 	mu sync.Mutex
-	// FsInstances map[int]*Rafs
-	// should be persisted to DB
-	// maps to at least one rafs instance.
-	// It is possible to be empty after the daemon object is created.
+	// Host all RAFS filesysstems managed by this daemon:
```
changweige, line 63: Typo filesystems ?

pkg/system/system.go
```diff
@@ -148,6 +148,21 @@ func NewSystemController(fs *filesystem.Filesystem, manager *manager.Manager, so
 	return &sc, nil
 }
 
+func (sc *Controller) Run() error {
```
changweige, line 151: Um, why several functions like this `Run` are moved within the same source file, which makes the review a little harder?

pkg/daemon/rafs.go
```diff
@@ -95,18 +94,18 @@ func (rs *rafsSet) List() map[string]*Rafs {
 	return instances
 }
 
-func (rs *rafsSet) ListUnlocked() map[string]*Rafs {
+func (rs *rafsSet) ListLocked() map[string]*Rafs {
```
changweige, line 97: `Locked` suffix indicates this function has lock inside its body?

jiangliu: The `Locked` suffix means this function should be called with related lock held. It's a code style from Linux kernel:)

pkg/system/system.go
jiangliu: I'm trying to group logically related functions near to each other.

pkg/daemon/daemon.go
jiangliu: Fixed

imeoer (approved):
LGTM.

PR #476, opened 2023-05-10
Minor enhancements and bugfixes to snapshot

jiangliu:
Minor enhancements and bugfixes to snapshot.

### DIFF
--- a/snapshot/process.go
+++ b/snapshot/process.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/mount"
 	snpkg "github.com/containerd/containerd/pkg/snapshotters"
 	"github.com/containerd/containerd/snapshots/storage"
@@ -80,6 +81,7 @@ func chooseProcessor(ctx context.Context, logger *logrus.Entry,
 					return false, nil, err
 				}
 
+				log.L.Infof("Nydus remote snapshot %s is ready", id)
 				mounts, err := sn.remoteMounts(ctx, s, id)
 				return false, mounts, err
 			}
--- a/snapshot/process.go
+++ b/snapshot/process.go
@@ -27,27 +27,48 @@ func chooseProcessor(ctx context.Context, logger *logrus.Entry,
 	sn *snapshotter, s storage.Snapshot,
 	key, parent string, labels map[string]string, storageLocater func() string) (_ func() (bool, []mount.Mount, error), target string, err error) {
 
-	handler := func() (bool, []mount.Mount, error) {
-		// Prepare a space for containerd to make snapshot from container image.
+	var handler func() (bool, []mount.Mount, error)
+
+	// Handler to prepare a directory for containerd to download and unpacking layer.
+	defaultHandler := func() (bool, []mount.Mount, error) {
 		mounts, err := sn.mounts(ctx, labels, s)
 		return false, mounts, err
 	}
 
-	target, remote := labels[label.TargetSnapshotRef]
-
+	// Handler to stop containerd from downloading and unpacking layer.
 	skipHandler := func() (bool, []mount.Mount, error) {
-		// The handler tells containerd do not download and unpack layer.
 		return true, nil, nil
 	}
+
+	remoteHandler := func(id string, labels map[string]string) func() (bool, []mount.Mount, error) {
+		return func() (bool, []mount.Mount, error) {
+			logger.Debugf("Found nydus meta layer id %s", id)
+			if err := sn.prepareRemoteSnapshot(id, labels); err != nil {
+				return false, nil, err
+			}
+
+			// Let Prepare operation show the rootfs content.
+			if err := sn.fs.WaitUntilReady(id); err != nil {
+				return false, nil, err
+			}
+
+			log.L.Infof("Nydus remote snapshot %s is ready", id)
+			mounts, err := sn.remoteMounts(ctx, s, id)
+			return false, mounts, err
+		}
+	}
+
+	target, remote := labels[label.TargetSnapshotRef]
 
 	if remote {
 		// Containerd won't consume mount slice for below snapshots
 		switch {
+		case label.IsNydusMetaLayer(labels):
+			logger.Debugf("found nydus meta layer")
+			handler = defaultHandler
 		case label.IsNydusDataLayer(labels):
 			logger.Debugf("found nydus data layer")
 			handler = skipHandler
-		case label.IsNydusMetaLayer(labels):
-			// Containerd has to download and unpack nydus meta layer for nydusd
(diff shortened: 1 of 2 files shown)


snapshot/snapshot.go
```diff
@@ -359,7 +351,7 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 		return nil, err
 	}
 
-	logger.Debugf("[Prepare] snapshot with labels %v", info.Labels)
+	logger.Infof("[Prepare] snapshot with labels %v", info.Labels)
```
changweige, line 354: The labels may contain sensitive information. So we should avoid print it as INFO level

snapshot/snapshot.go
```diff
@@ -378,22 +370,12 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
...
-	return snapshot.IterateParentSnapshots(ctx, o.ms, key, func(id string, i snapshots.Info) bool {
-		return label.IsNydusMetaLayer(i.Labels)
-	})
-}
-
 // The work on supporting View operation for nydus-snapshotter is divided into 2 parts:
 // 1. View on the topmost layer of nydus images or zran images
 // 2. View on the any layer of nydus images or zran images
 func (o *snapshotter) View(ctx context.Context, key, parent string, opts ...snapshots.Opt) ([]mount.Mount, error) {
+	log.L.Debugf("[View] snapshot with key %s, parent %s", key, parent)
```
changweige, line 377: Why duplicate one log line with L417?

snapshot/snapshot.go
```diff
@@ -509,7 +495,7 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
 	}
 
 	// For example: remove snapshot with key sha256:c33c40022c8f333e7f199cd094bd56758bc479ceabf1e490bb75497bf47c2ebf
-	log.L.Debugf("[Remove] snapshot with key %s snapshot id %s", key, id)
+	log.L.Infof("[Remove] snapshot with key %s snapshot id %s", key, id)
```
changweige, line 498: Please don't print it as INFO, it is tons of log lines

snapshot/snapshot.go
```diff
@@ -914,6 +889,11 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context) ([]string, erro
 		if _, ok := ids[d]; ok {
 			continue
 		}
+		// Skip temp snapshot directories
+		if strings.HasPrefix(d, "new-") {
```
changweige, line 893: Is there a reason why we should keep the temp directories?

jiangliu: snapshotter::createSnapshot() create a temporary directory named as `new-xxx` under the cache directory, which may be deleted by the gc thread, so we should protect it.

snapshot/snapshot.go
jiangliu: One is debug, another one is info. And print debug message at the entry for trouble shooting.

snapshot/snapshot.go
jiangliu: Fixed

snapshot/snapshot.go
changweige: What gc thread? Do you mean containerd GC routines?

snapshot/snapshot.go
jiangliu: Fixed

snapshot/process.go
```diff
@@ -64,35 +86,19 @@ func chooseProcessor(ctx context.Context, logger *logrus.Entry,
...
-
 		// Hope to find bootstrap layer and prepares to start nydusd
 		// TODO: Trying find nydus meta layer will slow down setting up rootfs to OCI images
 		if id, info, err := sn.findMetaLayer(ctx, key); err == nil {
 			logger.Infof("Prepares active snapshot %s, nydusd should start afterwards", key)
 			handler = remoteHandler(id, info.Labels)
 		}
 
-		if sn.fs.ReferrerDetectEnabled() {
+		if handler == nil && sn.fs.ReferrerDetectEnabled() {
```
imeoer, line 101: Is the `handler == nil` check right? It seems to have `handler = skipHandler` for the referrer.

snapshot/snapshot.go
```diff
@@ -343,14 +338,11 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
...
-func (o *snapshotter) prepareRemoteSnapshot(id string, labels map[string]string) error {
-	return o.fs.Mount(id, labels)
-}
-
 func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...snapshots.Opt) ([]mount.Mount, error) {
-	// `timer` can't be nil
-	timer := collector.NewSnapshotMetricsTimer(collector.SnapshotMethodPrepare)
-	defer timer.ObserveDuration()
+	log.L.Debugf("[Prepare] snapshot with key %s, parent %s", key, parent)
+	if timer := collector.NewSnapshotMetricsTimer(collector.SnapshotMethodRemove); timer != nil {
```
imeoer, line 343: Why using `SnapshotMethodRemove` instead of `collector.SnapshotMethodPrepare` here?

snapshot/snapshot.go
jiangliu: removed

snapshot/process.go
jiangliu: The `handler` variable is initialized to nil and it hasn't been assigned to any value for referrer case. So seems the check is right.

snapshot/snapshot.go
jiangliu: My mistake, fixed.

snapshot/snapshot.go
jiangliu: Aha, you are right. The should keep recent temp files and remove old temp files. So change it to:
```
		// Temporary files may still be in use, so only delete temp files elder than one day.
		if strings.HasPrefix(d, "new-") {
			if fi, err := os.Stat(o.snapshotDir(d)); err == nil {
				if fi.ModTime().Add(24 * time.Hour).After(time.Now()) {
					continue
				}
			}
		}
```

imeoer: It seems unnecessary to handle this race. containerd snapshot gc and prepare have rwlock protection.

https://github.com/containerd/containerd/blob/1f54e8fb21ff565bc240a1a9ca29e4566e4ce617/metadata/snapshot.go#L300

https://github.com/containerd/containerd/blob/1f54e8fb21ff565bc240a1a9ca29e4566e4ce617/metadata/snapshot.go#LL831C12-L831C12

imeoer:
others LGTM.

snapshot/snapshot.go
jiangliu: Thanks, I will remove this patch.

imeoer (approved):

PR #479, opened 2023-05-15
Support Nydus boostrap layer encryption and decryption

taoohong:
Use ocicrypt to support encryption and decryption of nydus boostrap layer.


go.mod
```diff
@@ -46,6 +46,14 @@ require (
 	k8s.io/cri-api v0.27.0-alpha.3
 )
 
+require (
```
imeoer, line 49: Can we merge this into below?

pkg/converter/convert_unix.go
```diff
@@ -932,7 +934,7 @@ func convertManifest(ctx context.Context, cs content.Store, oldDesc ocispec.Desc
 		opt.OCI = true
 	}
 
-	// Append bootstrap layer to manifest.
+	// Append bootstrap layer to manifest, encryp bootstrap layer if needed.
 	bootstrapDesc, blobDescs, err := MergeLayers(ctx, cs, manifest.Layers, opt)
```
imeoer, line 938: Typo `encryp`.

pkg/converter/convert_unix.go
```diff
@@ -1138,5 +1140,12 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 		},
 	}
 
+	if len(opt.EncryptRecipients) != 0 {
+		// Encrypt the Nydus bootstrap layer.
+		bootstrapDesc, err = EncryptNyudsBootstrap(ctx, cs, bootstrapDesc, opt.EncryptRecipients)
```
imeoer, line 1145: Typo `Nyuds`.

pkg/converter/convert_unix.go
```diff
@@ -919,7 +919,9 @@ func convertManifest(ctx context.Context, cs content.Store, oldDesc ocispec.Desc
 		return nil, errors.Wrap(err, "read manifest json")
 	}
 
-	if isNydusImage(&manifest) {
+	// If opt.EncryptRecipients is empty, the Nydus image does not need to be
+	// encrypted after conversion.
+	if len(opt.EncryptRecipients) == 0 && isNydusImage(&manifest) {
```
imeoer, line 924: Does this work if the user needs encryption on the first conversion but not on the second?

pkg/converter/encryption.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := cw.Commit(ctx, st.Offset, ""); err != nil {
+		if !errdefs.IsAlreadyExists(err) {
+			return "", 0, fmt.Errorf("failed commit on ref %q: %w", ref, err)
+		}
+	}
+
+	return cw.Digest(), st.Offset, nil
+}
+
+// Encrypt Nydus boostrap layer
```
imeoer, line 142: Typo `bootstrap`.

pkg/converter/encryption.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+	defer dataReader.Close()
+
+	cc, err := enchelpers.CreateCryptoConfig(encryptRecipients, []string{})
+	if err != nil {
+		return ocispec.Descriptor{}, fmt.Errorf("create crypt config failed: %w", err)
+	}
+	newDesc, resultReader, encLayerFinalizer, err = encryptLayer(&cc, dataReader, desc)
```
imeoer, line 160: The `err` needs to be handled.

pkg/converter/encryption.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+		for k, v := range annotations {
+			newDesc.Annotations[k] = v
+		}
+	}
+	return newDesc, err
+}
+
+// Decrypt the Nydus boostrap layer
+func DeryptNydusBootstrap(ctx context.Context, cs content.Store, desc ocispec.Descriptor, decryptKeys []string, unwrapOnly bool) (ocispec.Descriptor, error) {
```
imeoer, line 201: Maybe explain the `unwrapOnly` option.

pkg/converter/convert_unix.go
taoohong: Every time `convert` is called seperatly, the corresponding opt will be generated according to the configuration. If there are multiple calls to `convert` in the life cycle of one opt, they will be handled the same. Do we have this scenario?

adamqqqplay:
@taoohong Hi, could you please add some unit tests to your code? Especially the newly added file `pkg/converter/encryption.go`, thanks in advance.

pkg/converter/convert_unix.go
taoohong: If users want to do conversion with different encryption behaviors, they have to modify the configration file or the cmd args.

imeoer: > Every time `convert` is called seperatly, the corresponding opt will be generated according to the configuration. If there are multiple calls to `convert` in the life cycle of one opt, they will be handled the same. Do we have this scenario?

Multiple calls to convert are possible, such as using acceld/nerdctl (imported the Converter package) to convert an image from containerd's content store.

taoohong: > 



> > Every time `convert` is called seperatly, the corresponding opt will be generated according to the configuration. If there are multiple calls to `convert` in the life cycle of one opt, they will be handled the same. Do we have this scenario?
> 
> Multiple calls to convert are possible, such as using acceld/nerdctl (imported the Converter package) to convert an image from containerd's content store.

User may need to modify the conguration file or the cmd arg according to his needs.

pkg/converter/encryption.go
taoohong: done

changweige:
Could you please share some background information on this feature and its potential users or use cases to help us understand its necessity? How could we consume the package?

go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/containerd/cgroups v1.1.0 // indirect
 	github.com/containerd/ttrpc v1.2.1 // indirect
 	github.com/containerd/typeurl/v2 v2.1.0 // indirect
+	github.com/containers/ocicrypt v1.1.7
```
changweige, line 73: We don't usually claim an intended module in this require clause. Could you relocate it to the first require clause?

taoohong:
> Could you please share some background information on this feature and its potential users or use cases to help us understand its necessity? How could we consume the package?

Data security has become an important topic in the field of cloud computing. When users no longer trust CSPs, the confidentiality and security of user image must be guaranteed. A typical application scenario is confidential containers, where Nydus supports image encryption to address the need for users to accelerate encrypted images. This patch is only one aspect of Nydus' support for image encryption, as it only encrypts metadata. There are other subsequent patches that support chunk-level encryption of data blobs, ultimately achieving Nydus' support for encrypted images.

go.mod
taoohong: done

adamqqqplay:
> > Could you please share some background information on this feature and its potential users or use cases to help us understand its necessity? How could we consume the package?
> 
> Data security has become an important topic in the field of cloud computing. When users no longer trust CSPs, the confidentiality and integrity of user image must be guaranteed. A typical application scenario is confidential containers, where Nydus supports image encryption to address the need for users to accelerate encrypted images. This patch is only one aspect of Nydus' support for image encryption, as it only encrypts metadata. There are other subsequent patches that support chunk-level encryption of data blobs, ultimately achieving Nydus' support for encrypted images. How to use this? For example: `nydusify convert --source busybox --target localhost:5000/busybox:nydus-enc --encrypt-recipients jwe:/etc/pubkey.pem` `nydusify check --target localhost:5000/busybox:nydus-enc --decrypt-keys /etc/key.pem` It's amost the same in acceld/nerdctl, users can add encrypt & decrypt options to configuration file to do this.

Maybe we could add some docs later to introduce this new feature.

imeoer:
How do we support the blob layer encryption and decryption later?

taoohong:
> How do we support the blob layer encryption and decryption later?

JiaNang Huang is working on this. XD

imeoer (approved):
LGTM

PR #482, opened 2023-05-16
converter: do not append batch size if non is provided

bergwolf:
It actually creates an incompatibility between nydusify and nydus-image. Let's keep the default (0) working at least.

### DIFF
--- a/pkg/converter/tool/builder.go
+++ b/pkg/converter/tool/builder.go
@@ -133,7 +133,7 @@ func buildPackArgs(option PackOption) []string {
 	if option.ChunkSize != "" {
 		args = append(args, "--chunk-size", option.ChunkSize)
 	}
-	if option.BatchSize != "" {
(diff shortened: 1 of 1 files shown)


adamqqqplay:
> Maybe we need to change the codecov policy to verify overall coverage, not patch coverage. cc @adamqqqplay

OK, I will fix this.

imeoer:
@bergwolf Please rebase upstream/main, to skip the codecov test.

imeoer:
A better solution is using `pkg/converter/tool/feature.go` feature detection. cc @hangvane

hangvane:
> It actually creates an incompatibility between nydusify and nydus-image. Let's keep the default (0) working at least.

> A better solution is using `pkg/converter/tool/feature.go` feature detection. cc @hangvane

@bergwolf @imeoer Thanks for your suggestion! I will implement the feature detection to address the incompatibility issue with batch size.

PR #483, opened 2023-05-22
converter: ignore nonexistent blob data for unpack

imeoer:
When we convert nydus with empty blob data to gzip
(or other compression types) layer, will encounter the error:

```
unpack nydus blob: unpack nydus tar: unpack blob from nydus:
can't find target image.blob by seeking tar: data not found
```

The nydus layer may contain only bootstrap and no blob
data, which should be ignored.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -140,6 +140,11 @@ func unpackNydusBlob(bootDst, blobDst string, ra content.ReaderAt, unpackBlob bo
 		defer blob.Close()
 
 		if _, err = UnpackEntry(ra, EntryBlob, blob); err != nil {
+			if errors.Is(err, ErrNotFound) {
+				// The nydus layer may contain only bootstrap and no blob
+				// data, which should be ignored.
+				return nil
+			}
 			return errors.Wrap(err, "unpack blob from nydus")
 		}
 	}


changweige:
Does this PR relate to https://github.com/moby/buildkit/pull/3814#issuecomment-1554245344 ?

pkg/converter/convert_unix.go
```diff
@@ -140,6 +140,11 @@ func unpackNydusBlob(bootDst, blobDst string, ra content.ReaderAt, unpackBlob bo
 		defer blob.Close()
 
 		if _, err = UnpackEntry(ra, EntryBlob, blob); err != nil {
+			if errors.Is(err, ErrNotFound) {
+				// The nydus layer may contain only bootstrap and no blob
```
changweige, line 144: Looks like we have to delete `blob` if noi unpack is needed

imeoer: No need for this, the converter will clean up all files in workdir.

imeoer:
> Does this PR relate to [moby/buildkit#3814 (comment)](https://github.com/moby/buildkit/pull/3814#issuecomment-1554245344) ?

Yes, it's for the broken CI.

changweige (approved):

PR #487, opened 2023-05-26
Minor improvements for image optimizer

changweige:
1. Add its own makefile and invokes the makefile from the project root makefile
2. No need to use libc integer types, the standard integer types meet our reuqiurements


sctb512 (approved):
LGTM

PR #491, opened 2023-06-05
workflow: fix incorrect files number for optimizer test 

sctb512:
Sometimes, the number of files accessed is not equal to the expected values. In this test, we only need to judge whether the optimizer works. This is sufficient if the number is not equal to 0.

Related to https://github.com/containerd/nydus-snapshotter/pull/446


changweige (approved):
LGTM, Thanks

PR #493, opened 2023-06-15
Transmit the prefetch list to snapshotter during runp pod.yaml

billie60:
Hello, in this PR, I start an NRI plugin and subscribe to pod creation event. The plugin invokes the RunPodSandbox method to obtain  prefetch list file path recorded in pod.yaml when monitoring the pod creation command. Then send the prefetch list file path to snapshotter through the specified socket API. After receiving prefetch list file path, snapshotter will save the content and add --prefetch-files parameter when building nydusd command to customize prefetch files content. I will submit the modifications to snapshotter in the next PR.

PR #496, opened 2023-06-18
Prefetchfiles snapshotter

billie60:
Hello, in order to complete the transmitting and persistence of the prefetch-files param in the nydusd command(persistence means restoring the correct nydusd command when accidentally restarting the snapshotter) , the following modifications are made to the nydus-snapshotter. Those modifications can achieve prefetch-files transmitting when creating a single container, but it is not flexible. For example, when executing commands in the following order:
1. crictl runp busybox-sandbox.yaml
2. crictl runp wordpress-sandbox.yaml
3. crictl create <pod_id> busybox.yaml busybox-sandbox.yaml
At this point, the busybox container is created, but the prefetch-files is the path of wordpress. To solve this problem,  I attempted to define PrefetchList in the States structure as a dictionary and add each prefetch-files parameter to the dictionary. Then, monitor the create container event though NRI plugin and obtain the image name of the comtainer being created, and find the corresponding image name in PrefetchList dictionary to build nydusd cmd. However, this method did not achieve the effect I wanted, because the corresponding image name will transmit from NRI only after the nydusd command is completed, which results in the failure to correctly add the prefetch-files parameter. So currently, I haven't come up with a better way to optimize the code. What should I do next? Please give me some suggestions. Thanks.

### DIFF
--- a/cmd/prefetchfiles-nri-plugin/main.go
+++ b/cmd/prefetchfiles-nri-plugin/main.go
@@ -1,0 +1,216 @@
+/*
+* Copyright (c) 2023. Nydus Developers. All rights reserved.
+*
+* SPDX-License-Identifier: Apache-2.0
+ */
+
+package main
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"strings"
+	"log/syslog"
+	"os"
+
+	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


sctb512:
It seems that the file changes in https://github.com/containerd/nydus-snapshotter/pull/493 also existed in this PR. Please remove irrelevant code and make sure each PR is independent.

PR #498, opened 2023-06-20
feats: support encrypting nydus data blobs

taoohong:
support encrypting nydus data blobs

Related [issue](https://github.com/dragonflyoss/image-service/issues/1334)

### DIFF
--- a/pkg/converter/constant.go
+++ b/pkg/converter/constant.go
@@ -19,6 +19,7 @@ const (
 	LayerAnnotationNydusBlobSize      = "containerd.io/snapshot/nydus-blob-size"
 	LayerAnnotationNydusBootstrap     = "containerd.io/snapshot/nydus-bootstrap"
 	LayerAnnotationNydusSourceChainID = "containerd.io/snapshot/nydus-source-chainid"
+	LayerAnnotationNydusEncryptedBlob = "containerd.io/snapshot/nydus-encrypted-blob"
 
 	LayerAnnotationNydusReferenceBlobIDs = "containerd.io/snapshot/nydus-reference-blob-ids"
 
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -326,6 +326,9 @@ func Pack(ctx context.Context, dest io.Writer, opt PackOption) (io.WriteCloser, 
 	if opt.BatchSize != "" && opt.BatchSize != "0" {
 		requiredFeatures.Add(tool.FeatureBatchSize)
 	}
+	if opt.Encrypt {
+		requiredFeatures.Add(tool.FeatureEncrypt)
+	}
 
 	detectedFeatures, err := tool.DetectFeatures(builderPath, requiredFeatures, tool.GetHelp)
 	if err != nil {
@@ -407,6 +410,7 @@ func packFromDirectory(ctx context.Context, dest io.Writer, opt PackOption, buil
 				ChunkSize:        opt.ChunkSize,
 				Compressor:       opt.Compressor,
 				Timeout:          opt.Timeout,
+				Encrypt:          opt.Encrypt,
 
 				Features: opt.features,
 			})
@@ -510,6 +514,7 @@ func packFromTar(ctx context.Context, dest io.Writer, opt PackOption) (io.WriteC
 				BatchSize:        opt.BatchSize,
 				Compressor:       opt.Compressor,
 				Timeout:          opt.Timeout,
+				Encrypt:          opt.Encrypt,
 
 				Features: opt.features,
 			})
@@ -852,6 +857,10 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
 			newDesc.Annotations[label.NydusRefLayer] = desc.Digest.String()
 		}
 
+		if opt.Encrypt {
+			newDesc.Annotations[LayerAnnotationNydusEncryptedBlob] = "true"
+		}
+
 		if opt.Backend != nil {
 			if err := opt.Backend.Push(ctx, cs, newDesc); err != nil {
 				return nil, errors.Wrap(err, "push to storage backend")
@@ -1134,6 +1143,11 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 		if opt.OCIRef {
 			blobDesc.Annotations[label.NydusRefLayer] = layers[idx].OriginalDigest.String()
 		}
+
+		if len(opt.EncryptRecipients) != 0 {
+			blobDesc.Annotations[LayerAnnotationNydusEncryptedBlob] = "true"
+		}
+
 		blobDescs = append(blobDescs, blobDesc)
 	}
 
--- a/pkg/converter/tool/feature.go
+++ b/pkg/converter/tool/feature.go
@@ -32,6 +32,9 @@ const (
 	// into a big batch chunk, which can reduce the the size of the image
 	// and accelerate the runtime file loading.
 	FeatureBatchSize Feature = "--batch-size"
+	// The option `--encrypt` enables converting directories, tar files
+	// or OCI images into encrypted nydus blob.
+	FeatureEncrypt Feature = "--encrypt"
 )
 
 var requiredFeatures Features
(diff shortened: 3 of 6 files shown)


sctb512 (commented):
LGTM

pkg/converter/convert_unix.go
```diff
@@ -852,12 +857,6 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
 			newDesc.Annotations[label.NydusRefLayer] = desc.Digest.String()
 		}
 
-		if opt.Backend != nil {
```
jiangliu, line 855: Why removing this code block?

taoohong: > Why removing this code block? It's used to push generated blob to backend storage, and it should be irrelevented to this PR?

Yes, I accidentally deleted this code block and it should be saved. Sorry for this.

pkg/converter/constant.go
```diff
@@ -22,5 +22,6 @@ const (
 
 	LayerAnnotationNydusReferenceBlobIDs = "containerd.io/snapshot/nydus-reference-blob-ids"
 
-	LayerAnnotationUncompressed = "containerd.io/uncompressed"
+	LayerAnnotationUncompressed       = "containerd.io/uncompressed"
+	LayerAnnotationNydusEncryptedBlob = "containerd.io/snapshot/encrypted-blob"
```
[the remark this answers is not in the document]
taoohong: > Is this annotation defined by containerd or ocicrypt? Should we add "nydus" prefix to "encrypted-blob"?

Neither, this is only defined and utilized in nydus context now. Ocicrypt use annotations to save cipher info but we dont need it for encrypted data blob, so this is only for impling this blob is encrypted. Maybe I should define this just  after `LayerAnnotationNydusSourceChainID`

pkg/converter/constant.go
```diff
@@ -19,6 +19,7 @@ const (
 	LayerAnnotationNydusBlobSize      = "containerd.io/snapshot/nydus-blob-size"
 	LayerAnnotationNydusBootstrap     = "containerd.io/snapshot/nydus-bootstrap"
 	LayerAnnotationNydusSourceChainID = "containerd.io/snapshot/nydus-source-chainid"
+	LayerAnnotationNydusEncryptedBlob = "containerd.io/snapshot/encrypted-blob"
```
jiangliu, line 22: How about renaming it as "ontainerd.io/snapshot/nydus-encrypted-blob"?

pkg/converter/convert_unix.go
```diff
@@ -843,8 +848,9 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
 			Annotations: map[string]string{
 				// Use `containerd.io/uncompressed` to generate DiffID of
 				// layer defined in OCI spec.
-				LayerAnnotationUncompressed: blobDigest.String(),
-				LayerAnnotationNydusBlob:    "true",
+				LayerAnnotationUncompressed:       blobDigest.String(),
+				LayerAnnotationNydusBlob:          "true",
+				LayerAnnotationNydusEncryptedBlob: strconv.FormatBool(len(opt.EncryptRecipients) != 0),
```
changweige, line 853: Is it possible we only add the encrypt annotation to nydus image only if its value is true?

pkg/converter/convert_unix.go
```diff
@@ -1127,8 +1133,9 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 			Size:      blobInfo.Size,
 			MediaType: MediaTypeNydusBlob,
 			Annotations: map[string]string{
-				LayerAnnotationUncompressed: blobDigest.String(),
-				LayerAnnotationNydusBlob:    "true",
+				LayerAnnotationUncompressed:       blobDigest.String(),
+				LayerAnnotationNydusBlob:          "true",
+				LayerAnnotationNydusEncryptedBlob: strconv.FormatBool(len(opt.EncryptRecipients) != 0),
```
changweige, line 1138: Ditto

pkg/converter/types.go
```diff
@@ -79,6 +79,8 @@ type PackOption struct {
 	Backend Backend
 	// Timeout cancels execution once exceed the specified time.
 	Timeout *time.Duration
+	// EncryptRecipients holds keys to converting OCI tar(.gz) blob to nydus encrypted blob.
```
changweige, line 82: keys for ?

pkg/converter/convert_unix.go
taoohong: very good suggestion, thanks!

pkg/converter/types.go
taoohong: These key are the same with the keys which used to encrypt Nydus bootstrap with ocicrypt.
I decided to change this item to 'Encrypt' which is a bool value, because the  useage of these specific keys in the future is not very likely.

taoohong:
I've add a test for encrypt since last commit in order to revole the code coverage problem. But it seems that the CI is not based on the latest nydus-image, causing the smoke unit test to fail. 😢

changweige:
> I've add a test for encrypt since last commit in order to revole the code coverage problem. But it seems that the CI is not based on the latest nydus-image, causing the smoke unit test to fail. 😢 cc @changweige



> I've add a test for encrypt since last commit in order to revole the code coverage problem. But it seems that the CI is not based on the latest nydus-image, causing the smoke unit test to fail. 😢 cc @changweige

Is there any plan for the next nydusd/nydus-image release with an image encryption feature?

taoohong:
> > I've add a test for encrypt since last commit in order to revole the code coverage problem. But it seems that the CI is not based on the latest nydus-image, causing the smoke unit test to fail. 😢 cc @changweige
> 
> > I've add a test for encrypt since last commit in order to revole the code coverage problem. But it seems that the CI is not based on the latest nydus-image, causing the smoke unit test to fail. 😢 cc @changweige
> 
> Is there any plan for the next nydusd/nydus-image release with an image encryption feature?

cc @jiangliu

changweige (commented):
Overall looks good to me. But for the sake of compatibility, we have to check if the underlying nydus toolchain meets our feature requirements in the function `DetectFeatures`.  In this PR, the required feature would be `--encrypt`.

changweige:
If there is no concrete plan for the next nydus-image release, we can move the CI test to another PR and just check if the underlying nydus-image has the feature `--encrypt`?

taoohong:
> just check if the underlying nydus-image has the feature `--encrypt`?
yes

> If there is no concrete plan for the next nydus-image release, we can move the CI test to another PR and just check if the underlying nydus-image has the feature `--encrypt`?

codecov/patch would fail due to 0.00% of diff hit if I don't add the CI test. PR can still be merged even if this test failed?

changweige:
> > just check if the underlying nydus-image has the feature `--encrypt`?
> 
> yes
> 
> > If there is no concrete plan for the next nydus-image release, we can move the CI test to another PR and just check if the underlying nydus-image has the feature `--encrypt`?
> 
> codecov/patch would fail due to 0.00% of diff hit if I don't add the CI test. PR can still be merged even if this test failed?

In fact the result from `codecov/patch` is not a strong restriction, I can still press merge if it fails.

pkg/converter/tool/feature.go
```diff
@@ -32,6 +32,9 @@ const (
 	// into a big batch chunk, which can reduce the the size of the image
 	// and accelerate the runtime file loading.
 	FeatureBatchSize Feature = "--batch-size"
+	// The option `--encrypt` enables converting directories, tar files
+	// or OCI images into encrypted nydus blob.
+	FeatureEncrypt Feature = "--encrypt"
```
changweige, line 37: Should we detect if the required feature `encrypt` is supported by nydus-image? Seems we only define a constant here.

```
	requiredFeatures := tool.NewFeatures(tool.FeatureTar2Rafs)
	if opt.BatchSize != "" && opt.BatchSize != "0" {
		requiredFeatures.Add(tool.FeatureBatchSize)
	}

	detectedFeatures, err := tool.DetectFeatures(builderPath, required features, tool.GetHelp)
	if err != nil {
		return nil, err
```

taoohong: done

changweige (approved):
LGMT

changweige:
Thanks

Issue #474, opened 2023-05-10
Failed to pull images when using nydus-snapshotter with tarfs


ChengyuZhu6:
Failed to pull images with OCIv1 manifest when using nydus-snapshotter with tarfs.

test images: busybox, ubuntu, centos, pause
```
failed:     docker.io/ubuntu:latest,   k8s.gcr.io/pause:3.6
successful: docker.io/centos:latest  docker.io/busybox:latest
```
 errors:
```
# pause:
ERRO[2023-05-10T02:38:37.581320528Z] prepare Tarfs layer of snapshot ID 1, 
err: unexpected HEAD status code https://k8s.gcr.io/v2/pause/blobs/sha256:c2280d2f5f56cf9c9a01bb64b2db4651e35efd6d62a54dcfc12049fe6449c5e4: 404 Not Found  
key="default/1/extract-228310730-iOxr sha256:1021ef88c7974bfff89c5a0ec4fd3160daac6c48a075f74cff721f85dd104e68" parent=
# ubuntu:
ERRO[2023-05-10T02:40:24.734003392Z] prepare Tarfs layer of snapshot ID 2, 
err: unexpected HEAD status code https://registry-1.docker.io/v2/library/ubuntu/blobs/sha256:ca5534a51dd04bbcebe9b23ba05f389466cf0c190f1f8f182d7eea92a9671d00: 404 Not Found  
key="k8s.io/1/extract-322997560-jiyx sha256:b8a36d10656ac19ddb96ef3107f76820663717708fc37ce929925c36d1b1d157" parent=
```

The possible reason for the failure is due to the difference in the media type and we may need to support manifest v1 . There are the image information:
```
$ docker manifest inspect ubuntu:latest
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json"

$ docker manifest inspect centos:latest
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json"
```

imeoer:
cc @kevinXYin

PR #500, opened 2023-06-26
remote: enhance FetchByDigest() to handle more mediatypes

jiangliu:
Enhance FetchByDigest() to handle manifest with mediatype "application/vnd.oci.image.manifest.v1+json"

Fixes: https://github.com/containerd/nydus-snapshotter/issues/474

### DIFF
--- a/pkg/remote/remotes/docker/fetcher.go
+++ b/pkg/remote/remotes/docker/fetcher.go
@@ -151,10 +151,15 @@ func (r dockerFetcher) Fetch(ctx context.Context, desc ocispec.Descriptor) (io.R
 	})
 }
 
-func (r dockerFetcher) createGetReq(ctx context.Context, host RegistryHost, ps ...string) (*request, int64, error) {
+func (r dockerFetcher) createGetReq(ctx context.Context, host RegistryHost, mediatype string, ps ...string) (*request, int64, error) {
 	headReq := r.request(host, http.MethodHead, ps...)
 	if err := headReq.addNamespace(r.refspec.Hostname()); err != nil {
 		return nil, 0, err
+	}
+	if mediatype == "" {
+		headReq.header.Set("Accept", "*/*")
+	} else {
+		headReq.header.Set("Accept", strings.Join([]string{mediatype, `*/*`}, ", "))
 	}
 
 	headResp, err := headReq.doWithRetries(ctx, nil)
@@ -190,13 +195,14 @@ func (r dockerFetcher) FetchByDigest(ctx context.Context, dgst digest.Digest) (i
 	}
 
 	var (
-		getReq   *request
-		sz       int64
-		firstErr error
+		getReq    *request
+		sz        int64
+		firstErr  error
+		mediaType string
 	)
 
 	for _, host := range r.hosts {
-		getReq, sz, err = r.createGetReq(ctx, host, "blobs", dgst.String())
+		getReq, sz, err = r.createGetReq(ctx, host, mediaType, "blobs", dgst.String())
 		if err == nil {
 			break
 		}
@@ -208,8 +214,15 @@ func (r dockerFetcher) FetchByDigest(ctx context.Context, dgst digest.Digest) (i
 
 	if getReq == nil {
 		// Fall back to the "manifests" endpoint
+		// TODO: this change should be upstreamed to containerd.
+		mediaType = strings.Join([]string{
+			images.MediaTypeDockerSchema2Manifest,
+			images.MediaTypeDockerSchema2ManifestList,
+			ocispec.MediaTypeImageManifest,
+			ocispec.MediaTypeImageIndex,
+		}, ", ")
 		for _, host := range r.hosts {
-			getReq, sz, err = r.createGetReq(ctx, host, "manifests", dgst.String())
+			getReq, sz, err = r.createGetReq(ctx, host, mediaType, "manifests", dgst.String())
 			if err == nil {
 				break
 			}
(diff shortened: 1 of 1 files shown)


liubogithub:
Looks good to me.

pkg/remote/remotes/docker/fetcher.go
```diff
@@ -209,7 +215,22 @@ func (r dockerFetcher) FetchByDigest(ctx context.Context, dgst digest.Digest) (i
...
+				firstErr = err
+			}
+		}
+	}
+
+	if getReq == nil {
+		// Fall back to the "manifests" endpoint with specific mediatype
+		mediatype = "application/vnd.oci.image.manifest.v1+json"
+		for _, host := range r.hosts {
+			getReq, sz, err = r.createGetReq(ctx, host, mediatype, "manifests", dgst.String())
```
imeoer, line 233: Can we make a slice like `mediaTypes := []string{ images.MediaTypeDockerSchema2Manifest, images.MediaTypeDockerSchema2ManifestList, images.MediaTypeDockerSchema1Manifest, ocispec.MediaTypeImageManifest, ocispec.MediaTypeImageIndex }` ? And then add them into the `Accept` request header.

pkg/remote/remotes/docker/fetcher.go
```diff
@@ -151,11 +151,16 @@ func (r dockerFetcher) Fetch(ctx context.Context, desc ocispec.Descriptor) (io.R
 	})
 }
 
-func (r dockerFetcher) createGetReq(ctx context.Context, host RegistryHost, ps ...string) (*request, int64, error) {
+func (r dockerFetcher) createGetReq(ctx context.Context, host RegistryHost, mediatype string, ps ...string) (*request, int64, error) {
```
changweige, line 154: The whole remote package is copied from containerd. Why containerd does not have this problem?

pkg/remote/remotes/docker/fetcher.go
jiangliu: Fixed

pkg/remote/remotes/docker/fetcher.go
jiangliu: It does have this issue, only no client to trigger it:)

cardyok:
Maybe we should simply change the argument for FetchByDigest, use descriptor instead of digest and use the mediatype defined by client? Since original purpose of FetchByDigest is to use registry as a general purpose CAS, it makes more sense to avoid defining image related headers.

pkg/remote/remotes/docker/fetcher.go
```diff
@@ -208,8 +214,14 @@ func (r dockerFetcher) FetchByDigest(ctx context.Context, dgst digest.Digest) (i
 
 	if getReq == nil {
 		// Fall back to the "manifests" endpoint
+		mediatype = strings.Join([]string{
```
imeoer, line 217: Better rename to `mediaTypes`.

pkg/remote/remotes/docker/fetcher.go
```diff
@@ -208,8 +214,14 @@ func (r dockerFetcher) FetchByDigest(ctx context.Context, dgst digest.Digest) (i
...
+			images.MediaTypeDockerSchema2Manifest,
+			images.MediaTypeDockerSchema2ManifestList,
+			ocispec.MediaTypeImageManifest,
+			ocispec.MediaTypeImageIndex,
+		}, ",")
```
imeoer, line 222: `", "`

pkg/remote/remotes/docker/fetcher.go
jiangliu: fixed

pkg/remote/remotes/docker/fetcher.go
jiangliu: fixed

jiangliu:
> Maybe we should simply change the argument for FetchByDigest, use descriptor instead of digest and use the mediatype defined by client? Since original purpose of FetchByDigest is to use registry as a general purpose CAS, it makes more sense to avoid defining image related headers.

We only have the manifest digest and not known the mediaType, so need to probe the mediaType.

pkg/remote/remotes/docker/fetcher.go
imeoer, line 154: `mediaTypes string`

pkg/remote/remotes/docker/fetcher.go
```diff
@@ -190,13 +195,14 @@ func (r dockerFetcher) FetchByDigest(ctx context.Context, dgst digest.Digest) (i
...
-		firstErr error
+		getReq    *request
+		sz        int64
+		firstErr  error
+		mediaType string
 	)
 
 	for _, host := range r.hosts {
-		getReq, sz, err = r.createGetReq(ctx, host, "blobs", dgst.String())
+		getReq, sz, err = r.createGetReq(ctx, host, mediaType, "blobs", dgst.String())
```
imeoer, line 205: `mediaTypes`

imeoer (commented):
others LGTM, in addition, it is better to comment that we have modified this codes, and wait for upstream to accept it.

pkg/remote/remotes/docker/fetcher.go
jiangliu: let's keep consistence with open(), which uses `mediatype`?

imeoer (approved):

pkg/remote/remotes/docker/fetcher.go
imeoer: okay

imeoer (approved):

PR #501, opened 2023-06-27
dump logs of containerd if optimizer test fails

changweige:
The optimzer test occasionally fails, let's see what bad happens.


jiangliu (approved):

sctb512 (approved):
LGTM

PR #502, opened 2023-06-27
snapshot: extend the ExtraOption struct

jiangliu:
Move ExtraOption to dedicated file and extend it for more usage cases.


jiangliu:
Outdated by #530

Issue #388, opened 2023-02-28
rethink nydusd configuration file


imeoer:
1. Nydusd configuration file includes registry auth information, it shouldn't be saved to disk.
2. When the nydusd configuration file changes, the new nydusd should use the updated.
3. Once the auth subscribed from k8s secret changes, nydusd also needs to update, in case of using an expired auth.

imeoer:
cc @ccx1024cc

changweige:
> When the nydusd configuration file changes, the new nydusd should use the updated. 

nydus-snapshotter is providing the new nydusd config file to nydusd now

DarkMountain-wyz:
> 1. Nydusd configuration file includes registry auth information, it shouldn't be saved to disk (how to handle live upgrade and failover?).
> 2. When the nydusd configuration file changes, the new nydusd should use the updated.
> 3. Once the auth subscribed from k8s secret/docker config changes, nydusd also needs to update, in case of using an expired auth.

I am Guijie Wang, and I will complete this part of the work.

changweige:
> When the nydusd configuration file changes, the new nydusd should use the updated.

The already running nydusd should not change its configuration I suppose. The newly created nydusd can use the new nydusd configuration. Otherwise, it's not easy to handle the consistency between nydusd config file and nydusd DB records

Can we only add an interface or credential service to nydus-snapshotter?  The credential service can provide auth to nydusd when it requests

changweige:
> > The already running nydusd should not change its configuration I suppose. The newly created nydusd can use the new nydusd configuration. Otherwise, it's not easy to handle the consistency between nydusd config file and nydusd DB records
> 
> Yes, we said `"the new nydusd should use the updated."`.
> 
> > Can we only add an interface or credential service to nydus-snapshotter? The credential service can provide auth to nydusd when it requests
> 
> The registry credential is the one of wanting to be updated for nydusd, the new nydusd maybe also need the new prefetch configuration, etc. I think we'd better not let nydusd become aware of the credential, it should be handled by the containerd credential plugin with snapshotter in the future, and then nydusd only get the full configuration from snapshotter.

The registry auth ever passed to nydusd might be expired due to the registry configuration. So I suppose we need a way to refresh the auth nydusd is using.

changweige:
[...]

PR #504, opened 2023-07-09
Adding an interface to the system socket to help nydusd obtain config

DarkMountain-wyz:
# Issue
[https://github.com/containerd/nydus-snapshotter/issues/388](url)

# Details
- Add interface endpointGetDaemonConfig string="/api/v1/daemons/{id}/config" in system. go
- Add handler func getDaemonConfig() to deliver config
- Pass the system socket and URL to nydusd through command-line arguments

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -111,6 +111,7 @@ type Experimental struct {
 	EnableStargz         bool        `toml:"enable_stargz"`
 	EnableReferrerDetect bool        `toml:"enable_referrer_detect"`
 	TarfsConfig          TarfsConfig `toml:"tarfs"`
+	EnableBackendSource  bool        `toml:"enable_backend_source"`
 }
 
 type TarfsConfig struct {
--- a/config/daemonconfig/daemonconfig.go
+++ b/config/daemonconfig/daemonconfig.go
@@ -10,6 +10,8 @@ package daemonconfig
 import (
 	"encoding/json"
 	"os"
+	"reflect"
+	"strings"
 
 	"github.com/pkg/errors"
 
@@ -75,16 +77,16 @@ type BackendConfig struct {
 	// Registry backend configs
 	Host               string         `json:"host,omitempty"`
 	Repo               string         `json:"repo,omitempty"`
-	Auth               string         `json:"auth,omitempty"`
-	RegistryToken      string         `json:"registry_token,omitempty"`
+	Auth               string         `json:"auth,omitempty" secret:"true"`
+	RegistryToken      string         `json:"registry_token,omitempty" secret:"true"`
 	BlobURLScheme      string         `json:"blob_url_scheme,omitempty"`
 	BlobRedirectedHost string         `json:"blob_redirected_host,omitempty"`
 	Mirrors            []MirrorConfig `json:"mirrors,omitempty"`
 
 	// OSS backend configs
 	EndPoint        string `json:"endpoint,omitempty"`
-	AccessKeyID     string `json:"access_key_id,omitempty"`
-	AccessKeySecret string `json:"access_key_secret,omitempty"`
+	AccessKeyID     string `json:"access_key_id,omitempty" secret:"true"`
+	AccessKeySecret string `json:"access_key_secret,omitempty" secret:"true"`
 	BucketName      string `json:"bucket_name,omitempty"`
 	ObjectPrefix    string `json:"object_prefix,omitempty"`
 
@@ -124,6 +126,9 @@ type DeviceConfig struct {
 // We don't have to persist configuration file for fscache since its configuration
 // is passed through HTTP API.
 func DumpConfigFile(c interface{}, path string) error {
+	if config.IsBackendSourceEnabled() {
+		c = serializeWithSecretFilter(c)
+	}
 	b, err := json.Marshal(c)
 	if err != nil {
 		return errors.Wrapf(err, "marshal config")
@@ -178,4 +183,53 @@ func SupplementDaemonConfig(c DaemonConfig, imageID, snapshotID string,
 	}
 
 	return nil
+}
+
+func serializeWithSecretFilter(obj interface{}) map[string]interface{} {
+	result := make(map[string]interface{})
+	value := reflect.ValueOf(obj)
+	typeOfObj := reflect.TypeOf(obj)
+
+	if value.Kind() == reflect.Ptr {
+		value = value.Elem()
+		typeOfObj = typeOfObj.Elem()
+	}
+
+	for i := 0; i < value.NumField(); i++ {
+		field := value.Field(i)
+		fieldType := typeOfObj.Field(i)
+		secretTag := fieldType.Tag.Get("secret")
+		jsonTags := strings.Split(fieldType.Tag.Get("json"), ",")
+		omitemptyTag := false
+
+		for _, tag := range jsonTags {
+			if tag == "omitempty" {
+				omitemptyTag = true
+				break
+			}
+		}
+
+		if secretTag == "true" {
+			continue
+		}
+
+		if field.Kind() == reflect.Ptr && field.IsNil() {
+			continue
+		}
+
+		if omitemptyTag && reflect.DeepEqual(reflect.Zero(field.Type()).Interface(), field.Interface()) {
+			continue
+		}
+
+		switch fieldType.Type.Kind() {
+		case reflect.Struct:
+			result[jsonTags[0]] = serializeWithSecretFilter(field.Interface())
+		case reflect.Ptr:
+			result[jsonTags[0]] = serializeWithSecretFilter(field.Elem().Interface())
+		default:
+			result[jsonTags[0]] = field.Interface()
+		}
+	}
+
+	return result
 }
--- a/config/daemonconfig/daemonconfig.go
+++ b/config/daemonconfig/daemonconfig.go
@@ -221,11 +221,14 @@ func serializeWithSecretFilter(obj interface{}) map[string]interface{} {
 			continue
 		}
 
-		if fieldType.Type.Kind() == reflect.Struct {
+		switch fieldType.Type.Kind() {
+		case reflect.Struct:
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Interface())
-		} else if fieldType.Type.Kind() == reflect.Ptr {
+		case reflect.Ptr:
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Elem().Interface())
-		} else {
+		case reflect.Invalid: 
+			
+		default:
 			result[jsonTags[0]] = field.Interface()
 		}
 	}
--- a/config/daemonconfig/daemonconfig.go
+++ b/config/daemonconfig/daemonconfig.go
@@ -221,7 +221,7 @@ func serializeWithSecretFilter(obj interface{}) map[string]interface{} {
 			continue
 		}
 
-		//nolint:staticcheck
+		//nolint:exhaustive
 		switch fieldType.Type.Kind() {
 		case reflect.Struct:
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Interface())
--- a/config/daemonconfig/daemonconfig.go
+++ b/config/daemonconfig/daemonconfig.go
@@ -221,61 +221,12 @@ func serializeWithSecretFilter(obj interface{}) map[string]interface{} {
 			continue
 		}
 
+		//nolint:staticcheck
 		switch fieldType.Type.Kind() {
 		case reflect.Struct:
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Interface())
 		case reflect.Ptr:
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Elem().Interface())
-		case reflect.Bool:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Int:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Int8:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Int16:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Int32:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Int64:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Uint:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Uint8:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Uint16:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Uint32:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Uint64:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Uintptr:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Float32:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Float64:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Complex64:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Complex128:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Array:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Chan:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Func:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Interface:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Map:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Slice:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.String:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.UnsafePointer:
-			result[jsonTags[0]] = field.Interface()
-		case reflect.Invalid:
-			
 		default:
 			result[jsonTags[0]] = field.Interface()
 		}
--- a/config/daemonconfig/daemonconfig.go
+++ b/config/daemonconfig/daemonconfig.go
@@ -221,12 +221,11 @@ func serializeWithSecretFilter(obj interface{}) map[string]interface{} {
 			continue
 		}
 
-		switch fieldType.Type.Kind() {
-		case reflect.Struct:
+		if fieldType.Type.Kind() == reflect.Struct {
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Interface())
-		case reflect.Ptr:
+		} else if fieldType.Type.Kind() == reflect.Ptr {
 			result[jsonTags[0]] = serializeWithSecretFilter(field.Elem().Interface())
-		default:
+		} else {
 			result[jsonTags[0]] = field.Interface()
 		}
 	}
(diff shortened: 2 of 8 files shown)


pkg/daemon/command/command.go
```diff
@@ -122,12 +124,23 @@ func WithThreadNum(num int) Opt {
 	}
 }
 
-func WithConfig(config string) Opt {
+func WithConfig(id string) Opt {
```
imeoer, line 127: Better not to change the method, so that we can keep compatibility for old nydusd versions.

pkg/manager/daemon_adaptor.go
```diff
@@ -146,7 +146,8 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
 			}
 
 			cmdOpts = append(cmdOpts,
-				command.WithConfig(d.ConfigFile("")),
+				command.WithConfig(d.States.ID),
```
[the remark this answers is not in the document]
DarkMountain-wyz: I will consider compatibility issues with the old version of nydusd and make incremental updates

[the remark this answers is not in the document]
DarkMountain-wyz: > Before users upgrade nydusd, we still need to be compatible with old nydusd versions.
> 
> How about using the experimental configuration option such as `enable_config_source` to control the feature? Just like the experimental option `enable_referrer_detect`.

I will consider compatibility issues with the old version of nydusd and make incremental updates

pkg/daemon/command/command.go
```diff
@@ -122,12 +124,23 @@ func WithThreadNum(num int) Opt {
...
 	return func(cmd *DaemonCommand) {
+		config := strings.Replace("/api/v1/daemons/{id}/config", "{id}", id, 1)
 		cmd.Config = config
 	}
 }
 
+func WithConfigSource(source string) Opt {
+	return func(cmd *DaemonCommand) {
+		if source == "" {
+			cmd.ConfigSource = "/run/containerd-nydus/system.sock"
```
sctb512, line 137: It would be better to define "/run/containerd-nydus/system.sock" as a constant.

sctb512:
> https://github.com/containerd/nydus-snapshotter/issues/388

This hyperlink for URL is not valid.

pkg/daemon/command/command.go
```diff
@@ -122,12 +124,23 @@ func WithThreadNum(num int) Opt {
 	}
 }
 
-func WithConfig(config string) Opt {
+func WithConfig(id string) Opt {
 	return func(cmd *DaemonCommand) {
+		config := strings.Replace("/api/v1/daemons/{id}/config", "{id}", id, 1)
```
sctb512, line 129: Should we use `strings.Replace()` here?
`fmt.Sprintf("/api/v1/daemons/%s/config", id)` also works.

pkg/manager/daemon_adaptor.go
```diff
@@ -146,7 +146,8 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
...
+				command.WithConfigSource(""),
```
sctb512, line 150: Expose `ConfigSource` as a configuration entry? Otherwise, use `/run/containerd-nydus/system.sock` as an argument for this method.

sctb512:
> * Pass the system socket and URL to nydusd through command-line arguments

I am confused why we need this API? Could you tell me more about this?

sctb512:
Related to https://github.com/dragonflyoss/image-service/pull/1361

DarkMountain-wyz:
> > * Pass the system socket and URL to nydusd through command-line arguments
> 
> I am confused why we need this API? Could you tell me more about this?

My initial plan was to provide an API in nydusd to pass config from SN to nydusd, but this would require opening the socket before creating the daemon. The thread creating the daemon would need to block until the API server thread received the passed config. Similarly, in SN, it is necessary to block after sending the config until the daemon is created. In the original logic, once the SN establishes a connection with the socket, it is considered that the daemon has been created. I think such changes will have a significant impact, bringing many uncontrollable impacts. Therefore, the current solution is to request config from nydusd to SN. This will not change the original Logical framework, but also add control items that are compatible with the old version of nydusd.

pkg/manager/daemon_adaptor.go
DarkMountain-wyz: > Expose `ConfigSource` as a configuration entry? Otherwise, use `/run/containerd-nydus/system.sock` as an argument for this method.

I am also considering this point. This value is a property of the Controller class in system. go. To obtain this socket path, you should first obtain the Controller instance. I have no idea how to obtain this instance yet.

DarkMountain-wyz:
> > #388
> 
> The link for this issue is not valid.

The url is https://github.com/containerd/nydus-snapshotter/issues/388

pkg/daemon/command/command.go
DarkMountain-wyz: Okay, I'll take this better approach

pkg/daemon/command/command.go
DarkMountain-wyz: I need to find a method to obtain the Controller instance to obtain this value.

changweige:
Is it possible only add the capability to nydus-snapshotter to share registry credentials to nydusd when nydusd requires it?
Other part of items should be persisted on disk like what we do now.

imeoer (approved):

imeoer:
I think in addition to considering that credentials cannot be persisted, we also need to consider how to promptly renew the registry token using the registry user/pass when it expires (for example when a credential helper is used).

And how nydusd can use the updated config when serving new images after the user updates the nydusd config. 

Therefore, I think it is feasible to have the snapshotter provide this API.

config/daemonconfig/daemonconfig.go
```diff
@@ -125,6 +125,9 @@ type DeviceConfig struct {
 // We don't have to persist configuration file for fscache since its configuration
 // is passed through HTTP API.
 func DumpConfigFile(c interface{}, path string) error {
+	if config.IsConfigSourceEnabled() {
+		return nil
```
sctb512, line 129: If we don't store this configuration, how can we resolve the situation where the nydus-snapshotter is restarted?

changweige:
Sorry, need rebase

config/config.go
```diff
@@ -109,6 +109,7 @@ const (
 type Experimental struct {
 	EnableStargz         bool `toml:"enable_stargz"`
 	EnableReferrerDetect bool `toml:"enable_referrer_detect"`
+	EnableConfigSource   bool `toml:"enable_config_source"`
```
changweige, line 112: Can we have a doc or comments telling the configuration item is used for?

changweige: In addition, it will be great if we can add an example configuration item at `misc/snapshotter/config.toml` with a usage comment

config/daemonconfig/daemonconfig.go
changweige: Seems with the config `source_enabled`, nydusd does not need config file anymore?

pkg/daemon/command/command.go
```diff
@@ -26,14 +26,15 @@ type DaemonCommand struct {
...
-	Config     string `type:"param" name:"config"`
-	Bootstrap  string `type:"param" name:"bootstrap"`
-	Mountpoint string `type:"param" name:"mountpoint"`
-	APISock    string `type:"param" name:"apisock"`
-	LogLevel   string `type:"param" name:"log-level"`
-	Supervisor string `type:"param" name:"supervisor"`
-	LogFile    string `type:"param" name:"log-file"`
+	ID           string `type:"param" name:"id"`
+	Config       string `type:"param" name:"config"`
+	ConfigSource string `type:"param" name:"config-source"`
```
changweige, line 31: Does nydusd already have the CLI argument `config-source` ?

pkg/manager/daemon_adaptor.go
```diff
@@ -145,10 +148,20 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
...
 			}
 
-			cmdOpts = append(cmdOpts,
-				command.WithConfig(d.ConfigFile("")),
-				command.WithBootstrap(bootstrap),
-			)
+			if config.IsConfigSourceEnabled() {
+				configApiPath := fmt.Sprintf(endpointGetDaemonConfig, d.States.ID)
+				cmdOpts = append(cmdOpts,
+					command.WithConfig(configApiPath),
```
changweige, line 154: Is the `configApiPath` a local config file?

pkg/system/system.go
```diff
@@ -167,6 +169,38 @@ func (sc *Controller) registerRouter() {
...
+			defer ma.Unlock()
+
+			d := ma.GetByDaemonID(id)
+			if d != nil {
+				jsonResponse(w, d.Config)
+				return
+			}
+		}
+
+		err = errors.New("daemon not found")
```
changweige, line 201: Please try to use predefined errors in errdefs package

pkg/system/system.go
```diff
@@ -167,6 +169,38 @@ func (sc *Controller) registerRouter() {
...
+				http.Error(w, m.encode(), statusCode)
+			}
+		}()
+
+		vars := mux.Vars(r)
+		id := vars["id"]
+
+		for _, ma := range sc.managers {
+			ma.Lock()
+			defer ma.Unlock()
```
changweige, line 192: It's not good to hole all the locks all the time

changweige (changes requested):
Thanks, but we may need more work to do

config/config.go
DarkMountain-wyz: I will add an example

config/daemonconfig/daemonconfig.go
DarkMountain-wyz: After discussing with my supervisor, I will change my plan. The auth configuration will not be stored locally, but will be passed through an API in nydus-snapshotter.

pkg/manager/daemon_adaptor.go
DarkMountain-wyz: configApiPath is a URI path

pkg/daemon/command/command.go
DarkMountain-wyz: I made modifications to nydusd and submitted PR to add this CLI parameter

config/config.go
DarkMountain-wyz: > Can we have a doc or comments telling the configuration item is used for?

This option is for compatibility with older versions of nydusd, which require a config drop disk

DarkMountain-wyz:
The method has changed, and now snapshotter supports providing backend config to nydus instead of writing files。

config/config.go
```diff
@@ -107,8 +107,9 @@ const (
 )
 
 type Experimental struct {
-	EnableStargz         bool `toml:"enable_stargz"`
-	EnableReferrerDetect bool `toml:"enable_referrer_detect"`
+	EnableStargz           bool `toml:"enable_stargz"`
+	EnableReferrerDetect   bool `toml:"enable_referrer_detect"`
+	EnableCredentialSource bool `toml:"enable_credential_source"`
```
imeoer, line 112: `EnableBackendSource` ?

DarkMountain-wyz: yes, and I will modify it

misc/snapshotter/config.toml
```diff
@@ -104,3 +104,6 @@ enable_stargz = false
 # The option enables trying to fetch the Nydus image associated with the OCI image and run it.
 # Also see https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers
 enable_referrer_detect = false
+# Whether to enable authentication support
+# The option enables nydus snapshot to provide backend config to nydusd instead of writing to a file.
+enable_backend_source = true
```
imeoer, line 109: Should add a blank line for the file tail.

config/daemonconfig/fuse.go
```diff
@@ -94,9 +95,29 @@ func (c *FuseDaemonConfig) DumpString() (string, error) {
...
 		return err
 	}
+	if config.IsBackendSourceEnabled() {
+		configJSON, err := json.Marshal(c)
+		if err != nil {
+			return err
+		}
+
+		var configMap map[string]interface{}
+		err = json.Unmarshal(configJSON, &configMap)
```
imeoer, line 114: The process of Marshal and Unmarshal is a bit strange, in fact we only need to set some sensitive fields to an empty string (such as `BackendConfig.Auth`, `BackendConfig.RegistryToken`, `BackendConfig.AccessKeyID`, etc.). Or we can do nothing and let users not fill in sensitive fields in config.json.

In addition, we should not completely clear the `BackendConfig`, otherwise we will not be able to restore nydusd after the snapshotter restarts, please make sure that your PR still works in this case.

DarkMountain-wyz: @imeoer  I have now modified my plan and only these sensitive fields (such as BackendConfig.Auth, BackendConfig.RegistryToken, BackendConfig.AccessKeyID, etc.). will not be written to the local file.

DarkMountain-wyz:
Hello @sctb512 @changweige , I've submitted a Pull Request that introduces a new implementation approach. In this approach, the backend configuration's Auth information is not persisted. Instead, during the startup of nydusd, a URL is provided, and nydusd retrieves the Auth information by making a request to this URL. This design aims to enhance the security. Please review and share your thoughts on this approach. Thank you!

imeoer:
@DarkMountain-wyz It seems we need to fix the CI.

DarkMountain-wyz:
> @DarkMountain-wyz It seems we need to fix the CI.

I will fix it.

changweige:
need rebase

config/daemonconfig/fuse.go
```diff
@@ -94,9 +95,36 @@ func (c *FuseDaemonConfig) DumpString() (string, error) {
...
+		case BackendTypeRegistry:
+			backendAuth := c.Device.Backend.Config.Auth
+			registryToken := c.Device.Backend.Config.RegistryToken
+			c.Device.Backend.Config.Auth = ""
+			c.Device.Backend.Config.RegistryToken = ""
+			err := DumpConfigFile(c, f)
+			c.Device.Backend.Config.Auth = backendAuth
+			c.Device.Backend.Config.RegistryToken = registryToken
+			return err
+		case BackendTypeOss:
```
imeoer, line 118: The `s3` backend should also be handled. :)

imeoer: Perhaps a more elegant way would be to use struct tags and reflect, for example:

```
type BackendConfig struct {
	...
	Auth               string         `json:"auth,omitempty" secret:"true"`
	RegistryToken      string         `json:"registry_token,omitempty" secret:"true"`

	AccessKeyID     string `json:"access_key_id,omitempty" secret:"true"`
	AccessKeySecret string `json:"access_key_secret,omitempty" secret:"true"`
  ...
}
```

pkg/system/system.go
```diff
@@ -171,6 +173,63 @@ func (sc *Controller) registerRouter() {
...
+					}{
+						backendConfig.AccessKeyID,
+						backendConfig.AccessKeySecret,
+					}
+				default:
+					err = errdefs.ErrNotFound
+					statusCode = http.StatusNotFound
+					return
+				}
+				jsonResponse(w, credentialConfig)
```
imeoer, line 223: Maybe respond to the `backendConfig` directly here? In this way, when we add a new backend, we don't need to worry about handling miss here.

config/daemonconfig/fuse.go
imeoer: > The `s3` backend should also be handled. :)

Sorry for this, the `s3` backend only be added in https://github.com/dragonflyoss/image-service/blob/631db29759f4dceacaa7ff80b845a9d0107beef5/api/src/config.rs#L520, not in snapshshotter, maybe we don't need to handle it first.

config/daemonconfig/daemonconfig.go
```diff
@@ -171,11 +175,60 @@ func SupplementDaemonConfig(c DaemonConfig, imageID, snapshotID string,
...
+	case backendTypeLocalfs:
+	case backendTypeOss:
 	default:
 		return errors.Errorf("unknown backend type %s", backendType)
 	}
 
 	return nil
 }
+
+func serializeWithSecretFilter(obj interface{}) map[string]interface{} {
```
imeoer, line 187: Can we add a UT for this func?

config/config.go
```diff
@@ -107,9 +107,10 @@ const (
 )
 
 type Experimental struct {
-	EnableStargz         bool        `toml:"enable_stargz"`
-	EnableReferrerDetect bool        `toml:"enable_referrer_detect"`
-	TarfsConfig          TarfsConfig `toml:"tarfs"`
+	EnableStargz           bool        `toml:"enable_stargz"`
+	EnableReferrerDetect   bool        `toml:"enable_referrer_detect"`
+	TarfsConfig            TarfsConfig `toml:"tarfs"`
+	EnableCredentialSource bool        `toml:"enable_credential_source"`
```
imeoer, line 113: EnableBackendSource bool `toml:"enable_backend_source"`

config/daemonconfig/daemonconfig.go
```diff
@@ -124,11 +126,13 @@ type DeviceConfig struct {
 // We don't have to persist configuration file for fscache since its configuration
 // is passed through HTTP API.
 func DumpConfigFile(c interface{}, path string) error {
+	if config.IsCredentialSourceEnabled() {
```
imeoer, line 129: `IsBackendSourceEnabled`

pkg/manager/daemon_adaptor.go
```diff
@@ -25,6 +26,8 @@ import (
 	"github.com/containerd/nydus-snapshotter/pkg/prefetch"
 )
 
+const endpointGetCredential string = "/api/v1/daemons/%s/credential"
```
imeoer, line 29: const endpointGetBackend string = "/api/v1/daemons/%s/backend"

pkg/system/system.go
```diff
@@ -171,6 +172,47 @@ func (sc *Controller) registerRouter() {
 	sc.router.HandleFunc(endpointDaemonsUpgrade, sc.upgradeDaemons()).Methods(http.MethodPut)
 	sc.router.HandleFunc(endpointDaemonRecords, sc.getDaemonRecords()).Methods(http.MethodGet)
 	sc.router.HandleFunc(endpointPrefetch, sc.setPrefetchConfiguration()).Methods(http.MethodPut)
+	sc.router.HandleFunc(endpointGetCredential, sc.getCredential()).Methods(http.MethodGet)
+}
+
+func (sc *Controller) getCredential() func(w http.ResponseWriter, r *http.Request) {
```
imeoer, line 178: `getBackend`

config/daemonconfig/daemonconfig_test.go
```diff
@@ -62,3 +62,63 @@ func TestLoadConfig(t *testing.T) {
...
+	require.Equal(t, newCfg.Device.Cache.Config, cfg.Device.Cache.Config)
+	require.Equal(t, newCfg.Mode, cfg.Mode)
+	require.Equal(t, newCfg.DigestValidate, cfg.DigestValidate)
+	require.Equal(t, newCfg.IOStatsFiles, cfg.IOStatsFiles)
+	require.Equal(t, newCfg.Device.Backend.Config.Host, cfg.Device.Backend.Config.Host)
+	require.Equal(t, newCfg.Device.Backend.Config.Repo, cfg.Device.Backend.Config.Repo)
+	require.Equal(t, newCfg.Device.Backend.Config.Proxy, cfg.Device.Backend.Config.Proxy)
+	require.Equal(t, newCfg.Device.Backend.Config.BlobURLScheme, cfg.Device.Backend.Config.BlobURLScheme)
+	require.NotEqual(t, newCfg.Device.Backend.Config.Auth, cfg.Device.Backend.Config.Auth)
+}
```
imeoer, line 124: Add a blank line here.

imeoer (approved):
LGTM!

changweige (approved):

PR #509, opened 2023-07-14
converter: allow controlling referrer with option

imeoer:
```
Some registry implementations like Docker Hub don't support the
OCI reference type (aka referrer) yet, we shouldn't enable it by
default, instead, provide a `WithReferrer` option to control it.
```

To fix https://github.com/dragonflyoss/image-service/issues/1363


pkg/converter/types.go
```diff
@@ -107,6 +107,10 @@ type MergeOption struct {
 	OCI bool
 	// OCIRef enables converting OCI tar(.gz) blob to nydus referenced blob.
 	OCIRef bool
+	// WithReferrer associates a reference to the original OCI manifest.
+	// See the `subject` field description in
+	// https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions
+	WithReferrer bool
```
changweige, line 113: Could help introduce why nydus image has to refer to the original OCI image?

imeoer: Updated the comment, PTAL.

changweige: Thanks for the update, looks reasonable.

changweige (approved):
LGTM

PR #511, opened 2023-07-18
action: remove the broken step of installing hub

imeoer:
https://github.com/containerd/nydus-snapshotter/actions/runs/5584824903

The hub has been [integrated](https://github.com/mislav/hub#github-actions) to github action worker, so let's remove it to fix the failure.

A successful action example can be found here: https://github.com/imeoer/nydus-snapshotter/actions/runs/5586816647


changweige (approved):
Thanks for fixing this

PR #512, opened 2023-07-21
Store auth in keyring and transfer it to nydusd via environment variable

sctb512:
1. Store `auth` for `nydusd` in the keyring instead of the configuration file.
2. Set `auth` via environment variable to avoid displaying it in command parameter.

Related to https://github.com/dragonflyoss/image-service/pull/1381 and https://github.com/dragonflyoss/image-service/pull/1382


pkg/manager/daemon_adaptor.go
```diff
@@ -192,6 +195,19 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
 
 	cmd := exec.Command(nydusdPath, args...)
 
+	if config.IsKeyringEnabled() && !config.IsFusedevSharedModeEnabled() {
```
changweige, line 198: Daemon has its own working mode which means nydusd daemons can work in the different modes in a single node.
Better to judge by daemon, we already have some helper functions doing this

pkg/filesystem/fs.go
```diff
@@ -298,7 +300,14 @@ func (fs *Filesystem) Mount(snapshotID string, labels map[string]string) (err er
 			daemonconfig.CacheDir:  cacheDir,
 		}
 		cfg := deepcopy.Copy(fsManager.DaemonConfig).(daemonconfig.DaemonConfig)
-		err = daemonconfig.SupplementDaemonConfig(cfg, imageID, snapshotID, false, labels, params)
+		var updateErr error
+		err = daemonconfig.SupplementDaemonConfig(cfg, imageID, snapshotID, false, labels, params, func(imageHost string, keyChain *auth.PassKeyChain) {
+			logrus.Infof("add key for %s", imageHost)
```
changweige, line 305: It's better to lower the log level to Debug

pkg/manager/daemon_adaptor.go
```diff
@@ -192,6 +195,19 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
...
+		c, err := daemonconfig.NewDaemonConfig(d.States.FsDriver, configPath)
```
changweige, line 199: I am afraid the current way to identify a container image reference's host is not agile.  For the dedicated mode nydusd daemon, it only as a rafs instance in its Instances where we can calculate what host is

pkg/keyring/keyring.go
```diff
@@ -0,0 +1,42 @@
+package keyring
```
changweige, line 1: Please add license claim

pkg/daemon/daemon.go
```diff
@@ -66,6 +71,7 @@ type Daemon struct {
 	// fusedev shared mode: zero, one or more RAFS instances
 	// fscache shared mode: zero, one or more RAFS instances
 	Instances rafsSet
+	authCache *lru.Cache
```
changweige, line 74: Auth should be shared among different nydusd daemons. So why to bind the cache to a single nydusd daemon?

pkg/daemon/daemon.go
```diff
@@ -151,6 +158,31 @@ func (d *Daemon) RemoveInstance(snapshotID string) {
 	d.DecRef()
 }
 
+func (d *Daemon) UpdateAuth(imageHost, auth string) error {
+	key, err := keyring.Add(imageHost, auth)
```
changweige, line 162: Is it necessary to consider the atomicity when updating?

pkg/daemon/daemon.go
```diff
@@ -151,6 +158,31 @@ func (d *Daemon) RemoveInstance(snapshotID string) {
...
+	return nil
+}
+
+func (d *Daemon) GetAuth(imageHost string) (string, error) {
+	if auth, ok := d.authCache.Get(imageHost); ok {
+		return auth.(string), nil
+	}
+
+	data, err := keyring.Search(imageHost)
+	if err == nil {
```
changweige, line 180: Checking if err is nil looks strange.

imeoer:
This approach seems to only solve the nydusd registry credential persistence problem, but when the credential is changed (e.g. registry user/pass dynamically generated by a credential helper), nydusd may not be able to timely detect the update. I think https://github.com/containerd/nydus-snapshotter/pull/504 is a better implementation.

pkg/manager/daemon_adaptor.go
sctb512: Fixed.

pkg/filesystem/fs.go
sctb512: Fixed.

pkg/manager/daemon_adaptor.go
sctb512: Fixed.

pkg/keyring/keyring.go
sctb512: Fixed.

pkg/daemon/daemon.go
sctb512: Fixed. Move it to the manager.

pkg/daemon/daemon.go
sctb512: Done.

pkg/daemon/daemon.go
sctb512: Fixed.

sctb512:
> This approach seems to only solve the nydusd registry credential persistence problem, but when the credential is changed (e.g. registry user/pass dynamically generated by a credential helper), nydusd may not be able to timely detect the update. I think #504 is a better implementation.

I believe this PR does not conflict with https://github.com/containerd/nydus-snapshotter/pull/504. It is better to get authentication from the credential server if `snapshotter` does not send it through `env`.

pkg/auth/cache.go
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright (c) 2023. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package auth
+
+import (
+	"github.com/pkg/errors"
+	"k8s.io/utils/lru"
```
imeoer, line 11: It seems we have used `github.com/golang/groupcache/lru`.

pkg/auth/keyring.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	defaultSessionName = "_ses.nydus"
+)
+
+type KeyRing struct {
+	lock *sync.RWMutex
+}
+
+func NewKeyRing() *KeyRing {
```
imeoer, line 26: Maybe give some unit test cases for this?

pkg/manager/daemon_adaptor.go
```diff
@@ -192,6 +199,25 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
 
 	cmd := exec.Command(nydusdPath, args...)
 
+	if config.IsKeyringEnabled() && !d.IsSharedDaemon() {
```
imeoer, line 202: Can we support shared daemon?

config/daemonconfig/daemonconfig.go
```diff
@@ -33,6 +33,7 @@ type DaemonConfig interface {
 	FillAuth(kc *auth.PassKeyChain)
 	StorageBackend() (StorageBackendType, *BackendConfig)
 	UpdateMirrors(mirrorsConfigDir, registryHost string) error
+	GetImageHost() string
```
changweige, line 36: Seems there is no caller for the added method `GetImageHost`. Please remove it if so

config/daemonconfig/daemonconfig.go
```diff
@@ -168,7 +169,11 @@ func SupplementDaemonConfig(c DaemonConfig, imageID, snapshotID string,
 		// when repository is public.
 		keyChain := auth.GetRegistryKeyChain(registryHost, imageID, labels)
 		c.Supplement(registryHost, image.Repo, snapshotID, params)
-		c.FillAuth(keyChain)
+		if config.IsKeyringEnabled() && fn != nil {
+			fn(registryHost, keyChain)
```
changweige, line 173: Can `fn` fail ?

pkg/daemon/daemon.go
```diff
@@ -232,13 +236,13 @@ func (d *Daemon) SharedMount(rafs *Rafs) error {
...
 	case config.FsDriverFusedev:
-		return d.sharedFusedevMount(rafs)
+		return d.sharedFusedevMount(rafs, authCache)
 	default:
 		return errors.Errorf("unsupported fs driver %s", d.States.FsDriver)
 	}
 }
 
-func (d *Daemon) sharedFusedevMount(rafs *Rafs) error {
+func (d *Daemon) sharedFusedevMount(rafs *Rafs, authCache *auth.Cache) error {
```
changweige, line 245: Erofs/fscache fs driver should also need the auth

pkg/daemon/daemon.go
```diff
@@ -255,6 +259,25 @@ func (d *Daemon) sharedFusedevMount(rafs *Rafs) error {
 			d.ConfigFile(rafs.SnapshotID))
 	}
 
+	if config.IsKeyringEnabled() {
+		image, err := registry.ParseImage(rafs.ImageID)
```
changweige, line 263: It's would be great if we can wrap this code snippet into a helper function which tries to fill the auth into config string

pkg/filesystem/fs.go
```diff
@@ -298,7 +300,14 @@ func (fs *Filesystem) Mount(snapshotID string, labels map[string]string) (err er
 			daemonconfig.CacheDir:  cacheDir,
 		}
 		cfg := deepcopy.Copy(fsManager.DaemonConfig).(daemonconfig.DaemonConfig)
-		err = daemonconfig.SupplementDaemonConfig(cfg, imageID, snapshotID, false, labels, params)
+		var updateErr error
+		err = daemonconfig.SupplementDaemonConfig(cfg, imageID, snapshotID, false, labels, params, func(imageHost string, keyChain *auth.PassKeyChain) {
+			logrus.Debugf("add key for %s", imageHost)
```
changweige, line 305: We don't usually directly use logrus in this project but using log.L

pkg/filesystem/fs.go
```diff
@@ -298,7 +300,14 @@ func (fs *Filesystem) Mount(snapshotID string, labels map[string]string) (err er
...
+			updateErr = fsManager.AuthCache.UpdateAuth(imageHost, keyChain.ToBase64())
+		})
+		if updateErr != nil {
```
changweige, line 308: Passing closure's error code through an intermediate variable is a little tricky. How about give the closure a return value which can be returned by `SupplementDaemonConfig`

pkg/auth/keyring.go
changweige: Agree, please try to add some unit test, especially for keying garbage collection and if the boundary is correctly handled.

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+	return &KeyRing{}
+}
+
+func (k *KeyRing) Add(id, value string) (int, error) {
+	sessKeyID, err := unix.KeyctlJoinSessionKeyring(defaultSessionName)
+	if err != nil {
+		return 0, errors.Wrap(err, "unable to session key")
+	}
+
+	if err := k.modifyKeyringPerm(sessKeyID, 0xffffffff, 0x80000); err != nil {
```
changweige, line 36: Please add a comment telling what those magic numbers for

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func NewKeyRing() *KeyRing {
+	return &KeyRing{}
+}
+
+func (k *KeyRing) Add(id, value string) (int, error) {
+	sessKeyID, err := unix.KeyctlJoinSessionKeyring(defaultSessionName)
+	if err != nil {
+		return 0, errors.Wrap(err, "unable to session key")
```
changweige, line 33: join the session key

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if err := k.modifyKeyringPerm(sessKeyID, 0xffffffff, 0x80000); err != nil {
+		return 0, errors.Wrap(err, "unable to mod keyring permissions")
+	}
+
+	k.lock.Lock()
+	defer k.lock.Unlock()
```
changweige, line 41: Why do we need a lock here? Is `unix.AddKey` not atomic?

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return err
+	}
+	perm := (uint32(perm64) & mask) | setbits
+
+	return unix.KeyctlSetperm(ringID, perm)
+}
+
+func (k *KeyRing) Search(id string) (string, error) {
+	sessKeyID, err := unix.KeyctlJoinSessionKeyring(defaultSessionName)
```
changweige, line 72: Is it necessary to join the keyring session again and again?

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "unable to session key")
+	}
+
+	if err := k.modifyKeyringPerm(sessKeyID, 0xffffffff, 0x80000); err != nil {
```
changweige, line 77: Don't use negative sentence int he wrapped error please

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+		return "", errors.Wrap(err, "unable to mod keyring permissions")
+	}
+
+	key, err := unix.KeyctlSearch(sessKeyID, "user", id, 0)
+	if err != nil {
+		return "", err
```
changweige, line 83: Please wrap the error

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+	sizeRead := size + 1
+
+	for sizeRead > size {
+		sizeRead, err = unix.KeyctlBuffer(unix.KEYCTL_READ, key, buffer, size)
+		if err != nil {
+			return "", err
+		}
+
+		if sizeRead > size {
+			buffer = make([]byte, sizeRead)
```
changweige, line 102: Um. How is it possible `sizeRead` is greater then `size`?

changweige: Please try to add a UT for it

pkg/auth/keyring.go
```diff
@@ -0,0 +1,110 @@
...
+	return keyID, nil
+}
+
+func (k *KeyRing) modifyKeyringPerm(ringID int, mask, setbits uint32) error {
+	dest, err := unix.KeyctlString(unix.KEYCTL_DESCRIBE, ringID)
+	if err != nil {
+		return err
+	}
+
+	res := strings.Split(dest, ";")
```
changweige, line 57: It's better to add a output example as comment to helper code readers to understand

pkg/auth/keyring.go
sctb512: We need this. `Addkey` only launch a `syscall`, but can't make sure it's atomic.
```shell
r0, _, e1 := Syscall6(SYS_ADD_KEY, uintptr(unsafe.Pointer(_p0)), uintptr(unsafe.Pointer(_p1)), uintptr(_p2), uintptr(len(payload)), uintptr(ringid), 0)
```

pkg/manager/daemon_adaptor.go
sctb512: Yes, it call `c.FillAuth(keyChain)` to fill auth before sending mount request to nydus daemon.

config/daemonconfig/daemonconfig.go
sctb512: Done.

config/daemonconfig/daemonconfig.go
sctb512: Yes, it depends on the caller `SupplementDaemonConfig` and the `fn` may fail.

pkg/daemon/daemon.go
sctb512: Fixed.

pkg/daemon/daemon.go
sctb512: Do you mean to wrap the code for parsing images and get the `keyChain` to a function?

sctb512: > Do you mean to wrap the code for parsing images and get the `keyChain` to a function?

Done.

pkg/filesystem/fs.go
sctb512: Fixed.

pkg/filesystem/fs.go
sctb512: Done.

pkg/auth/keyring.go
sctb512: Done.

pkg/auth/keyring.go
sctb512: Fixed.

pkg/auth/keyring.go
sctb512: Fixed.

pkg/auth/keyring.go
sctb512: DOne.

pkg/auth/keyring.go
sctb512: Done.

pkg/auth/keyring.go
sctb512: Done.

config/daemonconfig/daemonconfig.go
```diff
@@ -168,7 +169,16 @@ func SupplementDaemonConfig(c DaemonConfig, imageID, snapshotID string,
 		// when repository is public.
 		keyChain := auth.GetRegistryKeyChain(registryHost, imageID, labels)
 		c.Supplement(registryHost, image.Repo, snapshotID, params)
-		c.FillAuth(keyChain)
+		if config.IsKeyringEnabled() && fn != nil {
+			if err := fn(registryHost, keyChain); err != nil {
+				if errors.Is(err, unix.EINVAL) {
```
changweige, line 174: I think the error type should be transparent for the caller

changweige (commented):
Could we also have an e2e test case defined in `integration/entrypoint.sh`

PR #515, opened 2023-08-01
config: remove useless auth_through option

imeoer:
We have removed the `auth_through` option in https://github.com/dragonflyoss/image-service/pull/1375

Let's remove it in snapshotter, since its default value is false, there is no compatibility impact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,14 +157,11 @@ Dragonfly supports both **mirror** mode and HTTP **proxy** mode to boost the con
   "mirrors": [
     {
       "host": "http://127.0.0.1:65001",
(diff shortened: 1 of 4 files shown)


changweige:
Does it mean that all auth requests directly go to the original registry? The capability to get auth from the mirror registry is removed.

imeoer:
> Does it mean that all auth requests directly go to the original registry? The capability to get auth from the mirror registry is removed.

It's right, all auth requests will be sent to original registry auth server.

changweige:
Though this PR is not a breaking change, should we have a place to record the commits that ever changed UI like configuration file? So we can public them when composing release notes.

PR #517, opened 2023-08-08
feat: prevent unnecessary conversions

PerseidMeteor:
Add remote cache of remote registry.

Changes:

- Skip conversion if layer cache exists.
- Add two LayerAnnotation  as constant labels

### DIFF
--- a/pkg/converter/constant.go
+++ b/pkg/converter/constant.go
@@ -20,6 +20,8 @@ const (
 	LayerAnnotationNydusBootstrap     = "containerd.io/snapshot/nydus-bootstrap"
 	LayerAnnotationNydusSourceChainID = "containerd.io/snapshot/nydus-source-chainid"
 	LayerAnnotationNydusEncryptedBlob = "containerd.io/snapshot/nydus-encrypted-blob"
+	LayerAnnotationNydusSourceDigest  = "containerd.io/snapshot/nydus-source-digest"
+	LayerAnnotationNydusTargetDigest  = "containerd.io/snapshot/nydus-target-digest"
 
 	LayerAnnotationNydusReferenceBlobIDs = "containerd.io/snapshot/nydus-reference-blob-ids"
 
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -771,6 +771,28 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
 			return nil, nil
 		}
 
+		// Use remote cache to avoid unnecessary conversion
+		info, err := cs.Info(ctx, desc.Digest)
+		if err != nil {
+			return nil, errors.Wrapf(err, "get blob info %s", desc.Digest)
+		}
+		if info.Labels[LayerAnnotationNydusTargetDigest] != "" {
+			targetInfo, err := cs.Info(ctx, digest.Digest(info.Labels[LayerAnnotationNydusTargetDigest]))
+			if err != nil {
+				return nil, errors.Wrapf(err, "get blob info %s", desc.Digest)
+			}
+			targetDesc := ocispec.Descriptor{
+				Digest:    targetInfo.Digest,
+				Size:      targetInfo.Size,
+				MediaType: MediaTypeNydusBlob,
+				Annotations: map[string]string{
+					LayerAnnotationUncompressed: targetInfo.Digest.String(),
+					LayerAnnotationNydusBlob:    "true",
+				},
+			}
+			return &targetDesc, nil
+		}
+
 		ra, err := cs.ReaderAt(ctx, desc)
 		if err != nil {
 			return nil, errors.Wrap(err, "get source blob reader")
@@ -825,7 +847,7 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
 		}
 
 		blobDigest := digester.Digest()
-		info, err := cs.Info(ctx, blobDigest)
+		info, err = cs.Info(ctx, blobDigest)
 		if err != nil {
 			return nil, errors.Wrapf(err, "get blob info %s", blobDigest)
 		}


pkg/converter/convert_unix.go
```diff
@@ -589,6 +589,18 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
 				}
 				defer bootstrap.Close()
 
+				// Check if the bootstrap is already generated
+				bootCacheName := filepath.Join(opt.WorkDir, layers[idx].Digest.Hex())
```
imeoer, line 593: My suggestion is that external callers provide `layer.ReaderAt` implementations, rather than replacing bootstrap here.

pkg/converter/convert_unix.go
```diff
@@ -771,6 +783,31 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
...
+		info, err := cs.Info(ctx, desc.Digest)
+		if err != nil {
+			return nil, errors.Wrapf(err, "get blob info %s", desc.Digest)
+		}
+		if info.Labels[LayerAnnotationDestDigest] != "" {
+			destInfo, err := cs.Info(ctx, digest.Digest(info.Labels[LayerAnnotationDestDigest]))
+			if err != nil {
+				return nil, errors.Wrapf(err, "get blob info %s", desc.Digest)
+			} else {
+				destDesc := ocispec.Descriptor{
```
imeoer, line 796: Should resolve the lint error.

pkg/converter/convert_unix.go
```diff
@@ -771,6 +783,31 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
...
+					Digest:    destInfo.Digest,
+					Size:      destInfo.Size,
+					MediaType: MediaTypeNydusBlob,
+					Annotations: map[string]string{
+						// Here we do not add annotation containerd.io/snapshot/nydus-source-digest
```
imeoer, line 801: Why don't add the annotation directly here?

pkg/converter/convert_unix.go
```diff
@@ -771,6 +783,31 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
...
+						LayerAnnotationUncompressed:               destInfo.Digest.String(),
+						LayerAnnotationNydusBlob:                  "true",
+						LayerAnnotationFromCache: 				   "true",
```
imeoer, line 804: Ditto, does this code has been formatted?

pkg/converter/constant.go
```diff
@@ -20,6 +20,9 @@ const (
 	LayerAnnotationNydusBootstrap     = "containerd.io/snapshot/nydus-bootstrap"
 	LayerAnnotationNydusSourceChainID = "containerd.io/snapshot/nydus-source-chainid"
 	LayerAnnotationNydusEncryptedBlob = "containerd.io/snapshot/nydus-encrypted-blob"
+	LayerAnnotationSourceDigest       = "containerd.io/snapshot/nydus-source-digest"
+	LayerAnnotationDestDigest 		  = "containerd.io/snapshot/nydus-dest-digest"
```
imeoer, line 24: Does this code has been formatted?

imeoer, line 24: How about renaming to `LayerAnnotationTargetDigest = "containerd.io/snapshot/nydus-target-digest"`.

pkg/converter/convert_unix.go
```diff
@@ -836,6 +873,9 @@ func LayerConvertFunc(opt PackOption) converter.ConvertFunc {
 		// diff id calculation to speed up the conversion.
 		// See: https://github.com/containerd/containerd/blob/e4fefea5544d259177abb85b64e428702ac49c97/images/diffid.go#L49
 		info.Labels[labels.LabelUncompressed] = blobDigest.String()
+		// Update layer labels in remote cache.Actually these two labels will be write to store, too.
+		info.Labels[LayerAnnotationNydusBlob] = "true" 
+		info.Labels[LayerAnnotationSourceDigest] = desc.Digest.String()
```
imeoer, line 878: Why do we need to write these labels into content store?

pkg/converter/constant.go
PerseidMeteor: It seems weird, because it is formatted locally. May need to push again.
image

pkg/converter/convert_unix.go
PerseidMeteor: It is formatted locally, too.May need to push again.

pkg/converter/convert_unix.go
PerseidMeteor: Annotation `Actually these two labels will be write to store, too. ` is wrong, in fact, it is not written into the store, it is only written to the cache.

pkg/converter/constant.go
PerseidMeteor: fixed

PerseidMeteor: fixed after re-push

pkg/converter/convert_unix.go
PerseidMeteor: fixed.

pkg/converter/convert_unix.go
PerseidMeteor: fixed

imeoer:
Please also add commit message explaining why and how. :)

pkg/converter/constant.go
```diff
@@ -20,6 +20,8 @@ const (
 	LayerAnnotationNydusBootstrap     = "containerd.io/snapshot/nydus-bootstrap"
 	LayerAnnotationNydusSourceChainID = "containerd.io/snapshot/nydus-source-chainid"
 	LayerAnnotationNydusEncryptedBlob = "containerd.io/snapshot/nydus-encrypted-blob"
+	LayerAnnotationSourceDigest       = "containerd.io/snapshot/nydus-source-digest"
```
sctb512, line 23: `LayerAnnotationNydusSourceDigest`?

pkg/converter/constant.go
```diff
@@ -20,6 +20,8 @@ const (
...
+	LayerAnnotationTargetDigest       = "containerd.io/snapshot/nydus-target-digest"
```
sctb512, line 24: ditto.

pkg/converter/constant.go
PerseidMeteor: If the content store of the acceld provides remote cache, the cache layer in remote cache will have a `LayerAnnotationNydusSourceDigest` annotation to identify this is a converted layer, 
`LayerAnnotationNydusTargetDigest` will identify what is the converted digest, so we can skip this unnecessary conversion.

pkg/converter/constant.go
PerseidMeteor: ditto.

> ditto.

pkg/converter/constant.go
sctb512: Got it.
How about changing `LayerAnnotationSourceDigest` to `LayerAnnotationNydusTargetDigest`.

pkg/converter/convert_unix.go
PerseidMeteor: fixed

pkg/converter/convert_unix.go
PerseidMeteor: fixed

pkg/converter/constant.go
PerseidMeteor: fixed

imeoer (approved):

PR #520, opened 2023-08-14
feat: failover and hot upgrade

loheagn:


changweige:
Could we have some background information about this PR?

loheagn:
Sorry, this pr may not be ready until the nydusd support save/restore vfs state. I should convert this pr to draft by now.

pkg/manager/daemon_adaptor.go
```diff
@@ -32,8 +32,16 @@ import (
...
+//
+// The param `upgrade` decides whether to start the new daemon with `--upgrade` parameter.
+// The daemon with `--upgrade` will not mount the backend filesystem
+// and go to "Ready" state automatically,
+// but gives the snapshotter a chance to use `taskOver` command
+// to re-mount the backend filesystem whose metadata
+// is saved in the snapshotter memory by the previous daemon.
+// The "upgrade mode" is useful to do failover and "hot upgrade".
+func (m *Manager) StartDaemon(d *daemon.Daemon, upgrade bool) error {
+	cmd, err := m.BuildDaemonCommand(d, "", upgrade)
```
kevinXYin, line 44: we do not need this upgrade param during failover , nydusd will check whether the previous daemon crashed
https://github.com/dragonflyoss/image-service/blob/master/service/src/fusedev.rs#L520

bergwolf: It is needed if each nydusd runs inside a container, in which case newly created nydusd would not be able to see the old mountpoint of the old nydusd via `/proc/self/mounts`.

kevinXYin: Ok, Thanks

imeoer:
@loheagn Any updates for this PR?

ccx1024cc (approved):
Any update? +1

pkg/system/system.go
```diff
@@ -354,6 +408,7 @@ func (sc *Controller) upgradeNydusDaemon(d *daemon.Daemon, c upgradeRequest, man
 	}
 
 	if err := new.TakeOver(); err != nil {
+		// TODO(loheagn) kill the new
```
ccx1024cc, line 411: Implement please

ccx1024cc (changes requested):

loheagn:
@imeoer @ccx1024cc I'm not sure if this PR still needs to exist, as it seems that nydus-snapshotter can now work well with capabilities like failover of nydusd. Could you give me some advice?

imeoer:
@loheagn Thanks! Just follow the doc https://github.com/dragonflyoss/nydus/blob/master/docs/nydus-failover-upgrade.md, let's close the PR.

Issue #516, opened 2023-08-07
Support to pass the image name to the remote mount slice


ChengyuZhu6:

imeoer:
Please describe the details and use case.

ChengyuZhu6:
In the scenario of a confidential container, the image needs to be downloaded inside the vm, so an external method is needed to pass the name of the image to the vm. Therefore, the snapshotter needs to transfer the image name information through the mount slice.

imeoer:
> In the scenario of a confidential containers, the image needs to be downloaded inside the vm, so an external method is needed to pass the name of the image to the vm. Therefore, the snapshotter needs to transfer the image name information through the mount slice.

How can we let VM know the registry auth and mirror info?

ChengyuZhu6:
> > In the scenario of a confidential containers, the image needs to be downloaded inside the vm, so an external method is needed to pass the name of the image to the vm. Therefore, the snapshotter needs to transfer the image name information through the mount slice.
> 
> How can we let VM know the registry auth and mirror info?

The auth information will be passed in the vm through a plugin with a confidential channel.

jiangliu:
Fixed

PR #521, opened 2023-08-15
Support to pass the image name to the remote mount slice

ChengyuZhu6:
In the scenario of confidential containers, the image needs to be downloaded inside the vm. So an external method is needed to pass the name of the image to the vm. Therefore, we hope the snapshotter can pass the image name information through the mount slice.

Fixes: #516

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -109,6 +109,7 @@ const (
 type Experimental struct {
 	EnableStargz         bool `toml:"enable_stargz"`
 	EnableReferrerDetect bool `toml:"enable_referrer_detect"`
+	EnablePassImageURL   bool `toml:"enable_pass_image_url"`
 }
 
 type CgroupConfig struct {
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -350,7 +350,9 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 	if err != nil {
 		return nil, errors.Wrapf(err, "get snapshot %s", key)
 	}
-
+	if o.enableNydusOverlayFS && config.GetPassImageURLEnabled() {
+		return o.passImageNameMounts(ctx, info.Parent)
+	}
 	if needRemoteMounts {
 		return o.remoteMounts(ctx, *snap, metaSnapshotID)
 	}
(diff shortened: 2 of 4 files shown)


snapshot/snapshot.go
```diff
@@ -716,7 +718,25 @@ func overlayMount(options []string) []mount.Mount {
 		},
 	}
 }
+func (o *snapshotter) passImageNameMounts(ctx context.Context, parent string) ([]mount.Mount, error) {
+	_, pInfo, _, err := snapshot.GetSnapshotInfo(ctx, o.ms, parent)
+	if err != nil {
+		return nil, errors.Wrapf(err, "passImageNameMounts get snapshot %q info failed", parent)
+	}
+	passImageOption := fmt.Sprintf("imageurl=%s", pInfo.Labels["containerd.io/snapshot/cri.image-ref"])
```
changweige, line 726: I suppose it's possible that the key `containerd.io/snapshot/cri.image-ref` does not exist in snapshot's labels. We should handle such case

snapshot/snapshot.go
```diff
@@ -716,7 +718,25 @@ func overlayMount(options []string) []mount.Mount {
...
 
+	var overlayOptions []string
+	overlayOptions = append(overlayOptions, passImageOption)
+
+	return []mount.Mount{
+		{
+			Type:    "fuse.nydus-overlayfs",
```
changweige, line 733: The filesystem or mount type `fuse.nydus-overlayfs` should only appear when nydus-snapshotter has enable_nydus_overlayfs enabled in its toml config.
Moreover, we should follow the what we do in `remoteMountWithExtraOptions` like encoding the mount options.

```
	// base64 to filter easily in `nydus-overlayfs`
	opt := fmt.Sprintf("extraoption=%s", base64.StdEncoding.EncodeToString(no))
	overlayOptions = append(overlayOptions, opt)
```

ChengyuZhu6: Thanks, @changweige . I have fixed it.

snapshot/snapshot.go
ChengyuZhu6: Thanks for review. I have fixed it.

snapshot/snapshot.go
```diff
@@ -716,7 +718,31 @@ func overlayMount(options []string) []mount.Mount {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "passImageNameMounts get snapshot %q info failed", parent)
+	}
+	imageURL, ok := pInfo.Labels["containerd.io/snapshot/cri.image-ref"]
 
+	if !ok {
+		return nil, errors.Errorf("passImageNameMounts failed to find image ref of snapshot %s, labels %v",
+			parent, pInfo.Labels)
+	}
+	passImageOption := fmt.Sprintf("imageurl=%s", base64.StdEncoding.EncodeToString([]byte(imageURL)))
```
imeoer, line 732: How about follow https://github.com/containerd/containerd/blob/fb786b2f17b2c7d7ea5d9c9009d980bc03379952/pkg/snapshotters/annotations.go#L34, name it `image-ref` ?

snapshot/snapshot.go
```diff
@@ -350,7 +350,9 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 	if err != nil {
 		return nil, errors.Wrapf(err, "get snapshot %s", key)
 	}
-
+	if o.enableNydusOverlayFS && config.GetPassImageURLEnabled() {
+		return o.passImageNameMounts(ctx, info.Parent)
```
imeoer, line 354: Can we move it into `remoteMountWithExtraOptions`?

config/config.go
```diff
@@ -109,6 +109,7 @@ const (
 type Experimental struct {
 	EnableStargz         bool `toml:"enable_stargz"`
 	EnableReferrerDetect bool `toml:"enable_referrer_detect"`
+	EnablePassImageURL   bool `toml:"enable_pass_image_url"`
```
imeoer, line 112: Is it necessary to add the option?

PR #528, opened 2023-09-04
Make RAFS independent of daemon

jiangliu:
Originally RAFS instances are always associated with daemons, or daemon owns RAFS instance. With introduction of tarfs, rafs may be independent of daemon. So refactor rafs and daemon related code to make RAFS independent of daemon.


imeoer (approved):
LGTM

changweige (approved):
Looks pretty neat, thanks

PR #530, opened 2023-09-07
Support host sharing and guest pulling for Confidential Container image management

jiangliu:
Support host sharing and guest pulling for Confidential Container image management:
- import a `KataVirtualVolume` structure, which acts as a superset of `ExtraOptions`
- support build a disk image for a container image with EROFS and passing disk image info to kata-runtime through `KataVirtualVolume`.
- support downloading container image inside guest by introducing `proxy` mode, which doesn't download any layer content on host but only insert a `KataVirutalVolume` into `Mount.Options`.

This PR is based on #528

### DIFF
--- a/internal/constant/values.go
+++ b/internal/constant/values.go
@@ -25,6 +25,8 @@ const (
 	FsDriverFscache string = "fscache"
 	// Only prepare/supply meta/data blobs, do not mount RAFS filesystem.
 	FsDriverNodev string = "nodev"
+	// Relay layer content download operation to other agents.
+	FsDriverProxy string = "proxy"
 )
 
 const (
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -199,13 +199,12 @@ func (fs *Filesystem) TryStopSharedDaemon() {
 // WaitUntilReady wait until daemon ready by snapshotID, it will wait until nydus domain socket established
 // and the status of nydusd daemon must be ready
 func (fs *Filesystem) WaitUntilReady(snapshotID string) error {
-	// If NoneDaemon mode, there's no need to wait for daemon ready
-	if !fs.DaemonBacked() {
-		return nil
-	}
-
 	rafs := racache.RafsGlobalCache.Get(snapshotID)
 	if rafs == nil {
+		// If NoneDaemon mode, there's no need to wait for daemon ready
+		if config.GetDaemonMode() == config.DaemonModeNone {
+			return nil
+		}
 		return errors.Wrapf(errdefs.ErrNotFound, "no instance %s", snapshotID)
 	}
 
@@ -229,11 +228,6 @@ func (fs *Filesystem) WaitUntilReady(snapshotID string) error {
 // this method will fork nydus daemon and manage it in the internal store, and indexed by snapshotID
 // It must set up all necessary resources during Mount procedure and revoke any step if necessary.
 func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[string]string, s *storage.Snapshot) (err error) {
-	// Do not create RAFS instance in case of nodev.
-	if !fs.DaemonBacked() {
-		return nil
-	}
-
 	rafs := racache.RafsGlobalCache.Get(snapshotID)
 	if rafs != nil {
 		// Instance already exists, how could this happen? Can containerd handle this case?
@@ -309,7 +303,7 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
 			daemonconfig.WorkDir:   workDir,
 			daemonconfig.CacheDir:  cacheDir,
 		}
-		cfg := deepcopy.Copy(fsManager.DaemonConfig).(daemonconfig.DaemonConfig)
+		cfg := deepcopy.Copy(*fsManager.DaemonConfig).(daemonconfig.DaemonConfig)
 		err = daemonconfig.SupplementDaemonConfig(cfg, imageID, snapshotID, false, labels, params)
 		if err != nil {
 			return errors.Wrap(err, "supplement configuration")
@@ -362,6 +356,8 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
 		if err != nil {
 			err = errors.Wrapf(err, "mount tarfs for snapshot %s", snapshotID)
 		}
+	case config.FsDriverNodev:
+		// Nothing to do
 	default:
 		err = errors.Errorf("unknown filesystem driver %s for snapshot %s", fsDriver, snapshotID)
 	}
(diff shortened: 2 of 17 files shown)


changweige (commented):
We can rebase this PR now since its preceding changes have been merged

snapshot/mount_option.go
```diff
@@ -97,3 +99,178 @@ func (o *snapshotter) remoteMountWithExtraOptions(ctx context.Context, s storage
...
+type DmVerityInfo struct {
+	HashType  string `json:"hashtype"`
+	Hash      string `json:"hash"`
+	BlockNum  uint64 `json:"blocknum"`
+	Blocksize uint64 `json:"blocksize"`
+	Hashsize  uint64 `json:"hashsize"`
+	Offset    uint64 `json:"offset"`
+}
+
+func (d *DmVerityInfo) IsValid() error {
```
imeoer, line 265: How about `Validate() error`

snapshot/mount_option.go
```diff
@@ -97,3 +99,178 @@ func (o *snapshotter) remoteMountWithExtraOptions(ctx context.Context, s storage
...
+	case "sha256":
+		return d.isValidHash(64, "sha256")
+	case "sha1":
+		return d.isValidHash(40, "sha1")
+	default:
+		return fmt.Errorf("Unsupported hash algorithm %s for DmVerity device %s", d.HashType, d.Hash)
+	}
+}
+
+func (d *DmVerityInfo) isValidHash(expectedLen int, hashType string) error {
```
imeoer, line 297: How about `validateHash(...) error`

pkg/filesystem/fs.go
```diff
@@ -352,61 +350,66 @@ func (fs *Filesystem) Mount(snapshotID string, labels map[string]string, s *stor
...
-	if err := fsManager.AddRafsInstance(rafs); err != nil {
-		return errors.Wrapf(err, "create instance %s", snapshotID)
+	if err == nil {
+		if err := fsManager.AddRafsInstance(rafs); err != nil {
+			return errors.Wrapf(err, "create instance %s", snapshotID)
+		}
+	}
+
+	if err != nil {
+		_ = fs.Umount(ctx, snapshotID)
```
imeoer, line 380: I think we'd better ensure the `umounted` state is always true when `mount` error happens, instead of handling it outside `mount` func.

snapshot/snapshot.go
```diff
@@ -98,6 +93,22 @@ func NewSnapshotter(ctx context.Context, cfg *config.SnapshotterConfig) (snapsho
...
+		config, err := daemonconfig.NewDaemonConfig(config.GetFsDriver(), cfg.DaemonConfig.NydusdConfigPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "load daemon configuration")
+		}
+		daemonConfig = &config
+		_, backendConfig := config.StorageBackend()
+		skipSSLVerify = backendConfig.SkipVerify
+	} else {
+		// TODO: config skipSSLKVerify for tarfs etc
+		skipSSLVerify = true
```
imeoer, line 109: Why the option can't be received from configuration?

snapshot/mount_option.go
```diff
@@ -97,3 +100,312 @@ func (o *snapshotter) remoteMountWithExtraOptions(ctx context.Context, s storage
...
+	Config      string `json:"config"`
+	SnapshotDir string `json:"snapshot_dir"`
+}
+
+func (n *NydusImageVolume) IsValid() bool {
+	return len(n.Config) > 0 || len(n.SnapshotDir) > 0
+}
+
+// KataVirtualVolume encapsulates information for extra mount options and direct volumes.
+type KataVirtualVolume struct {
```
bergwolf, line 349: Please reference Kata's implementation here so that it is easier to validate the format for future changes.

snapshot/mount_option.go
jiangliu: Fixed

snapshot/mount_option.go
jiangliu: Fixed

snapshot/mount_option.go
jiangliu: Yes, it's a temporary solution to speed up the process. It will be merged into one when both kata and nydus snapshotter have been settled down.

snapshot/snapshot.go
jiangliu: My bad, it should be
```
@@ -105,8 +105,7 @@ func NewSnapshotter(ctx context.Context, cfg *config.SnapshotterConfig) (snapsho
                _, backendConfig := config.StorageBackend()
                skipSSLVerify = backendConfig.SkipVerify
        } else {
-               // TODO: config skipSSLKVerify for tarfs etc
-               skipSSLVerify = true
+               skipSSLVerify = config.GetSkipSSLVerify()
        }
 
        var blockdevManager *mgr.Manager

```

pkg/filesystem/fs.go
jiangliu: It's a hard situation, error happens on error handling code path. Any suggestion about how to handling such a case?

jiangliu: I'm reusing `Umount()` to do error recover, do you mean I should not call `Umount()` here?

imeoer: Yes, how about putting `defer Umount()` into `mountRemote()` and `MountTarErofs()`? The returned error of these funcs indicates that the `mount` operation has failed, it should be `umounted` state now. But I think maybe it's not a blocker for this PR.

imeoer:
May be add some usage docs later.

sctb512:
Maybe also need to take some E2E tests?

adamqqqplay:
Maybe need to add some unit tests, coverage has dropped too much.

jiangliu:
Due to the tight schedule, we will add doc/tests with following-on PRs.

changweige (commented):
Otherwise, LGTM

internal/constant/values.go
```diff
@@ -25,6 +25,8 @@ const (
 	FsDriverFscache string = "fscache"
 	// Only prepare/supply meta/data blobs, do not mount RAFS filesystem.
 	FsDriverNodev string = "nodev"
+	// Relay layer content download operation to other agents.
+	FsDriverProxy string = "proxy"
```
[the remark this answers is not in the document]
jiangliu: I have a version which combines proxy into nodev, but that cause some troubles. Then split it out as `proxy` mode.
`proxy` mode is to relay all layers to external agents,  `nodev` just relay nydus data layers to external agents.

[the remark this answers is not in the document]
jiangliu: One possible way is to define submode for proxy, "ociv1", "nydus" etc. 
I have a plan to refactor `nodev`/`DaemonNone` related code next step, I will give another try to merge `proxy` and `nodev` then.

changweige (approved):
LGTM, thanks

PR #531, opened 2023-09-07
Manager: make nydusd daemon reach desired state

sctb512:
In the faileover process, if the new daemon is unexpectedly killed before the new daemon is subscribed to by monitor, the manager will not retry to start the daemon to ensure that the new daemon runs successfully.

In this PR, the manager will retry to start the daemon if the process is not running.

Test command:
```shell
for i in $(seq 500);do ps -ef | grep nydusd| grep -v grep | awk '{print $2}'|xargs kill -9;sleep 0.01; done
```

Expected result: all nydusd will eventually recover.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -582,6 +582,20 @@ func (d *Daemon) Wait() error {
 	return nil
 }
 
+func (d *Daemon) IsProcessRunning() (bool, error) {
+	d.Lock()
+	defer d.Unlock()
+
+	if d.Pid() > 0 {
+		if _, err := os.FindProcess(d.Pid()); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+
+	return false, nil
+}
+
 // When daemon dies, clean up its vestige before start a new one.
 func (d *Daemon) ClearVestige() {
 	mounter := mount.Mounter{}
--- a/pkg/manager/daemon_adaptor.go
+++ b/pkg/manager/daemon_adaptor.go
@@ -25,6 +25,28 @@ import (
 	metrics "github.com/containerd/nydus-snapshotter/pkg/metrics/tool"
 )
 
+func (m *Manager) IsSubscribedDaemon(id, path string) bool {
+	return m.monitor.IsSubscribed(id, path)
+}
+
+func (m *Manager) StartDaemonUntilSubscribed(d *daemon.Daemon) {
+	for {
+		if err := m.StartDaemon(d); err != nil {
+			log.L.WithError(err).Errorf("fail to start daemon %s when recovering", d.ID())
+		}
+
+		running, err := d.IsProcessRunning()
+		if err != nil {
+			log.L.WithError(err).Errorf("fail to get process state, pid %d", d.Pid())
+		}
+		if m.IsSubscribedDaemon(d.ID(), d.GetAPISock()) && running {
+			break
+		}
+
+		log.L.Warnf("daemon %s, retry...", d.ID())
+	}
+}
+
 // Fork the nydusd daemon with the process PID decided
 func (m *Manager) StartDaemon(d *daemon.Daemon) error {
 	cmd, err := m.BuildDaemonCommand(d, "", false)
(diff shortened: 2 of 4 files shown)


pkg/manager/manager.go
```diff
@@ -206,9 +216,20 @@ func (m *Manager) doDaemonRestart(d *daemon.Daemon) {
 	}
 
 	d.ClearVestige()
-	if err := m.StartDaemon(d); err != nil {
-		log.L.Errorf("fails to start daemon %s when recovering", d.ID())
-		return
+	for {
```
imeoer, line 219: Maybe wrap the code block into a func so that it can be reused.

pkg/manager/daemon_adaptor.go
```diff
@@ -25,6 +25,28 @@ import (
...
+
+func (m *Manager) StartDaemonUntilSubscribed(d *daemon.Daemon) {
+	for {
+		if err := m.StartDaemon(d); err != nil {
+			log.L.Errorf("fail to start daemon %s when recovering", d.ID())
+		}
+
+		running, err := d.IsProcessRunning()
+		if err != nil {
+			log.L.Errorf("fail to get process state, pid %d", d.Pid())
```
imeoer, line 40: Better to throw the `err.Error()` message together.

imeoer (commented):
others LGTM!

pkg/manager/daemon_adaptor.go
sctb512: Fixed.

pkg/manager/manager.go
```diff
@@ -176,15 +176,10 @@ func (m *Manager) doDaemonFailover(d *daemon.Daemon) {
 	}
 
 	// Failover nydusd still depends on the old supervisor
-
-	if err := m.StartDaemon(d); err != nil {
-		log.L.Errorf("fail to start daemon %s when recovering", d.ID())
-		return
-	}
+	m.StartDaemonUntilSubscribed(d)
```
changweige, line 179: Does the retrial never stop?

pkg/manager/manager.go
```diff
@@ -176,15 +176,10 @@ func (m *Manager) doDaemonFailover(d *daemon.Daemon) {
...
 
 	if err := d.WaitUntilState(types.DaemonStateInit); err != nil {
-		log.L.WithError(err).Errorf("daemon didn't reach state %s,", types.DaemonStateInit)
-		return
+		log.L.WithError(err).Errorf("daemon didn't reach state %s", types.DaemonStateInit)
```
changweige, line 182: If the daemon fails to reach `Init` state, we can't invoke `takeOver` API, it conflicts the nydusd internal state machine

pkg/daemon/daemon.go
```diff
@@ -582,6 +582,20 @@ func (d *Daemon) Wait() error {
 	return nil
 }
 
+func (d *Daemon) IsProcessRunning() (bool, error) {
+	d.Lock()
+	defer d.Unlock()
```
changweige, line 587: Why need a lock when finding process

pkg/daemon/daemon.go
```diff
@@ -582,6 +582,20 @@ func (d *Daemon) Wait() error {
...
+
+	if d.Pid() > 0 {
+		if _, err := os.FindProcess(d.Pid()); err != nil {
```
changweige, line 590: It's better to judge the process state if it is a runnable state

pkg/manager/daemon_adaptor.go
```diff
@@ -25,6 +25,28 @@ import (
...
 
+func (m *Manager) IsSubscribedDaemon(id, path string) bool {
+	return m.monitor.IsSubscribed(id, path)
+}
+
+func (m *Manager) StartDaemonUntilSubscribed(d *daemon.Daemon) {
+	for {
+		if err := m.StartDaemon(d); err != nil {
+			log.L.WithError(err).Errorf("fail to start daemon %s when recovering", d.ID())
+		}
```
changweige, line 36: Even it fails to startDaemon, we do nothing but continue to process ?

PR #534, opened 2023-09-12
manager: avoid to start daemon in dead loop 

sctb512:
Revert https://github.com/containerd/nydus-snapshotter/pull/531 
If nydusd is not subscribed and is unexpectedly killed, send an event to the death event handler.


changweige (approved):
LGTM, thanks

PR #536, opened 2023-09-14
cache: fix null pointer access when cachemanager.Disable is true

jiangliu:
Filesystem.cacheMgr is null when cacheManager.Disable is configured to be true, thus cause null pointer access. Fix it by always create the cacheMgr object no matter cacheManager.Disable is true or false.

### DIFF
--- a/pkg/cache/manager.go
+++ b/pkg/cache/manager.go
@@ -37,6 +37,7 @@ type Manager struct {
 }
 
 type Opt struct {
+	Disabled bool
 	CacheDir string
 	Period   time.Duration
 	Database *store.Database
--- a/misc/snapshotter/config.toml
+++ b/misc/snapshotter/config.toml
@@ -35,7 +35,6 @@ threads_number = 4
 # Log rotation size for nydusd, in unit MB(megabytes)
 log_rotation_size = 100
 
-
 [cgroup]
 # Whether to use separate cgroup for nydusd.
 enable = true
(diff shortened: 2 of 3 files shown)


sctb512:
Why not check if the cacheMgr is nil?
It seems `WithEnableStargz` and `referrerMgr`,  also have the same problem.

pkg/cache/manager.go
```diff
@@ -37,6 +37,7 @@ type Manager struct {
 }
 
 type Opt struct {
+	Disabled bool
```
imeoer, line 40: Good catch! Maybe also update these `config.toml`:

- misc/snapshotter/config.toml
- tests/e2e/k8s/snapshotter-cri.yaml
- tests/e2e/k8s/snapshotter-kubeconf.yaml

jiangliu:
> Why not check if the cacheMgr is nil? It seems `stargzResolver` and `referrerMgr` also have the same problem.

With the evolution of nydus-snapshotter, `cache manager` becomes mandatory instead optional now. On the other hand, `stargzResolver` and `referrerMgr` are optional.

pkg/cache/manager.go
jiangliu: `cache_manager.disable` is still set to be `false` for those three files. I'm keeping the configuration item for compatibility though it doesn't take effect. Not sure whether `cache_manager.disable` will be used to control `auto garbage collect` in future or not.

jiangliu: After a second thought, I think we can support `recyclebin` to keep deleted files for a while. Those configuration item can be used for it.

sctb512 (approved):
LGTM

pkg/cache/manager.go
imeoer: Introducing recyclebin seems to add complexity and doesn't feel necessary, it is better to let kubelet & containerd control the GC policy.

changweige (approved):
LGTM

changweige (approved):
LGTM

PR #538, opened 2023-09-19
Enable retry and recover for tarfs

jiangliu:
Enhance tarfs implementation by:
- retry downloading from registry if the background downloading task fails
- recover snapshot information related to tarfs on startup
- remount EROFS associated with tarfs when restart
- avoid redundant merge operation if the image has only one layer
- auto detach loop devices when they are not used anymore


pkg/tarfs/tarfs.go
```diff
@@ -715,32 +693,52 @@ func (t *Manager) DetachLayer(snapshotID string) error {
...
-func (t *Manager) getSnapshotStatus(snapshotID string, lock bool) (*snapshotStatus, error) {
+func (t *Manager) getSnapshotStatus(snapshotID string) (*snapshotStatus, error) {
 	t.mutex.Lock()
 	defer t.mutex.Unlock()
 	st, ok := t.snapshotMap[snapshotID]
 	if ok {
-		if lock {
-			st.mutex.Lock()
-		}
+		st.mutex.Lock()
```
sctb512, line 947: Why lock this by default, Can we add lock and release it after calling `getSnapshotStatus`?

Otherwise, change `getSnapshotStatus` to `getSnapshotStatusWithLock`?

pkg/tarfs/tarfs.go
```diff
@@ -715,32 +693,52 @@ func (t *Manager) DetachLayer(snapshotID string) error {
...
+			return nil, err
+		}
 	}
-	st.wg.Wait()
+
 	if st.status != TarfsStatusReady {
-		return errors.Errorf("snapshot %s is in state %d instead of ready state", snapshotID, st.status)
+		st.mutex.Unlock()
+		var state string
+		switch st.status {
```
sctb512, line 973: Maybe we can wrap this into a method?

pkg/tarfs/tarfs.go
```diff
@@ -305,83 +305,86 @@ func (t *Manager) getImageBlobInfo(metaFilePath string) (string, error) {
...
+				msg := fmt.Sprintf("Nydus tarfs for snapshot %s is ready", snapshotID)
+				return epilog(nil, msg)
 			}
 		}
-	}()
+	}
+
+	keyChain, err := auth.GetKeyChainByRef(ref, nil)
+	if err != nil {
+		epilog(err, "create key chain for connection")
```
sctb512: It seems the `epilog(err, "create key chain for connection")` should be the value of `return`?

pkg/tarfs/tarfs.go
```diff
@@ -362,33 +362,85 @@ func (t *Manager) blobProcess(ctx context.Context, snapshotID, ref string,
...
 }
 
+func (t *Manager) retryPrepareLayer(snapshotID, upperDirPath string, labels map[string]string) error {
+	ref, ok := labels[label.CRIImageRef]
+	if !ok {
+		return errors.Errorf("not found image reference label")
+	}
+	layerDigest := digest.Digest(labels[label.CRILayerDigest])
+	if layerDigest.Validate() != nil {
+		return errors.Errorf("not found layer digest label")
```
sctb512, line 408: Maybe wrap the error of `Validate()`?  The `err != nil` does not indicate  `not found layer digest label`.

pkg/tarfs/tarfs.go
```diff
@@ -362,33 +362,85 @@ func (t *Manager) blobProcess(ctx context.Context, snapshotID, ref string,
...
+	}
+	manifestDigest := digest.Digest(labels[label.CRIManifestDigest])
+	if manifestDigest.Validate() != nil {
+		return errors.Errorf("not found manifest digest label")
```
sctb512, line 412: ditto.

pkg/tarfs/losetup_linux.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	return unix.IoctlSetInt(int(file.Fd()), unix.LOOP_CLR_FD, 0)
+}
+
+// AttachLoopDevice attaches a specified backing file to a loop device
+func AttachLoopDevice(backingFile string) (string, error) {
```
sctb512, line 188: Is it enough if we use it by importing `"github.com/containerd/containerd/mount"`?

pkg/tarfs/tarfs.go
```diff
@@ -768,6 +851,43 @@ func (t *Manager) DetachLayer(snapshotID string) error {
...
+	if _, ok := t.snapshotMap[id]; ok {
+		// RecoverSnapshotInfo() is called after RecoverRafsInstance(), so there may be some snapshots already exist.
+		return nil
+	}
+
+	layerMetaFilePath := t.layerMetaFilePath(upperPath)
+	if _, err := os.Stat(layerMetaFilePath); err == nil {
+		layerDigest := digest.Digest(info.Labels[label.CRILayerDigest])
+		if layerDigest.Validate() != nil {
+			return errors.Errorf("not found layer digest label")
```
sctb512, line 868: ditto.

pkg/tarfs/tarfs.go
jiangliu: It's ensure atomicity by acquiring the lock on `snapshotStatus` with the `Manager.mutex" locked.
So renamed `getSnapshotStatus` to `getSnapshotStatusWithLock`.

pkg/tarfs/losetup_linux.go
jiangliu: Some modification is needed, so coopied it here.

pkg/tarfs/tarfs.go
jiangliu: Fixed

pkg/tarfs/tarfs.go
jiangliu: Fixed

pkg/tarfs/tarfs.go
jiangliu: Fixed

pkg/tarfs/tarfs.go
jiangliu: Fixed

pkg/tarfs/tarfs.go
jiangliu: fixed.

pkg/tarfs/tarfs_test.go
ChengyuZhu6, line 1: Should we add a test to verify that the download retry feature works as expected?

jiangliu: done

.github/workflows/ci.yml
```diff
@@ -28,6 +28,14 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go
+      - name: Setup Nydus
+        run: |
+          # Download nydus components
+          NYDUS_VER=v$(curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -s "https://api.github.com/repos/dragonflyoss/nydus/releases/latest" | jq -r .tag_name | sed 's/^v//')
```
bergwolf, line 34: You don't need auth info to access the GH public release info file.

jiangliu: Fixed

.github/workflows/ci.yml
```diff
@@ -135,6 +143,14 @@ jobs:
...
```
imeoer, line 149: No auth is required when requesting the latest version.

pkg/tarfs/tarfs.go
```diff
@@ -306,88 +322,145 @@ func (t *Manager) getImageBlobInfo(metaFilePath string) (string, error) {
...
+	}
+
+	if retry {
+		// Download and convert layer content in synchronous mode when retry for error recovering
+		err = process(rc, remote)
+	} else {
+		// Download and convert layer content in background.
+		// Will retry when the content is actually needed if the background process failed.
+		go func() {
+			_ = process(rc, remote)
```
imeoer, line 411: It should also give a warning log if an error happens.

imeoer:
Can we add some e2e test cases in https://github.com/containerd/nydus-snapshotter/blob/main/integration/entrypoint.sh?

PR #540, opened 2023-09-21
e2e: always disable erofs ondemand-read test and failover test

changweige:
Github action has upgraded its worker kernel to 6.2 with CONFIG_EROFS_FS_ONDEMAND enabled. The e2e tests of failover depends on private kernel patches rather upstream kernel. So we have to totally disable the failover tests for now

### DIFF
--- a/integration/entrypoint.sh
+++ b/integration/entrypoint.sh
@@ -102,8 +102,9 @@ function retry {
 }
 
 function can_erofs_ondemand_read {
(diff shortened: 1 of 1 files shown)


sctb512 (approved):
LGTM

sctb512:
Sometimes, the optimizer test also fails. It seems we should verify if the containerd is ready before generating the accessed list. I will post a PR soon for it.

PR #542, opened 2023-09-28
converter: move out encryption package

imeoer:
```
  Due to the introduction of the github.com/containers/ocicrypt package
  in converter, which brings in many other dependencies, but encryption
  should currently be an experimental optional feature, we need to move
  this part of the codes out to reduce the burden on other projects that
  reference the package.
```

Ref: https://github.com/moby/buildkit/pull/4288


imeoer:
@changweige @jiangliu @taoohong This PR is blocking https://github.com/moby/buildkit/pull/4288, let's go forward to release a minor version, and then we can adjust it on the next release again.

PR #545, opened 2023-10-19
converter: fix missed BatchSize arg for dir-rafs packing

hangvane:
dir-rafs lacks the `batch-size` build parameter,
making the converting failed when tar-rafs is disabled and fallbacked to dir-rafs


imeoer (approved):
LGTM!

Issue #548, opened 2023-10-23
release: Provide s390x and arm64 released tarballs


fidencio:
Confidential Containers is now depending on the nydus-snapshotter, and having an official release for the architecures we support (amd64, which is already there; and s390x) and will support in the future (arm64) would be amazing!

imeoer:
Have you time to take a look? @sctb512 @changweige

sctb512:
> Have you time to take a look? @sctb512 @changweige

Yes, it is necessary to release multi-architecture packages. I will look into it soon.

sctb512:
The binaries we released are now statically linked.
```shell
$ wget https://github.com/containerd/nydus-snapshotter/releases/download/v0.13.3/nydus-snapshotter-v0.13.3-x86_64.tgz
$ tar -zxvf nydus-snapshotter-v0.13.3-x86_64.tgz
$ ldd nydus-snapshotter/*
nydus-snapshotter/containerd-nydus-grpc:
	not a dynamic executable
nydus-snapshotter/nydus-overlayfs:
	not a dynamic executable
nydus-snapshotter/optimizer-nri-plugin:
	not a dynamic executable
nydus-snapshotter/optimizer-server:
	statically linked
```

So, I believe it can run on different architectures now.

sctb512:
Additionally, we can release packages for different architectures as well.  See https://github.com/containerd/containerd/releases/tag/v1.7.7

sctb512:
Related to https://github.com/containerd/nydus-snapshotter/pull/418

PR #552, opened 2023-10-27
release: Upload s390x tarball

fidencio:
This series is the very start to allow shipping tarball including binaries for different architectures than x86_64.

The focus here is s390x, which is used by Confidential Containers, but this can easily be expanded in the future.

Last but not least, for now we have *no* interest in publishing a multi-architecture image.

Fixes: https://github.com/containerd/nydus-snapshotter/issues/548

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,12 +12,26 @@ env:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        goarch:
+          - amd64
+          - s390x
+        include:
+          - goarch: amd64
+            rust_arch: x86_64
+          - goarch: s390x
+            rust_arch: s390x
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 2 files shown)


fidencio:
https://github.com/fidencio/nydus-snapshotter/actions/runs/6670584297 is an example of the action successfully running.

fidencio:
Ideally, we'd need a release as soon as this one gets merged. :-)

bergwolf:
Unfortunately, we haven't provided s390x nydusd yet. And there's a build failure on s390x. See https://github.com/cloud-hypervisor/fuse-backend-rs/pull/170

bergwolf:
@fidencio No, it does not block this PR but nydus-snapshotter does not work w/o nydusd. So the functionality is still blocked.

fidencio:
> @fidencio No, it does not block this PR but nydus-snapshotter does not work w/o nydusd. So the functionality is still blocked.

It actually does for the confidential containers use case we don't actually need nydusd for pulling the image in the guest.

fidencio:
I'm closing this one in favour of #553.

PR #554, opened 2023-11-03
daemon: change `amplify_io` into nullable int

hangvane:
`amplify_io` should be nullable in config file.
The absence of `amplify_io` in the config file means it follows the default value in nydusd, i.e., 1M.
`amplify_io=0` means it is set to zero.
So let's change this field into nullable to pass the value as what it is read from template config file.


changweige (approved):
LGTM

imeoer (approved):
Thanks!

PR #558, opened 2023-11-27
Fix incorrect nydusd count metric

sctb512:
Currently, there may be a negative nydusd count in some situations.

We only need to add this value when the daemon is running and decrease it when the daemon is actually shut down.


pkg/utils/mount/mount.go
```diff
@@ -65,11 +85,17 @@ func IsMountpoint(path string) (bool, error) {
...
 	if err != nil {
+		if strings.Contains(err.Error(), "transport endpoint is not connected") {
+			return true, errdefs.ErrNotConnected
+		}
 		return false, err
 	}
 
 	parentStat, err := os.Stat(filepath.Dir(realPath))
 	if err != nil {
+		if strings.Contains(err.Error(), "transport endpoint is not connected") {
```
changweige, line 96: This is not a reliable way to judge if the dir is a mount point. Moreover,  did you try to catch the standard Linux error code rather than matching error string?

sctb512: Yes, got it.

changweige (approved):
LGTM

PR #561, opened 2023-12-04
release: get latest nydusd version from Github API

adamqqqplay:
In our released nydus-snapshotter images, the nydusd version is always v2.1.5, and this patch fixes that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,6 +104,12 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+      - name: Get the nydusd version
+        id: step_one
+        run: |
+          export NYDUS_STABLE_VER=$(curl https://api.github.com/repos/dragonflyoss/nydus/releases/latest | jq -r .tag_name)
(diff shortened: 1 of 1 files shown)


sctb512:
> I haven't found a way to test the modified release.yml yet. I wonder if the maintainer has a better way to test it. Thanks in advance!

Maybe you can push a tag to your own repository.

.github/workflows/release.yml
```diff
@@ -104,6 +104,11 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+      - name: Get the nydusd version
+        id: step_one
+        run: |
+          export NYDUS_VER=$(curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' -s "https://api.github.com/repos/dragonflyoss/image-service/releases/latest" | jq -r .tag_name | sed 's/^v//')
```
imeoer, line 110: It seems no need the authorization.

adamqqqplay: Yeah, I fixed it.

adamqqqplay:
@sctb512 @imeoer The image build was successful at: ghcr.io/adamqqqplay/nydus-snapshotter:v0.0.2, which includes the v2.2.4 nydusd version.

imeoer (approved):

PR #566, opened 2023-12-11
replace docker/distribution for github.com/distribution/reference v0.5.0

thaJeztah:
The "reference' package has now moved to a separate repository, which means we can replace the local fork, and use the upstream implementation again.

The new module was extracted from the distribution repository at commit: https://github.com/distribution/distribution/commit/b9b19409cf458dcb9e1253ff44ba75bd0620faa6


imeoer (approved):
LGTM!

PR #569, opened 2023-12-15
misc: add daemon_mode option to guest pull config 

### LINKED ISSUE #568: misc: add daemon_mode option to guest pull config
`daemon_mode` should be specified in guest pull config, even if it is none.
This ensures that guest pull config is consistent with other configs and avoids confusion.

ChengyuZhu6:
`daemon_mode` should be specified in guest pull config, even if it is none.
This ensures that guest pull config is consistent with other configs and avoids confusion.

Fixes #568


imeoer (approved):
LGTM! Thanks.

PR #576, opened 2024-01-30
action: fix failed image build

imeoer:
The nydus latest reelase version should be got from dragonfly/nydus repo.

Tested CI: https://github.com/imeoer/nydus-snapshotter/actions/runs/7709106171/job/21009677183


adamqqqplay (approved):
LGTM

PR #582, opened 2024-02-22
misc: fix incorrect nydus-overlayfs version

imeoer:
1. The nydus-overlayfs component has been moved into the cmd/nydus-overlayfs directory in snapshotter repository, we should also use the nydus-overlayfs binary in docker image.

2. Bump the nydus binary version v2.2.4, it's the latest stable release.


adamqqqplay (approved):
Thanks, let’s see if we can get the latest version tag through the Github API next time.

PR #586, opened 2024-03-20
misc: Fix cleanup with k3s / rke2

fidencio:
When performing the cleanup with k3s / rke2, we must specify the containerd socket address that's been created by them, instead of using the default one.

For both k3s and rke2 the containerd address socket is the same: `/run/k3s/containerd/containerd.sock`.

### DIFF
--- a/misc/snapshotter/snapshotter.sh
+++ b/misc/snapshotter/snapshotter.sh
@@ -197,8 +197,13 @@ function cleanup_snapshotter() {
 
     pid=$(ps -ef | grep containerd-nydus-grpc | grep -v grep | awk '{print $1}')
     if [ ! -z "$pid" ]; then
-        for i in $(nsenter -t 1 -m ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1); do
-            nsenter -t 1 -m ctr -n k8s.io snapshot --snapshotter nydus rm $i || true
+        local ctr_args=""
+        if [[ " k3s k3s-agent rke2-agent rke2-server " =~ " ${CONTAINER_RUNTIME} " ]]; then
+            ctr_args="--address /run/k3s/containerd/containerd.sock "
+        fi
+        ctr_args+="--namespace k8s.io snapshot --snapshotter nydus"
+        for i in $(nsenter -t 1 -m ctr ${ctr_args} list | grep -v KEY | cut -d' ' -f1); do
+            nsenter -t 1 -m ctr ${ctr_args} rm $i || true
         done
     fi
     echo "Recover containerd config"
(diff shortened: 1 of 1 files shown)


ChengyuZhu6 (approved):
LGTM thanks @fidencio !

fidencio:
I don't think the E2E failures are related to my changes.

misc/snapshotter/snapshotter.sh
```diff
@@ -197,8 +197,13 @@ function cleanup_snapshotter() {
 
     pid=$(ps -ef | grep containerd-nydus-grpc | grep -v grep | awk '{print $1}')
     if [ ! -z "$pid" ]; then
-        for i in $(nsenter -t 1 -m ctr -n k8s.io snapshot --snapshotter nydus list | grep -v KEY | cut -d' ' -f1); do
-            nsenter -t 1 -m ctr -n k8s.io snapshot --snapshotter nydus rm $i || true
+        local ctr_args=""
+        if [ "${CONTAINER_RUNTIME}" == "k3s" ] || [ "${CONTAINER_RUNTIME}" == "k3s-agent" ] || [ "${CONTAINER_RUNTIME}" == "rke2-agent" ] || [ "${CONTAINER_RUNTIME}" == "rke2-server" ]; then
```
wainersm, line 201: Hi @fidencio !

`if [[ " k3s k3s-agent rke2-agent rke2-server " =~ " ${CONTAINER_RUNTIME} " ]]; then` to simplify.

wainersm: hmmm I'm not sure k3s/rk2 are strictly a `CONTAINER_RUNTIME` . Wouldn't be better introduce a new variable, e.g., `${CONTAINER_RUNTIME_SOCK:-}` that the caller would export if it wants something different from the default? We also avoid having code in snapshotter to handle different Kubernetes distributions.

fidencio: This follows the pattern already present in other parts of the script, @wainersm.

fidencio: > hmmm I'm not sure k3s/rk2 are strictly a `CONTAINER_RUNTIME` . Wouldn't be better introduce a new variable, e.g., `${CONTAINER_RUNTIME_SOCK:-}` that the caller would export if it wants something different from the default? We also avoid having code in snapshotter to handle different Kubernetes distributions.

That's not exactly a good idea.  We want the experience to just work for whoever is using whatever k8s distro.
The user usually doesn't know and shouldn't know about the containerd socket for k3s, things should just work, and I think in the same way we have the overlays and yatta-yatta, this is our responsiblity as developers to hide it from them, actually making their lives easier.

ChengyuZhu6:
cc @imeoer

imeoer:
LGTM, thanks! @fidencio

imeoer:
@fidencio Will release `v0.13.11` if we needed.

ChengyuZhu6:
> @fidencio Will release `v0.13.11` if we needed.

Thanks @imeoer. Please release a new version.

PR #591, opened 2024-04-25
converter: support appending files to bootstrap layer

imeoer:
Support `AppendFiles` option on `converter.Merge` method,
`AppendFiles` specifies the files that need to be appended
to the bootstrap layer.

This option allows the user to place some files in the nydus
bootstrap layer, for example, the user can place metadata
info generated at build time, e.g. named `image/metadata.json`.

And this usage does not affect the `image/image.boot` bootstrap
file that is accessed by the nydus runtime.

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containerd/containerd/archive"
 	"github.com/containerd/containerd/archive/compression"
 	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/content/local"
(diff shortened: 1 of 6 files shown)


sctb512:
LGTM.

Is there any example for this PR?

sctb512 (approved):

adamqqqplay (approved):
LGTM, Thanks!

It would be better if you could provide some docs.

imeoer:
> LGTM.
> 
> Is there any example for this PR?

Haven't yet, will submit in another PR, thanks!

PR #598, opened 2024-06-21
go.mod: github.com/containerd/containerd v1.7.7, containerd/nri v0.4.0, switch to github.com/containerd/log module

thaJeztah:
- relates to https://github.com/containerd/containerd/pull/9106
- relates to https://github.com/containerd/containerd/pull/9099
- relates to https://github.com/containerd/nri/pull/40

### go.mod: github.com/containerd/containerd v1.7.7, containerd/nri v0.4.0

Update to the first release of containerd to alias the log package to the new containerd/log module.

This update of the containerd module also comes with an updated version of github.com/containerd/nr, which has some breaking changes in the interface definitions.

the optimizer-nri-plugin and prefetchfilees-nri-pugin pacakges had to be updated due to [containerd/nri@9d86150] changing the interface to require a context to be passed.

[containerd/nri@9d86150]: https://github.com/containerd/nri/commit/9d86150fce4318491481e354d0f34c6b76e8806e

Full diffs:

- https://github.com/containerd/containerd/v1.7.0...v1.7.7
- https://github.com/containerd/nri/compare/v0.3.0...v0.4.0

### switch to github.com/containerd/log module

containerd 1.7.7 and up alias the log package to the new module,
and deprecate the package.

This patch:

- replaces existing uses of the log package to use the module
- replaces some direct uses of logrus to use the log module

There is one change in behavior in the nri-plugins, which now
use the default format as used by the containerd log module.


cmd/optimizer-nri-plugin/main.go
```diff
@@ -237,13 +239,13 @@ func main() {
 
 			cfg = flags.Args.Config
 
-			log = logrus.StandardLogger()
-			log.SetFormatter(&logrus.TextFormatter{
-				PadLevelText: true,
-			})
+			// FIXME(thaJeztah): ucontainerd's log does not set "PadLevelText: true"
+			_ = log.SetFormat(log.TextFormat)
+			ctx := log.WithLogger(context.Background(), log.L)
```
thaJeztah, line 244: This currently uses a different format than what was used before; 

- Now uses `RFC3339NanoFixed` format
- Now uses `FullTimestamp: true`
- :x: does not use `PadLevelText: true`


https://github.com/containerd/log/blob/v0.1.0/context.go#L153-L158


```go
	case TextFormat:
		L.Logger.SetFormatter(&logrus.TextFormatter{
			TimestampFormat: RFC3339NanoFixed,
			FullTimestamp:   true,
		})
		return nil
```

cmd/prefetchfiles-nri-plugin/main.go
```diff
@@ -146,35 +146,33 @@ func main() {
 				err  error
 			)
 
-			log = logrus.StandardLogger()
+			// FIXME(thaJeztah): ucontainerd's log does not set "PadLevelText: true"
+			_ = log.SetFormat(log.TextFormat)
+			ctx := log.WithLogger(context.Background(), log.L)
```
thaJeztah, line 151: Same here

thaJeztah:
@imeoer ptal 🤗

imeoer:
LGTM, thanks for the careful work! :)

thaJeztah:
@imeoer thanks! I also move https://github.com/containerd/nydus-snapshotter/pull/599 out of draft, but it looks like there's an issue with OTEL 🤔

Issue #402, opened 2023-03-07
Wasted space in image


vflaux:
37MB is wasted here as the `COPY` layer is completely overlayed by the `RUN chmod` one:
https://github.com/containerd/nydus-snapshotter/blob/1e7a12e61cae61f5b451aa1c0b8492805c17bfb5/misc/snapshotter/Dockerfile#L25-L26
Should use `COPY --chmod` instead from buildkit.

Dive output:
```
Analyzing image...
  efficiency: 74.4988 %
  wastedBytes: 73633235 bytes (74 MB)
  userWastedPercent: 53.6183 %
Inefficient Files:
Count  Wasted Space  File Path
    2         74 MB  /usr/local/bin/containerd-nydus-grpc
    2         30 kB  /lib/apk/db/installed
    2         22 kB  /lib/apk/db/scripts.tar
    2         152 B  /lib/apk/db/triggers
    2         131 B  /etc/apk/world
```

changweige:
Thanks for reporting. Could please help raise a PR to fix it?

PR #600, opened 2024-07-02
misc: merge chmod into copy in Dockerfile

SToPire:
Let's use Docker's `COPY --chmod` feature instead of using a separate chmod command. This will reduce the generated image size by ~12KB.

Fixes: #402


imeoer (approved):
LGTM!

SToPire:
Sorry for the late reply! It seems that COPY --chmod in Docker buildkit only supports octal notation instead of "+x". [1] 

This was my mistake and I've updated the new patch.  It looks like the current version could pass the CI correctly.

[1] https://docs.docker.com/reference/dockerfile/#copy

sctb512 (approved):
LGTM

PR #601, opened 2024-07-04
docs: fix incorrect file path for nginx result file

daiyongxuan:
The documentation incorrectly stated that the result file for the nginx image is located at `/opt/nri/optimizer/results/nginx:latest`. However, the correct file path is `/opt/nri/optimizer/results/library/nginx:latest`.


imeoer (approved):
LGTM, thanks!

PR #603, opened 2024-07-10
fix: error in Kubelet when using default root directory

fappy1234567:
## Relevant Issue 
Fixes #288 
## Details
Setting the snapshotter's default root path to /var/lib/containerd/io.containerd.snapshotter.v1.nydus can resolve the issue.

PR #606, opened 2024-07-23
CI fix, part I: update and cleanup

apostasie:
Work on #604 required a set of changes, fixes and updates to the test rig and CI, and has grown really large.

This PR does split out some of the more minor changes to make it easier to review.

Furthermore, the CI is in a shaky state right now and is just red, as the build fails because of delve latest being dependent on a more recent version of go (which in turns requires a series of changes), and also as #603 seems to break e2e kube tests (or this test is very flaky).

Changes:
- quiet apk, apt-get, wget and curl to make the logs less noisy and easier to read
- update github actions to their latest versions
- update golang to 1.21 - delve requires it - also, go 1.23 is soon to be out - and containerd v2 requires 1.22
- remove no longer useful variables
- increase sleep after kill to avoid racy metrics server restart causing port conflict
- bump and pin ubuntu to 22.04 (and golang from bullseye to bookworm) - fix GLIBC issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     name: Build and Lint
     timeout-minutes: 10
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -34,7 +34,7 @@ jobs:
   build-optimizer:
     name: Build optimizer
     timeout-minutes: 10
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Check out code
         uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


apostasie:
@imeoer fixing kube tests will take more work, that I would like to do in a separate PR, so, I am removing my "kube" commit from here for now.

This here is just dealing with the rest.

Do you mind having a look at this at your convenience so that we can rebase other ongoing work?
I believe this will fix a bunch of CI issues.

apostasie:
Also, I did split "unrelated" changes into separate commits so that "history" is easier to review, but of course happy to squash if you prefer that.

apostasie:
Note: we might want to pin delve to a specific version to avoid being forced into the same situation wrt golang version later on.

imeoer:
LGTM, thanks for the hard work!

apostasie:
You welcome @imeoer :)

PR #607, opened 2024-07-25
CI fix, part II: kube

apostasie:
This is fixing the recent kube testing bustage.

NOTE that this PR is on-top of #606 as I needed a working CI environment as a starting point to debug.
We can either merge this and close 606, or merge 606 first, then I will rebase here.

As far as I can tell, the issues with kube were:
1. fallout of #603 that did miss a bunch of strings
2. containerd-nydus-grpc was built *dynamically* on the host, but copied into the kind control plane container that could (and does) have a different libc
3. the rig was tightly coupled with github actions, making it really hard to test locally. Specifically, we depended upon engineerd/setup-kind which has been untouched in two years and is using deprecated features from github action anyhow

What this PR does (in commit "Fix kube test bustage"):
- add a new make `static` target so that we get truly static `bin/containerd-nydus-grpc` binaries that we can safely copy into the kind control plane
- finish replacing `/var/lib/containerd/io.containerd.snapshotter.v1.nydus` (follow-up to #603)
- pin delve version so that we don't have the go version problem anymore in the future (mentioned in #606)
- bump nerdctl version
- replace the entire github-action-coupled rig setup by a `kind.sh` shell script (under `tests/helpers`) that takes care of setting everything up, so that we can also use that locally / outside of github actions

CI is now green and both PRs are ready to review.

Hope this is helpful.


apostasie:
@imeoer going to rebase this ASAP so that we can consider just the Kube fixes

.github/workflows/k8s-e2e.yml
```diff
@@ -25,115 +26,22 @@ jobs:
         with:
           go-version-file: 'go.mod'
           cache-dependency-path: "go.sum"
-      - name: Setup Kind
```
apostasie, line 28: All of this logic, currently tied into github actions / engineerd/setup-kind, is making it (quite) difficult to test locally.

With this proposed PR, the logic gets moved to an independent shell script that should make it much easier to test locally, and also to control what version of what gets used.

Makefile
```diff
@@ -58,6 +58,11 @@ build:
 	GOOS=${GOOS} GOARCH=${GOARCH} ${PROXY} go build -ldflags "$(LDFLAGS)" -v -o bin/containerd-nydus-grpc ./cmd/containerd-nydus-grpc
 	GOOS=${GOOS} GOARCH=${GOARCH} ${PROXY} go build -ldflags "$(LDFLAGS)" -v -o bin/nydus-overlayfs ./cmd/nydus-overlayfs
 
+.PHONY: static
```
apostasie, line 61: Because we do not know what is going inside the kind control plane where the binary will get copied, we do not a static build. There is already a 'release-static' target, but it does require rust / the optimization thing.

docs/tarfs.md
```diff
@@ -24,21 +24,21 @@ $ nerdctl run --snapshotter nydus --rm nginx
 
 # Show mounted rootfs a container
 $ mount
-/dev/loop17 on /var/lib/containerd-nydus/snapshots/7/mnt type erofs (ro,relatime,user_xattr,acl,cache_strategy=readaround)
```
apostasie, line 27: This, and any other similar change is just to complete the transition started in the earlier PR.

integration/Dockerfile
```diff
@@ -1,11 +1,11 @@
-
 ARG CONTAINERD_VER=1.6.15
 ARG CONTAINERD_PROJECT=/containerd
 ARG RUNC_VERSION=1.1.4
 ARG NYDUS_SNAPSHOTTER_PROJECT=/nydus-snapshotter
 ARG DOWNLOADS_MIRROR="https://github.com"
 ARG NYDUS_VER=2.2.4
-ARG NERDCTL_VER=1.0.0
+ARG NERDCTL_VER=1.7.6
+ARG DELVE_VER=1.23.0
```
apostasie, line 8: Pinning delve will avoid one of the recent breakage of the CI, where the latest delve started depending on go features only available in a more recent version of go that what we used here.

tests/helpers/helpers.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
```
apostasie, line 1: Generic helpers, that are not specific to nydus, and help with making shellscript less painful :)

tests/helpers/kind.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
```
apostasie, line 1: The kind script that replaces the github action logic and can be used locally.

tests/helpers/lib.sh
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env bash
```
apostasie, line 1: Generic bash helpers, not specific to nydus (provides logging, http, heathcheck, minimal github client).

apostasie:
@imeoer rebased - one commit left - CI is green.

I just added a bunch of inline comments to facilitate review.
PTAL at your convenience and LMK if you want anything changed in this.

Thanks for your help!

Makefile
imeoer: Thanks, how about splitting `static-release` into `static` and `static-optimizer`? but the current changes is also LGTM.

imeoer (approved):
LGTM, thanks!

Makefile
apostasie: Yep, cleaning-up the Makefile in that context would definitely make sense.

PR #608, opened 2024-07-25
go.mod: dario.cat/mergo v1.0.0

thaJeztah:
- relates to https://github.com/containerd/containerd/pull/9569

The module was renamed to dario.cat/mergo. The old dependency isn't fully gone yet, because kubernetes still uses the old location, but that may go away eventually; https://github.com/kubernetes/client-go/blob/v0.26.2/go.mod#L18

full diff: https://github.com/darccio/mergo/compare/v0.3.13...v1.0.0

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -10,7 +10,7 @@ package config
 import (
 	"os"
 
-	"github.com/imdario/mergo"
+	"dario.cat/mergo"
 	"github.com/pelletier/go-toml"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


thaJeztah:
@changweige  @imeoer ptal 🤗

thaJeztah:
> The old dependency isn't fully gone yet, because kubernetes still uses the old location, but that may go away eventually; https://github.com/kubernetes/client-go/blob/v0.26.2/go.mod#L18

I opened a PR in kubernetes to switch it to the new name;

- https://github.com/kubernetes/kubernetes/pull/126351

apostasie:
@thaJeztah same as the other - if you could rebase to get a clean CI run, that would be lovely
cc @imeoer

PR #605, opened 2024-07-23
cleanup: regex not being evaluated in quoted expression

ryansavino:
Cleanup never happening for these directories.
Regex not evaluated in the rm statement.

PR #609, opened 2024-07-26
misc: remove related images/contents/snapshots in cleanup progress 

ChengyuZhu6:
The cleanup process exits before completion. In addition to snapshots, we also need to clean up related content and images.

Depends on Ryan's PR #605.

### DIFF
--- a/misc/snapshotter/snapshotter.sh
+++ b/misc/snapshotter/snapshotter.sh
@@ -192,36 +192,81 @@ function deploy_snapshotter() {
 
 }
 
+function remove_images() {
+    local SNAPSHOTTER="nydus"
+    local NAMESPACE="k8s.io"
+    local ctr_args="nsenter -t 1 -m ctr"
+
+    if [[ " k3s k3s-agent rke2-agent rke2-server " =~ " ${CONTAINER_RUNTIME} " ]]; then
+        ctr_args+=" --address /run/k3s/containerd/containerd.sock"
+    fi
+    ctr_args+=" --namespace $NAMESPACE"
+
+    # List all snapshots for nydus snapshotter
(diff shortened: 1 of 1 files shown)


ChengyuZhu6:
cc @ryansavino

ryansavino:
I fixed a few things in #605 per review comments. You may want to re-cherrypick the commits.

ChengyuZhu6:
> I fixed a few things in #605 per review comments. You may want to re-cherrypick the commits.

Done.

misc/snapshotter/snapshotter.sh
```diff
@@ -192,36 +192,81 @@ function deploy_snapshotter() {
...
     if [ ! -z "$pid" ]; then
-        local ctr_args=""
-        if [[ " k3s k3s-agent rke2-agent rke2-server " =~ " ${CONTAINER_RUNTIME} " ]]; then
-            ctr_args="--address /run/k3s/containerd/containerd.sock "
-        fi
-        ctr_args+="--namespace k8s.io snapshot --snapshotter nydus"
-        for i in $(nsenter -t 1 -m ctr ${ctr_args} list | grep -v KEY | cut -d' ' -f1); do
-            nsenter -t 1 -m ctr ${ctr_args} rm $i || true
-        done
+        remove_images
```
ryansavino, line 252: @ChengyuZhu6 wondering if this call to cleanup images even needs to be inside the pid check conditional statement. Why not just run it every time?

ryansavino (approved):

AdithyaKrishnan (approved):
LGTM

imeoer:
LGTM

misc/snapshotter/snapshotter.sh
ChengyuZhu6: If the snapshotter isn’t running, the `remove_images` function won’t have any effect.

PR #610, opened 2024-07-31
go.mod: github.com/containerd/containerd v1.7.20, switch to github.com/containerd/platforms, github.com/distribution/reference module

thaJeztah:
- follow-up to https://github.com/containerd/nydus-snapshotter/pull/599

### go.mod: github.com/containerd/containerd v1.7.20

- switches to separate github.com/containerd/containerd/api module
- switches to separate github.com/containerd/platforms module

full diff: https://github.com/containerd/containerd/compare/v1.7.18...v1.7.20

### switch to github.com/containerd/platforms module

containerd 1.7.19 and up alias the platforms package to the new module,
and deprecate the package.

### switch to github.com/distribution/reference module

containerd 1.7.19 and up alias the reference/docker package to the new module,
and deprecate the package.

PR #611, opened 2024-07-31
pkg/auth: remove CRI Alpha API

thaJeztah:
- relates to https://github.com/containerd/stargz-snapshotter/pull/1175

CRI v1alpha2 is deprecated since containerd 1.7:
https://github.com/containerd/containerd/blob/v1.7.0/RELEASES.md#deprecated-features

### DIFF
--- a/pkg/auth/image_proxy.go
+++ b/pkg/auth/image_proxy.go
@@ -14,10 +14,8 @@ import (
 	"github.com/containerd/containerd/defaults"
 	"github.com/containerd/containerd/pkg/dialer"
 	"github.com/containerd/containerd/reference"
-	runtime_alpha "github.com/containerd/containerd/third_party/k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 	"github.com/containerd/log"
 	"github.com/containerd/stargz-snapshotter/service/keychain/cri"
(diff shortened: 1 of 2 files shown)


thaJeztah:
@imeoer rebased, and I included some of the fixes from @apostasie (from https://github.com/containerd/nydus-snapshotter/pull/604) in this patch, adding them as co-author.


My goal with these sets of PRs was to see if we could have a release with the new modules in place (removing the ones deprecated in containerd) before this module moves to v2, so that I can update the version used in BuildKit in preparation of dependencies starting to move to containerd v2 (these module moves in containerd are intended to help with that transition to make sure a common module is used between v1 and v2 codebases).


:warning: Given that this PR removes some functionality, perhaps it warrants a v0.14.0 
❓ maybe tagging current main /  https://github.com/containerd/nydus-snapshotter/pull/610 as v0.13.15 ?

imeoer:
cc @changweige

thaJeztah:
FWIW: I opened a draft PR to test current main branch to verify it all works ok in BuildKit in case doing a v0.13.15 from current main is an option;

- https://github.com/moby/buildkit/pull/5205

imeoer:
LGTM!

PR #613, opened 2024-08-16
supervisor: fix large opaque data handle

imeoer:
```
When the supervisor needs to process a large amount of opaque data (for example,
a nydusd instance has a large number of sub mounts, and the opaque is usually
larger than 32kb), a recvmsg / sendmsg system call cannot process all data.

The commit fixes the problem with multiple recvmsg/sendmsg calls and updates
the test cases.
```

### DIFF
--- a/pkg/supervisor/supervisor.go
+++ b/pkg/supervisor/supervisor.go
@@ -8,6 +8,7 @@ package supervisor
 
 import (
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/supervisor/supervisor.go
```diff
@@ -25,45 +25,39 @@ import (
...
 func newMemStatesStorage() *MemStatesStorage {
-	return &MemStatesStorage{data: make([]byte, MaxOpaqueLen)}
+	return &MemStatesStorage{
+		data: []byte{},
+	}
 }
 
-func (mss *MemStatesStorage) Write(data []byte) {
-	l := copy(mss.data[mss.head:], data)
-	mss.head += l
```
changweige, line 57: Without `head`, States storage can't support append-write

imeoer: Thanks, we haven't needed append yet. :)

pkg/supervisor/supervisor.go
```diff
@@ -88,29 +84,97 @@ func (su *Supervisor) save(data []byte, fd int) {
...
 
-	if su.fd > 0 {
-		b := syscall.UnixRights(su.fd)
-		nOob = copy(oob, b)
+	data, err := su.dataStorage.Load()
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return data, su.fd, nil
```
changweige, line 104: Should we check if fd is positive?

imeoer: Should be ok, but maybe we don't have to limit it.

pkg/supervisor/supervisor.go
```diff
@@ -311,13 +325,19 @@ func (su *Supervisor) FetchDaemonStates(trigger func() error) error {
...
-	if err != nil {
+	eg := errgroup.Group{}
+	eg.Go(func() error {
+		err := trigger()
 		return errors.Wrapf(err, "trigger on %s", su.Sock())
-	}
+	})
+
+	eg.Go(func() error {
+		err := receiver()
```
changweige, line 335: Why `trigger` and `receiver` have to run in parallel in separted goroutines?

changweige (approved):

PR #614, opened 2024-09-01
tests/converter: modify unpack testcase

SToPire:
This patch modifies the unpack testcase to keep consistent with the new `nydus-image unpack` interface in nydus-image.

### DIFF
--- a/pkg/converter/tool/builder.go
+++ b/pkg/converter/tool/builder.go
@@ -305,7 +305,33 @@ func Unpack(option UnpackOption) error {
 	}
 
 	if option.BackendConfigPath != "" {
-		args = append(args, "--backend-config", option.BackendConfigPath)
+		configBytes, err := os.ReadFile(option.BackendConfigPath)
+		if err != nil {
+			return errors.Wrapf(err, "fail to read backend config file %s", option.BackendConfigPath)
+		}
+
+		var config map[string]interface{}
+		if err := json.Unmarshal(configBytes, &config); err != nil {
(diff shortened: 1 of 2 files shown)


SToPire:
The smoke test should work properly if [this](https://github.com/dragonflyoss/nydus/pull/1606) PR is applied.

pkg/converter/tool/builder.go
```diff
@@ -305,7 +305,41 @@ func Unpack(option UnpackOption) error {
 	}
 
 	if option.BackendConfigPath != "" {
-		args = append(args, "--backend-config", option.BackendConfigPath)
+		configFile, err := os.Open(option.BackendConfigPath)
```
imeoer, line 308: It seems the `configFile` not used.

pkg/converter/tool/builder.go
```diff
@@ -305,7 +305,41 @@ func Unpack(option UnpackOption) error {
...
+		if !ok {
+			return errors.New("failed to get backend config with type " + backendConfigType.(string))
+		}
+
+		backendConfigBytes, err := json.Marshal(backendConfig)
+		if err != nil {
+			return errors.Wrapf(err, "fail to marshal backend config %v", backendConfig)
+		}
+
+		logrus.Errorf("Backend config: %s", backendConfigBytes)
```
imeoer, line 339: The backend config maybe contains secret data, we shouldn't print it.

imeoer (commented):
Thanks! others LGTM, should we create a new builder (nydus-image) release?

PR #619, opened 2024-10-10
fix: update chunkmap suffix

Desiki-high:
Nydus latest chunk_map path has changed to `{blob_id}.blob.data.chunk_map`. Add `blobChunkMapSuffixedPath` to keep backward compatibility.


imeoer (approved):
LGTM, thanks!

PR #620, opened 2024-10-11
upgrade nydus version to v2.3.0 in test

imeoer:
nydus v2.3.0 released: https://github.com/dragonflyoss/nydus/releases/tag/v2.3.0


Desiki-high:
It looks like we need to update the chunkmap suffix in test:
https://github.com/containerd/nydus-snapshotter/blob/d6ca59c83a1ac05fcb8c555026a268249b60225d/tests/converter_test.go#L525-L528

PR #622, opened 2024-11-14
Dealing with optimizer-server zombie processes

coder-y04:
### Problem Description
Based on the process outlined in [optimize_nydus_image](https://github.com/containerd/nydus-snapshotter/blob/main/docs/optimize_nydus_image.md), we attempted to build an optimizer to generate the accessed files list. The optimizer-server and optimizer-nri-plugin were deployed on the server. After some time, we noticed that some of the optimizer-servers became defunct. The specific screenshots are as follows:
image
The parent processes of these zombie processes all belong to the optimizer-nri-plugin. We have also verified that this issue occurs on every server we deployed. 
We believe that if the servers continue to retain these zombie processes, they will accumulate to a certain number and eventually affect the normal operation of the business.
### Modified Sections
We checked and found that the issue primarily lies in two areas:
1. When the optimizer-nri-plugin performs the fanotifyServer operation, it needs to invoke the optimizer-server executable, which creates a child process. It is necessary to wait for the completion of this process and properly clean up the resources.
2. When the optimizer-server is running, it also creates a child process for the fanotify operation. Similarly, it is necessary to wait for this process to complete and clean up the resources.
### Validation Results
We deployed the modified functionality on four servers for observation. After a full day of monitoring, no zombie processes were observed.
### Testing Method
The modified code was recompiled, and the newly generated optimizer-server and optimizer-nri-plugin binaries replaced the existing ones.
The process status was checked using the command ps -ef | grep optimizer to verify that no zombie processes had appeared.

### DIFF
--- a/pkg/fanotify/fanotify.go
+++ b/pkg/fanotify/fanotify.go
@@ -79,6 +79,12 @@ func (fserver *Server) RunServer() error {
 	fserver.Cmd = cmd
 
 	go func() {
+		if err := cmd.Wait(); err != nil {
+			logrus.WithError(err).Errorf("Failed to wait for command to finish")
+		}
+	}()
+
+	go func() {
 		if err := fserver.RunReceiver(); err != nil {
 			logrus.WithError(err).Errorf("Failed to receive event information from server")
 		}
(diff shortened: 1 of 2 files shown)


coder-y04:
@imeoer  PTLK

imeoer (approved):
Sorry for the delay, LGTM, thanks!

PR #625, opened 2024-12-10
Revert "feat: Remove DumpFile operations"

imeoer:
This reverts commit 29243e35cccb595a8d3201aaad090430e7fd2946.

For the design of non-persistent configuration, we need to consider more factors
and cannot simply place the configuration files in the local metadata database.


Desiki-high (approved):
LGTM

PR #639, opened 2025-05-21
Fix converter to be able to push manifests to ECR

Fricounet:
This attempts to solve both https://github.com/dragonflyoss/nydus/issues/1691 and https://github.com/dragonflyoss/nydus/issues/1692.

Attempting to push some ECR manifests has 2 issues currently:
- `with-referrer` fails because the `subject` field in the manifest has a `platform` field. Although this is ok from an OCI spec perspective, ECR doesn't support it and it is not really needed so we can remove it
- `platform` fails when attempting to build multi-arch manifests because the `os.feature` field in the manifest is not supported either. We can remove it because this is not really supported yet in any container runtime. However, we still keep it when building images through `merge-platform` as it is the only way to differentiate a regular OCI image from a nydus one in that case. Fortunately, the merge-platform code goes through a different code path that uses Harbor's acceleration service and the os.feature field is set [there](https://github.com/goharbor/acceleration-service/blob/main/pkg/driver/nydus/nydus.go#L355).

I did test this by rebuilding nydusify using the fixed converter and tried to push both images to ECR, which worked.
Additionally, I made sure that the `merge-platform` field was still working:
```
./cmd/nydusify convert --source foo:bar --target-suffix -nydus-platform --merge-platform --platform "linux/amd64,linux/arm64"
INFO[2025-05-21T18:32:49+02:00] pulling image foo:bar  module=converter
INFO[2025-05-21T18:32:56+02:00] pulled image foo:bar , elapse 7.645556534s  module=converter
INFO[2025-05-21T18:32:56+02:00] converting image foo:bar  module=converter
INFO[2025-05-21T18:32:59+02:00] converted image foo:bar-nydus-platform , elapse 2.528599639s  module=converter
INFO[2025-05-21T18:32:59+02:00] pushing image foo:bar-nydus-platform  module=converter
INFO[2025-05-21T18:33:03+02:00] pushed image foo:bar-nydus-platform, elapse 4.147958056s  module=converter


crane manifest foo:bar-nydus-platform | jq
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d794231637e1cb6b40d30712718b82e47a306350295bc349dc38ac96fe0c6958",
      "size": 721,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:821c0238e5f7d2f95d90b897872f4584e56c9d89fd582c7eb2562583164768dd",
      "size": 721,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:8b32e06f97f9c3423b0eea97b18946d11ca3a1fc86df5ce1339c4634465b2d05",
      "size": 1588,
      "platform": {
        "architecture": "amd64",
        "os": "linux",
        "os.features": [
          "nydus.remoteimage.v1"
        ]
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:8c087d44e299989ea4a68329141bfb771883b4442ce1997ed5794e8936c258ef",
      "size": 1588,
      "platform": {
        "architecture": "arm64",
        "os": "linux",
        "os.features": [
          "nydus.remoteimage.v1"
        ]
      }
    }
  ]
}
```

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -1054,6 +1054,8 @@ func convertManifest(ctx context.Context, cs content.Store, oldDesc ocispec.Desc
 		// See the `subject` field description in
 		// https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions
 		manifest.Subject = &oldDesc
+		// Remove the platform field as it is not supported by certain registries like ECR.
+		manifest.Subject.Platform = nil
 	}
 
 	// Update image manifest in content store.
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -922,35 +922,17 @@ func ConvertHookFunc(opt MergeOption) converter.ConvertHookFunc {
 	}
 }
 
-// convertIndex modifies the original index by appending "nydus.remoteimage.v1"
-// to the Platform.OSFeatures of each modified manifest descriptors.
+// convertIndex modifies the original index converting it to manifest directly if it contains only one manifest.
 func convertIndex(ctx context.Context, cs content.Store, orgDesc ocispec.Descriptor, newDesc *ocispec.Descriptor) (*ocispec.Descriptor, error) {
 	var orgIndex ocispec.Index
 	if _, err := readJSON(ctx, cs, &orgIndex, orgDesc); err != nil {
 		return nil, errors.Wrap(err, "read target image index json")
-	}
-	// isManifestModified is a function to check whether the manifest is modified.
-	isManifestModified := func(manifest ocispec.Descriptor) bool {
-		for _, oldManifest := range orgIndex.Manifests {
-			if manifest.Digest == oldManifest.Digest {
-				return false
-			}
-		}
-		return true
 	}
 
 	var index ocispec.Index
 	indexLabels, err := readJSON(ctx, cs, &index, *newDesc)
 	if err != nil {
 		return nil, errors.Wrap(err, "read index json")
-	}
-	for i, manifest := range index.Manifests {
-		if !isManifestModified(manifest) {
-			// Skip the manifest which is not modified.
-			continue
-		}
-		manifest.Platform.OSFeatures = append(manifest.Platform.OSFeatures, ManifestOSFeatureNydus)
-		index.Manifests[i] = manifest
 	}
 
 	// If the converted manifest list contains only one manifest,


Fricounet:
If the PR is merged, a new release will have to be cut for the snapshotter in order to use it in nydusify right?

pkg/converter/convert_unix.go
```diff
@@ -922,36 +922,18 @@ func ConvertHookFunc(opt MergeOption) converter.ConvertHookFunc {
...
 	indexLabels, err := readJSON(ctx, cs, &index, *newDesc)
 	if err != nil {
 		return nil, errors.Wrap(err, "read index json")
 	}
-	for i, manifest := range index.Manifests {
-		if !isManifestModified(manifest) {
-			// Skip the manifest which is not modified.
-			continue
-		}
-		manifest.Platform.OSFeatures = append(manifest.Platform.OSFeatures, ManifestOSFeatureNydus)
```
BraveY, line 952: Why is it not sufficient to remove only the line `manifest.Platform.OSFeatures = append(manifest.Platform.OSFeatures, ManifestOSFeatureNydus)` ? I'm concerned that omitting `index.Manifests[i] = manifest` might introduce unexpected issues elsewhere.

BraveY:
> If the PR is merged, a new release will have to be cut for the snapshotter in order to use it in nydusify right?

Once the PR is merged, a new release of the snapshotter will indeed need to be published to use it in nydusify. For testing purposes, you can temporarily pin the commit hash directly in go.mod instead of referencing a tag.

pkg/converter/convert_unix.go
Fricounet: @BraveY just removing the line you mentioned would indeed work. I removed the whole block because functionally, it now becomes:
```
for i, manifest := range index.Manifests {
  index.Manifests[i] = manifest
}
```
which, in my opinion does not serve much purpose :sweat_smile: except complicating the code. If you really prefer that we keep it, I can add it back if you want. What do you prefer?

BraveY: > @BraveY just removing the line you mentioned would indeed work. I removed the whole block because functionally, it now becomes:
> 
> ```
> for i, manifest := range index.Manifests {
>   index.Manifests[i] = manifest
> }
> ```
> 
> which, in my opinion does not serve much purpose 😅 except complicating the code. If you really prefer that we keep it, I can add it back if you want. What do you prefer?

Appreciate the clarification! My mistake stemmed from assuming `index` was empty, but since readJSON already initializes it, the loop adds no value. Agreed—let’s remove the redundant code.

BraveY:
> > If the PR is merged, a new release will have to be cut for the snapshotter in order to use it in nydusify right?
> 
> Once the PR is merged, a new release of the snapshotter will indeed need to be published to use it in nydusify. For testing purposes, you can temporarily pin the commit hash directly in go.mod instead of referencing a tag.

For testing purposes, let’s proceed with pinning the snapshotter’s commit hash in go.mod (using replace directive) to validate the proposed changes in nydusify.
Once the tests pass and there are no further updates to this PR, we can proceed with merging it. Does this plan work for you?

pkg/converter/convert_unix.go
```diff
@@ -922,36 +922,18 @@ func ConvertHookFunc(opt MergeOption) converter.ConvertHookFunc {
 	}
 }
 
-// convertIndex modifies the original index by appending "nydus.remoteimage.v1"
-// to the Platform.OSFeatures of each modified manifest descriptors.
+// convertIndex modifies the original index converting it to manifest directly if it contains only one manifest.
 func convertIndex(ctx context.Context, cs content.Store, orgDesc ocispec.Descriptor, newDesc *ocispec.Descriptor) (*ocispec.Descriptor, error) {
 	var orgIndex ocispec.Index
```
imeoer, line 927: It seems we can remove the useless codes.

Fricounet: Very good point, I hadn't caught that. Fixed in 6542a89

Fricounet:
> For testing purposes, let’s proceed with pinning the snapshotter’s commit hash in go.mod (using replace directive) to validate the proposed changes in nydusify.
Once the tests pass and there are no further updates to this PR, we can proceed with merging it. Does this plan work for you?

Sure, I'm currently struggling in go version incompatibility in the go mod of nydusify (which is also present on the latest release of the snapshotter), but I'll try to get this done asap

imeoer (approved):
Thanks, LGTM!

imeoer:
Let's tag the next release as `v0.15.2`.

Issue #612, opened 2024-08-07
Proposal: Add Helm Chart for nydus snapshotter


changweige:
Nydus-snapshotter has no Helm Chart to be conventionally deployed in the K8s cluster yet. We can compose such a Helm Chart to pack all necessary binaries, configs and K8s manifests into a single Helm Chart package. Depoly nydus-snapshotter as Daemonset  running a each node. Run nydusd by `systemd-run` which makes it possible that nydusd runs in the host namespace while nydus-snapshotter runs in the its container namespace.  
Nydus-snapshotter container image packs：
1. nydus-snapshotter a.k.a `containerd-nydus-grpc`
2. nydusd
3. TOML configs

nydus-snapshotter pod's init-container installs nydusd binary to host's system path during pod startup, which means systemd can find it in the host namespace.

In such manner, even nydus-snapshotter pod is destroyed for the reason like Daemonset rolling upgrade or other maintenance purpose or unintentional operation, the container image IO is not affected.

imeoer:
https://github.com/dragonflyoss/helm-charts/tree/main/charts/nydus-snapshotter is it enough for this?

gane5hvarma:
@imeoer The above helm chart is not updated with latest version of nydus snapshotter. When updated to latest snapshotter version v0.15.0. it fails with the below error
```
Successfully pulled image "ghcr.io/containerd/nydus-snapshotter:v0.15.0" in 118ms (118ms including waiting). Image size: 64560834 bytes.
  Warning  Failed     10s               kubelet            Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/containerd/io.containerd.grpc.v1.cri/containers/nydus-snapshotter/volumes/34e0211ad3175a3621a689759309c8138927f6d053df6d181dc4270e8a5d1b84" to rootfs at "/var/lib/containerd/io.containerd.snapshotter.v1.nydus": possibly malicious path detected -- refusing to operate on /run/containerd/io.containerd.runtime.v2.task/k8s.io/nydus-snapshotter/rootfs/var/lib/containerd/io.containerd.snapshotter.v1.nydus (deleted): unknown
```

PR #643, opened 2025-06-05
Add helm charts for nydus-snapshotter for convenient K8s deploytment 

changweige:
Address #612 

Currently nydus-snapshotter does not have an elegant method to be deployed to a productive **Kubernetes** cluster. Nydus-snapshotter only release binary artifacts w/o consideration for Kubernetes. So this PR tries to pack nydus-snapshotter together with necessary nydus node components and its configurations into a Helm Chart. It makes the nydus-snapshotter installation and deployment convenient for users even if the users only intends to give a quick trial and evaluation if nydus suites their business. It's totally a standard K8s fashion. The users don't have to scan tons of documents. Yes it is just packed as a whole.

The challenge  we have to solve first is that K8s applications/components(on top of DaemonSet) can be killed, rollout to upgrade, the progress should not affect any running nydusd as the user space fuse service. So we cant let nydus-snapshotter directly spawn a nydusd which is also running in the same Linux namespaces and cgroups with nydus-snapshotter. By design, nydus-snapshotter is a privileged container with host PID namespace, this means nydus-snapshotter can `nsenter` the host namespace and delegate nydusd creation to  `systemd` by its utility `systemd-run`. Nydus-snapshotter will enter host  mnt namespace and calls `systemd-run` to start nydusd as a transient service. So deleting nydus-snapshotter  daemonset and helm release, rollout daemonset, killing nydus-snapshotter won't affect running nydusd, which means user's business is continuous across those processes.

As for the details of the deployment architecture:
DaemonSet:
- nydus-snapshotter
- nydus node components: nydusd, nydus-image, nydusctl
ConfigMap:
- nydus-snapshotter config.toml

Besides, nydus-snapshotter will modify containerd config.toml to integrate with nydus-snapshotter and restart containerd if necessary.

To test and try, just at the root of the code project and perform
```
helm install nydus-snapshotter charts
```

### DIFF
--- a/charts/Chart.yaml
+++ b/charts/Chart.yaml
@@ -1,0 +1,15 @@
+apiVersion: v2
+name: nydus-snapshotter
+description: A Helm chart for nydus-snapshotter
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: v0.15.2
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: v0.15.2
--- a/charts/templates/cluster-role-binding.yaml
+++ b/charts/templates/cluster-role-binding.yaml
@@ -1,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: nydus-snapshotter-cluster-role-binding
+subjects:
+  - kind: ServiceAccount
+    name: nydus-snapshotter-sa
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: nydus-snapshotter-cluster-role
+  apiGroup: rbac.authorization.k8s.io
--- a/charts/templates/cluster-role.yaml
+++ b/charts/templates/cluster-role.yaml
@@ -1,0 +1,5 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: nydus-snapshotter-cluster-role
+rules: 
--- a/charts/templates/daemonset.yaml
+++ b/charts/templates/daemonset.yaml
@@ -1,0 +1,108 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: nydus-snapshotter
+  namespace: {{ .Release.Namespace }}
+spec:
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 1
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: nydus-snapshotter
+      app.kubernetes.io/name: nydus-snapshotter
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/instance: nydus-snapshotter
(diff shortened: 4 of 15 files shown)


lukasmrtvy:
@changweige can we make the nydus-snapshotter-config optional?  or make it available as a string in values.yaml

for example ( values.yaml ):
```
config: |
    version = 1
    # Snapshotter's own home directory where it stores and creates necessary resources
    root = "/var/lib/containerd/io.containerd.snapshotter.v1.nydus"
    # The snapshotter's GRPC server socket, containerd will connect to plugin on this socket
    address = "/run/containerd-nydus/containerd-nydus-grpc.sock"
    # The nydus daemon mode can be one of the following options: multiple, dedicated, shared, or none. 
    # If `daemon_mode` option is not specified, the default value is multiple.
    daemon_mode = "dedicated"
```
or 
```
config: ""
configFrom: 
```

changweige:
> @changweige can we make the nydus-snapshotter-config optional? or make it available as a string in values.yaml
> 
> for example ( values.yaml ):
> 
> ```
> config: |
>     version = 1
>     # Snapshotter's own home directory where it stores and creates necessary resources
>     root = "/var/lib/containerd/io.containerd.snapshotter.v1.nydus"
>     # The snapshotter's GRPC server socket, containerd will connect to plugin on this socket
>     address = "/run/containerd-nydus/containerd-nydus-grpc.sock"
>     # The nydus daemon mode can be one of the following options: multiple, dedicated, shared, or none. 
>     # If `daemon_mode` option is not specified, the default value is multiple.
>     daemon_mode = "dedicated"
>     ...
> ```
> 
> or
> 
> ```
> config: ""
> configFrom: "external-configmap"
> configFromKey: "myconfig.toml"
> ```
> 
> Thanks

Looks reasonable for users who has their own nydus-snapshotter config. I can improve the PR like still keep the configmap template as part of the Helm Chart, but we add a value to the `value.yaml to take the custom nydus-snapshotter. If `config` in the `values.yaml` is not set, the default nydus-snapshotter config will be mounted to each K8s node. In this way, unexperienced user can quickly have a sane nydus system running in the K8s cluster.

Does it looks good to you?

changweige:
@lukasmrtvy 

> Yes, default `config` should be defined in values.yaml to get started with, and if `configFrom` is defined, use it instead.
> 
> I guess the ultimate solution would be to merge TOML with tool like https://github.com/TomWright/dasel, this way one could edit single property like `address`

Good idead. I am not sure Helm gives us a chance to merge those two when rendering templates, however, at least, I can merge  both the default nydus-snapshotter toml config and user's (maybe a set of override values) when performing `setup.sh` before exec `containerd-nydus-grpc`. In fact, `dasel` has been utilized in the `setup.sh` to modify containerd's config.toml.

> 
> ```
> # will be merged with default config
> overrideConfig: |
>     address = "my/path/ontainerd-nydus/containerd-nydus-grpc.sock"
> ```
> 
> I also noticed the RBAC is missing, for example this https://github.com/containerd/nydus-snapshotter/blob/main/misc/snapshotter/nydus-snapshotter-rbac.yaml could improve the experience with adding the additional node label after nydus installation.
> 

Yes, this PR does not contain any role definition. What kind of node labels do you think is helpful? 

> WDYT?

changweige:
@lukasmrtvy Right now, the nydus-snapshotter's configuration file can be specified in the values.yaml while a default configuration toml content is defined at configmap Helm template. If it is specified in the values.yaml, it will take effect to running nydus-snapshotter

imeoer:
@fidencio I agree to migrate nydus-snapshotter chart from https://github.com/dragonflyoss/helm-charts/tree/main/charts/nydus-snapshotter into this improved PR, @changweige If there's a basic CI test for the PR, we can merge it.

PR #651, opened 2025-07-24
[Optimize] improve the stability of hot upgrade daemon process

Zephyrcf:
For daemons that have been hot upgraded, subsequent rmi mirroring cannot clear the daemons.
This pr optimizes this.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -604,6 +604,11 @@ func (d *Daemon) ClearVestige() {
 func (d *Daemon) CloneRafsInstances(src *Daemon) {
 	instances := src.RafsCache.List()
 	d.RafsCache.SetIntances(instances)
+	ref := src.GetRef()
+	for ref > 0 {
+		d.IncRef()
+		ref--
+	}
 }
 
 // Daemon must be started and reach RUNNING state before call this method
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -604,6 +604,11 @@ func (d *Daemon) ClearVestige() {
 func (d *Daemon) CloneRafsInstances(src *Daemon) {
 	instances := src.RafsCache.List()
 	d.RafsCache.SetIntances(instances)
+	ref := src.GetRef()
+	for ref > 0 {
+		d.IncRef()
+		ref--
```
imeoer, line 610: Does this work here?

imeoer:
Thanks, plz squash the commits. :)

pkg/system/system.go
```diff
@@ -384,15 +384,20 @@ func (sc *Controller) upgradeNydusDaemon(d *daemon.Daemon, c upgradeRequest, man
 		return err
 	}
 
-	su := manager.SupervisorSet.GetSupervisor(d.ID())
-	if err := su.SendStatesTimeout(time.Second * 10); err != nil {
-		return errors.Wrap(err, "Send states")
+	if d.Supervisor != nil {
```
imeoer: Maybe change to:

```
if d.Supervisor == nil {
  return errors.New("should set recover policy to failover to enable hot upgrade")
}
```

And put it at the beginning of the func.

pkg/daemon/daemon.go
Zephyrcf: After many regression tests, the modification here is effective.

pkg/system/system.go
Zephyrcf: done.

imeoer (approved):
LGTM, thanks!

PR #652, opened 2025-07-25
[Fix] proper handling of nydusd binary copies in hot upgrades

Zephyrcf:
The use of os.Rename (which has mv semantics) would cause the file at the original source path (newNydusdPath) to no longer exist after the hot upgrade operation.

### DIFF
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -345,14 +345,18 @@ func (sc *Controller) upgradeDaemons() func(w http.ResponseWriter, r *http.Reque
 				}
 			}
 
-			// TODO: why renaming?
-			err = os.Rename(c.NydusdPath, manager.NydusdBinaryPath)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/system/system.go
```diff
@@ -345,10 +346,11 @@ func (sc *Controller) upgradeDaemons() func(w http.ResponseWriter, r *http.Reque
 				}
 			}
 
-			// TODO: why renaming?
-			err = os.Rename(c.NydusdPath, manager.NydusdBinaryPath)
+			cmd := exec.Command("sudo", "install", "-D", "-m", "755", c.NydusdPath, manager.NydusdBinaryPath)
```
imeoer: Thanks for the fixup! Can we use `io.Copy` and `os.Chmod` instead of shell command here? Not all envs have sudo and install commands.

Zephyrcf: done.

Issue #655, opened 2025-08-07
Nydusd fails to start when snapshotter root path leads to an oversized Unix socket path


Zephyrcf:
### Summary

When the path to the nydus-snapshotter root directory is too long, the nydusd daemon will fail to start due to the wrong path length.

### Logs
nerdctl Error:
`FATA[0004] wait for instance id x: wait until daemon is RUNNING: get daemon state: daemon socket ${rootPath}/socket/${xid}/api.sock: not found`

nydusd Error:
`Error: Error { kind: InvalidInput, message: "path must be shorter than SUN_LEN" }`
`time="xxxxxxx" level=error msg="Process 190090 has been a zombie"`
`time="xxxxxxx" level=error msg="Nydusd d2a7fb7lnnuukgvqigt0 probably not started"`
`time="xxxxxxx" level=error msg="Process 190090 has been a zombie"`

The daemon failed to start because the combined Unix socket path exceeds the system's SUN_LEN limit.

I think we should do a check for possible unix path lengths when nydus-snapshotter starts, it would be bad if it fails when running the nydusd daemon

imeoer:
Hi @Zephyrcf, thanks for the issue! What are `rootPath` and `xid` like?

Zephyrcf:
> Hi [@Zephyrcf](https://github.com/Zephyrcf), thanks for the issue! What are `rootPath` and `xid` like?

rootPath is the root parameter configured in nydus-snapshotter config. xid is randomly generated through *xid.New()* and is used to identify a nydusd daemon.

PR #656, opened 2025-08-07
[Fix] refine Unix socket path length validation

Zephyrcf:
close https://github.com/containerd/nydus-snapshotter/issues/655

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -47,6 +47,16 @@ const (
 	// to manage those snapshots.
 	DaemonModeNone    DaemonMode = DaemonMode(constant.DaemonModeNone)
 	DaemonModeInvalid DaemonMode = DaemonMode(constant.DaemonModeInvalid)
+	// MaxRootPathLen defines the maximum allowed length of the root portion of a Unix domain socket path.
+	//
+	// This value is calculated based on the hard length limit of `sun_path` on Linux systems (108 bytes).
+	// Nydusd's socket path format is "${rootPath}/socket/${xid}/api?.sock".
+	// - The length of the fixed part "/socket/${xid}/api?.sock" is 38 bytes.
+	//
+	// Since the maximum upper limit of the total path length is 108 bytes, in order to avoid exceeding the limit, the maximum allowed length of rootPath is:
+	// 108 - len("/socket/${xid}/api?.sock") = 108 - 38 = 70.
+	// Therefore, we must set the effective maximum length of the root path to 70 bytes.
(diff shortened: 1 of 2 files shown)


Zephyrcf:
For reference on the `108`bytes `sun_path` hard limit on Linux, Nydus itself relies on the `libc v0.2` crate's definition. Here is the relevant documentation:

https://docs.rs/libc/0.2.169/libc/struct.sockaddr_un.html

config/config.go
```diff
@@ -274,9 +284,13 @@ func ValidateConfig(c *SnapshotterConfig) error {
 		}
 	}
 
-	if len(c.Root) == 0 {
+	rootPathLen := len(c.Root)
+	if rootPathLen == 0 {
 		return errors.New("empty root directory")
 	}
+	if rootPathLen > MaxRootPathLen {
```
BraveY, line 291: Maybe we could add some unit tests for this simple function. Others LGTM!

mfrankm: Hello, when will the version be released? Thanks@Zephyrcf @BraveY

Zephyrcf: > Maybe we could add some unit tests for this simple function. Others LGTM!

done.

Zephyrcf: > Hello, when will the version be released? Thanks@Zephyrcf @BraveY

We will be releasing a new version this month, so stay tuned.

BraveY:
@imeoer cc

config/config.go
```diff
@@ -47,6 +47,16 @@ const (
...
+	// MaxRootPathLen defines the maximum allowed length of the root portion of a Unix domain socket path.
+	//
+	// This value is calculated based on the hard length limit of `sun_path` on Linux systems (108 bytes).
+	// Nydusd's socket path format is "${rootPath}/socket/${xid}/api?.sock".
+	// - The length of the fixed part "/socket/${xid}/api?.sock" is 38 bytes.
+	//
+	// Since the maximum upper limit of the total path length is 108 bytes, in order to avoid exceeding the limit, the maximum allowed length of rootPath is:
+	// 108 - len("/socket/${xid}/api?.sock") = 108 - 38 = 70.
+	// Therefore, we must set the effective maximum length of the root path to 70 bytes.
+	MaxRootPathLen = 70
```
imeoer, line 59: Thanks for the PR! Perhaps a better approach would be to bypass this limitation using a symlink instead of throwing an error, but this check is reasonable for now.

PR #663, opened 2025-08-16
[Feature] Support nydusd failover-policy config option

lsc2001:
Add a new option `failover-policy` in config.toml and support passing nydusd failover-policy
command line options when creating nydusd.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -107,6 +107,12 @@ const (
 	FsDriverProxy    string = constant.FsDriverProxy
 )
 
+const (
+	FailoverPolicyNone   string = constant.FailoverPolicyNone
+	FailoverPolicyResend string = constant.FailoverPolicyResend
+	FailoverPolicyFlush  string = constant.FailoverPolicyFlush
+)
+
 type Experimental struct {
 	EnableStargz         bool        `toml:"enable_stargz"`
 	EnableReferrerDetect bool        `toml:"enable_referrer_detect"`
@@ -136,6 +142,7 @@ type DaemonConfig struct {
 	FsDriver         string `toml:"fs_driver"`
 	ThreadsNumber    int    `toml:"threads_number"`
 	LogRotationSize  int    `toml:"log_rotation_size"`
+	FailoverPolicy   string `toml:"failover_policy"`
 }
 
 type LoggingConfig struct {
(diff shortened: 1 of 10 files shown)


pkg/manager/daemon_adaptor.go
```diff
@@ -200,6 +200,8 @@ func (m *Manager) BuildDaemonCommand(d *daemon.Daemon, bin string, upgrade bool)
 		cmdOpts = append(cmdOpts, command.WithLogFile(d.LogFile()))
 	}
 
+	cmdOpts = append(cmdOpts, command.WithFailoverPolicy(d.States.FailoverPolicy))
```
Zephyrcf, line 203:
```suggestion
	if m.RecoverPolicy == config.RecoverPolicyFailover {
		cmdOpts = append(cmdOpts, command.WithFailoverPolicy(d.States.FailoverPolicy))
	}
```
This parameter only needs to be passed when RecoverPolicy is Failover.

config/config.go
```diff
@@ -289,6 +296,11 @@ func ValidateConfig(c *SnapshotterConfig) error {
 	if c.DaemonConfig.ThreadsNumber > 1024 {
 		return errors.Errorf("nydusd worker thread number %d is too big, max 1024", c.DaemonConfig.ThreadsNumber)
 	}
+	if c.DaemonConfig.FailoverPolicy != FailoverPolicyNone &&
+		c.DaemonConfig.FailoverPolicy != FailoverPolicyResend &&
+		c.DaemonConfig.FailoverPolicy != FailoverPolicyFlush {
+		return errors.Errorf("invalid failover policy %q", c.DaemonConfig.FailoverPolicy)
+	}
```
Zephyrcf, line 317: Thanks for the PR! If the RecoverPolicy is not Failover, but the FailoverPolicy is resend, this will not actually take effect. Can we make a reminder here?

internal/constant/values.go
```diff
@@ -53,3 +53,10 @@ const (
 	DefaultRotateLogLocalTime     = true
 	DefaultRotateLogCompress      = true
 )
+
+const (
+	FailoverPolicyNone    string = "none"
+	FailoverPolicyResend  string = "resend"
+	FailoverPolicyFlush   string = "flush"
+	DefaultFailoverPolicy string = FailoverPolicyNone
```
imeoer: May be make `resend` as default.

misc/snapshotter/config.toml
```diff
@@ -37,6 +37,8 @@ recover_policy = "restart"
 threads_number = 4
 # Log rotation size for nydusd, in unit MB(megabytes). (default 100MB)
 log_rotation_size = 100
+# Nydusd failover policy, can be "none", "resend" or "flush"
+failover_policy = "none"
```
imeoer: Ditto.

pkg/manager/daemon_adaptor.go
imeoer: @lsc2001 Let's also change recover_policy default value to `failover` ?

PR #664, opened 2025-08-16
feat(merge): add blob.meta support in merge logic (v0.13.11)

sf1999817:
### Background
This patch optimizes the merge process for RAFS v6 by including the `blob.meta` file in the bootstrap tar archive during image build.

### Change Base
The modification is based on the nydus-snapshotter v0.13.11 release tag. In my deployment, the nydusify tool uses this specific version to convert and load container images.

### Optimization
The update modifies the merge logic to package the `blob.meta` file together with the bootstrap data, ensuring full metadata is available in the output tar archive.

### Result
This improves metadata handling for RAFS v6 images and simplifies loading for deployments based on v0.13.11.

Bootstrap tar with blob.meta (RAFS v6)


pkg/converter/utils.go
```diff
@@ -83,39 +88,27 @@ func newSeekReader(ra io.ReaderAt) *seekReader {
...
 
-// packToTar makes .tar(.gz) stream of file named `name` and return reader.
-func packToTar(src string, name string, compress bool) (io.ReadCloser, error) {
-	fi, err := os.Stat(src)
-	if err != nil {
-		return nil, err
-	}
-
+// packToTar packs files to .tar(.gz) stream then return reader.
+func packToTar(files []File, compress bool) io.ReadCloser {
```
BraveY, line 92: It would be better to return the error. So we could hander the error.

PR #665, opened 2025-08-20
[Feature] make metrics intervals configurable

Zephyrcf:
Summary:
This PR introduces configurable hung_io_interval and collect_interval parameters in the MetricsConfig section. Previously, these were hard-coded, limiting user control over metrics collection behavior.

Key Changes:
- MetricsConfig now includes HungIOInterval and CollectInterval.
- Fill in default values if the fields are not specified in the configuration.
- Implemented a robust parsing logic with a new helper function (ParseDurationWithDefault) to handle user-provided values.
- Updated TOML configuration examples.

Testing:
Verified with unit tests for the parsing helper and integration tests for the configuration loading logic.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -9,6 +9,7 @@ package config
 
 import (
 	"os"
+	"time"
 
 	"dario.cat/mergo"
 	"github.com/pelletier/go-toml"
@@ -186,8 +187,8 @@ type CacheManagerConfig struct {
 	Disable bool `toml:"disable"`
 	// Trigger GC gc_period after the specified period.
 	// Example format: 24h, 120min
-	GCPeriod string `toml:"gc_period"`
-	CacheDir string `toml:"cache_dir"`
+	GCPeriod time.Duration `toml:"gc_period"`
+	CacheDir string        `toml:"cache_dir"`
 }
 
 // Configure how nydus-snapshotter receive auth information
@@ -213,7 +214,12 @@ type MirrorsConfig struct {
 }
 
 type MetricsConfig struct {
+	// Address is the network address for the metrics server. Empty indicates the server is disabled.
 	Address string `toml:"address"`
+	// HungIOInterval defines the timeout for a single I/O operation to be considered "hung".
+	HungIOInterval time.Duration `toml:"hung_io_interval"`
+	// CollectInterval defines how often metrics are collected and reported.
+	CollectInterval time.Duration `toml:"collect_interval"`
 }
 
 type DebugConfig struct {
--- a/config/global.go
+++ b/config/global.go
@@ -12,13 +12,11 @@ package config
 import (
 	"os"
 	"path/filepath"
-	"time"
 
 	"github.com/containerd/log"
-	"github.com/pkg/errors"
-
 	"github.com/containerd/nydus-snapshotter/internal/logging"
 	"github.com/containerd/nydus-snapshotter/pkg/utils/mount"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -36,7 +34,6 @@ type GlobalConfig struct {
 	ConfigRoot       string
 	RootMountpoint   string
 	DaemonThreadsNum int
-	CacheGCPeriod    time.Duration
 	MirrorsConfig    MirrorsConfig
 }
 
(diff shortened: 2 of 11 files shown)


config/config.go
```diff
@@ -196,7 +196,12 @@ type MirrorsConfig struct {
 }
 
 type MetricsConfig struct {
+	// Address is the network address for the metrics server. Empty indicates the server is disabled.
 	Address string `toml:"address"`
+	// HungIOInterval defines the timeout for a single I/O operation to be considered "hung".
+	HungIOInterval string `toml:"hung_io_interval"`
+	// CollectInterval defines how often metrics are collected and reported.
+	CollectInterval string `toml:"collect_interval"`
```
Fricounet: why not use time.Duration type directly? As the TOML package can handle the parsing

internal/constant/values.go
```diff
@@ -52,4 +54,15 @@ const (
...
+	// metrics configuration
+	DefaultHungIOInterval  string = "30s"
+	DefaultCollectInterval string = "1m"
+)
+
+const (
+	// Time Constants
+	TwentyFourHoursDuration = 24 * time.Hour
+	OneMinuteDuration       = 1 * time.Minute
+	ThirtySecondsDuration   = 30 * time.Second
```
Fricounet: wouldn't it be simpler to use the duration directly?
i feel like that 24 * time.Hour is more easy to understand than reading `TwentyFourHoursDuration`

pkg/metrics/serve.go
```diff
@@ -41,7 +38,7 @@ func WithProcessManagers(managers []*manager.Manager) ServerOpt {
 	}
 }
 
-func NewServer(ctx context.Context, opts ...ServerOpt) (*Server, error) {
+func NewServer(ctx context.Context, hungIOInterval time.Duration, opts ...ServerOpt) (*Server, error) {
```
Fricounet: how about passing the info using a WithOption pattern (like WithProcessManager above)? This way you don't change the function's signature. And the value becomes available to all the methods of Server (like StartCollectMetrics below)

config/global.go
```diff
@@ -76,6 +80,14 @@ func GetCacheGCPeriod() time.Duration {
...
 }
 
+func GetMetricsHungIOInterval() time.Duration {
+	return globalConfig.MetricsHungIOInterval
+}
+
+func GetMetricsCollectInterval() time.Duration {
+	return globalConfig.MetricsCollectInterval
+}
+
```
Fricounet: do you really need the global configs? The only place using `GetMetrics*Interval` is in pkg/metrics/serve.go which already has accessed to the parsed config

internal/constant/values.go
```diff
@@ -52,6 +54,10 @@ const (
 	DefaultRotateLogMaxAge        = 0 // days
 	DefaultRotateLogLocalTime     = true
 	DefaultRotateLogCompress      = true
+
+	// metrics configuration
+	DefaultHungIOInterval  = 30 * time.Second
```
Fricounet: is it expected that you are bumping the hungIOInterval from the original 10s to 30s?

pkg/utils/parser/parser.go
```diff
@@ -75,3 +77,17 @@ func MemoryConfigToBytes(data string, totalMemoryBytes int) (int64, error) {
 	multiplier := unitMultipliers[unit]
 	return int64(value * float64(multiplier)), nil
 }
+
+func ParseDurationWithDefault(
```
Fricounet: I think this function is unneeded now, no?

pkg/metrics/serve.go
```diff
@@ -22,16 +23,15 @@ import (
...
 type Server struct {
 	managers          []*manager.Manager
 	snCollectors      []*collector.SnapshotterMetricsCollector
 	fsCollector       *collector.FsMetricsVecCollector
 	inflightCollector *collector.InflightMetricsVecCollector
+	hungIOInterval    time.Duration
+	collectInterval   time.Duration
 }
 
 func WithProcessManagers(managers []*manager.Manager) ServerOpt {
```
BraveY, line 37: tiny nits:
Consider placing the definitions of WithCollectInterval and WithHungIOInterval right after WithProcessManagers to keep related option functions grouped together.

BraveY (commented):
others LGTM

pkg/metrics/serve.go
Zephyrcf: Done.

imeoer (approved):
LGTM, thanks!

PR #666, opened 2025-08-20
feat(merge): add blob.meta support in merge logic (v0.13.11)

sf1999817:
### Background
This patch optimizes the merge process for RAFS v6 by including the blob.meta file in the bootstrap tar archive during image build.

### Change Base
The modification is based on the nydus-snapshotter v0.13.11 release tag. In my deployment, the nydusify tool uses this specific version to convert and load container images.

### Optimization
The update modifies the merge logic to package the blob.meta file together with the bootstrap data, ensuring full metadata is available in the output tar archive.

### Result
This improves metadata handling for RAFS v6 images and simplifies loading for deployments based on v0.13.11.

Bootstrap tar with blob.meta (RAFS v6)
merge

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -25,6 +25,7 @@ import (
 	"github.com/containerd/containerd/archive"
 	"github.com/containerd/containerd/archive/compression"
 	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/content/local"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/images"
 	"github.com/containerd/containerd/images/converter"
@@ -565,6 +566,11 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
 		return filepath.Join(workDir, digestHex)
 	}
 
+	getBlobMetaPath := func(layerIdx int, suffix string) string {
+		digestHex := layers[layerIdx].Digest.Hex()
+		return filepath.Join(workDir, digestHex+suffix)
+	}
+
 	eg, _ := errgroup.WithContext(ctx)
 	sourceBootstrapPaths := []string{}
 	rafsBlobDigests := []string{}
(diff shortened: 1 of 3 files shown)


pkg/converter/convert_unix.go
```diff
@@ -594,6 +600,16 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
 					return errors.Wrap(err, "unpack nydus tar")
 				}
 
+				blobMetaPath := getBlobMetaPath(idx, ".blob.meta")
+				blobMetaFile, err := os.Create(blobMetaPath)
+				if err != nil {
+					return errors.Wrap(err, "create blob.meta file")
+				}
+
```
Zephyrcf, line 608:
```suggestion
				if err != nil {
					return errors.Wrap(err, "create blob.meta file")
				}
				defer blobMetaFile.Close()
```
Ensure the file handle is closed.

pkg/converter/utils.go
```diff
@@ -83,39 +88,27 @@ func newSeekReader(ra io.ReaderAt) *seekReader {
...
 
-// packToTar makes .tar(.gz) stream of file named `name` and return reader.
-func packToTar(src string, name string, compress bool) (io.ReadCloser, error) {
-	fi, err := os.Stat(src)
-	if err != nil {
-		return nil, err
-	}
-
+// packToTar packs files to .tar(.gz) stream then return reader.
+func packToTar(files []File, compress bool) io.ReadCloser {
```
Zephyrcf, line 92: Propagate errors to the caller instead of handling them internally.

pkg/converter/utils.go
```diff
@@ -83,39 +88,27 @@ func newSeekReader(ra io.ReaderAt) *seekReader {
...
 	dirHdr := &tar.Header{
-		Name:     filepath.Dir(name),
+		Name:     "image",
```
Zephyrcf, line 94: Define "image" as a constant.

pkg/converter/convert_unix.go
```diff
@@ -594,6 +600,22 @@ func Merge(ctx context.Context, layers []Layer, dest io.Writer, opt MergeOption)
...
+       			// 使用修复的逆向搜索算法
+        		if err := UnpackEntryFixed(layers[idx].ReaderAt, EntryBlobMeta, blobMetaFile); err != nil {
```
BraveY: Why not use the existing UnpackEntry function directly?

pkg/converter/convert_unix.go
```diff
@@ -290,18 +291,44 @@ func UnpackEntry(ra content.ReaderAt, targetName string, target io.Writer) (*TOC
...
+	}
+	
+	if !found {
+		return nil, errors.Wrapf(ErrNotFound, "target %s not found", targetName)
+	}
+	
+	// For tar header found files, return minimal TOCEntry
+	var nameBytes [16]byte
+	copy(nameBytes[:], targetName)
+	return &TOCEntry{Name: nameBytes}, nil
```
Zephyrcf:
```suggestion
	return &TOCEntry{Name: nameBytes}, handleErr
```

PR #456, opened 2023-04-24
cmd: improve user interface for help message by add default text

adamqqqplay:
- pkg: introduce internal/constant package 
- cmd: improve user interface for help message by add default text

Current help message:
``` 
NAME:
   containerd-nydus-grpc - Nydus remote snapshotter for containerd

USAGE:
   containerd-nydus-grpc [global options] command [command options] [arguments...]

COMMANDS:
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --root value                                directory to store snapshotter data and working states (default: /var/lib/containerd-nydus)
   --address value                             remote snapshotter gRPC socket path (default: /run/containerd-nydus/containerd-nydus-grpc.sock)
   --config value                              path to nydus-snapshotter configuration (such as: config.toml)
   --nydus-image nydus-image                   path to nydus-image binary, default to search in $PATH (such as: /usr/local/bin/nydus-image)
   --nydusd nydusd                             path to nydusd binary, default to search in $PATH (such as: /usr/local/bin/nydusd)
   --nydusd-config value, --config-path value  path to nydusd configuration (such as: nydusd-config.json or nydusd-config-v2.toml) (default: /etc/nydus/nydusd-config.json)
   --daemon-mode value                         nydusd daemon working mode, possible values: "multiple", "shared" or "none" (default: multiple)
   --fs-driver value                           driver to mount RAFS filesystem, possible values: "fusedev", "fscache" (default: fusedev)
   --log-level value                           logging level, possible values: "trace", "debug", "info", "warn", "error" (default: info)
   --log-to-stdout                             print log messages to standard output (default: false)
   --version                                   print version and build information (default: false)
   --help, -h                                  show help
```

PR #669, opened 2025-08-28
Add OCI index detection for Nydus alternatives

Fricounet:
## Overview
_Please briefly describe the changes your pull request makes._

Implement index detection feature that automatically discovers Nydus alternative manifests within OCI index manifests, similar to what SOCI snashotter is able to do. This enables transparent fallback to optimized Nydus images when available while keeping the original index manifest OCI compliant as non-nydus client can simply pull the regular OCI image. The feature is heavily based on the already existing ReferrerDetect feature which looks for nydus manifests using the referrer API. However, the referrer API is not supported by all registries and as it relies on changes happening outside the manifest being pulled it's a bit dangerous from a supply chain perspective. On the other hand, this new IndexDetect feature is supported everywhere and does not have the supply chain issue as everyting is packed in a single manifest. Building a manifest compatible with this feature can be easily done with the `--merge-platform` flag in `nydusify convert`.

## Related Issues
_Please link to the relevant issue. For example: `Fix #123` or `Related #456`._

Comes from discussions we had in https://github.com/dragonflyoss/nydus/issues/1692#issuecomment-3144921344. Additionally, this will help with https://github.com/dragonflyoss/nydus/issues/463 since it will be easier to have OCI-compliant nydus images

## Change Details

Key changes:
- Add EnableIndexDetect config option to experimental features
- Implement index package with manifest finding logic. Very similar to referrer package. The detection is based on the `platform.os.feature` (nydus.remoteimage.v1) or the `artifactType` (application/vnd.nydus.image.manifest.v1+json) field in the index manifest
- Result is cached to avoid multiple API calls for the same digest
- Integrate index detection into filesystem and snapshot layers. The index detection is done before the referrer detection: if both exist, we prefer to use the index detection for the supply chain issues mentionned above.
- Add e2e test for index detection
- Add documentation for the new feature

## Test Results

Added an e2e test and some unit tests to cover this new feature.

I have not added a test in the `integration/` directory mainly because there is no image built with `merge-platform` in `ghcr.io/dragonflyoss/image-service/wordpress`. If one is added, I can add an integration test as well.

## Change Type
- [ ] Bug Fix
- [x] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/.github/workflows/k8s-e2e.yml
+++ b/.github/workflows/k8s-e2e.yml
@@ -6,6 +6,9 @@ on:
       auth-type:
         required: true
         type: string
+      index-detect:
+        required: false
+        type: boolean
 
 env:
   DOCKER_USER: testuser
@@ -27,8 +30,10 @@ jobs:
           go-version-file: 'go.mod'
           cache-dependency-path: "go.sum"
       - name: Test
+        env:
+          AUTH_TYPE: ${{ inputs.auth-type }}
+          INDEX_DETECT: ${{ inputs.index-detect }}
         run: |
-          AUTH_TYPE='${{ inputs.auth-type }}'
           ./tests/helpers/kind.sh
       - name: Dump logs
         if: failure()
--- a/.github/workflows/k8s-e2e-run.yml
+++ b/.github/workflows/k8s-e2e-run.yml
@@ -19,3 +19,9 @@ jobs:
     uses: ./.github/workflows/k8s-e2e.yml
     with:
       auth-type: kubeconf
+
+  index_detect:
+    uses: ./.github/workflows/k8s-e2e.yml
+    with:
+      auth-type: kubeconf
+      index-detect: true
--- a/config/config.go
+++ b/config/config.go
@@ -127,6 +127,7 @@ const (
 type Experimental struct {
 	EnableStargz         bool        `toml:"enable_stargz"`
 	EnableReferrerDetect bool        `toml:"enable_referrer_detect"`
+	EnableIndexDetect    bool        `toml:"enable_index_detect"`
 	TarfsConfig          TarfsConfig `toml:"tarfs"`
 	EnableBackendSource  bool        `toml:"enable_backend_source"`
 }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -31,6 +31,7 @@ func TestLoadSnapshotterTOMLConfig(t *testing.T) {
 		Experimental: Experimental{
 			EnableStargz:         false,
 			EnableReferrerDetect: false,
+			EnableIndexDetect:    false,
 		},
 		CleanupOnClose: false,
 		SystemControllerConfig: SystemControllerConfig{
--- a/docs/index_detection.md
+++ b/docs/index_detection.md
@@ -1,0 +1,89 @@
+# Index Detection
+
+## Overview
+
+Index Detection is a feature that automatically discovers Nydus alternative manifests within OCI index manifests. This enables transparent fallback to optimized Nydus images when available while keeping the original index manifest OCI compliant, allowing non-Nydus clients to pull regular OCI images.
+
+## Motivation
+
+The Index Detection feature addresses several limitations of the existing Referrer API-based detection:
+
+- **Registry Support**: Not all registries support the Referrer API
+- **Supply Chain Security**: Referrer API relies on external changes to the manifest, creating potential supply chain risks
+- **Universal Compatibility**: Index detection works with all OCI-compliant registries
+
+Unlike referrer-based detection, Index Detection packages everything into a single manifest, eliminating supply chain concerns while maintaining universal registry support.
+
+## How It Works
+
+### Detection Process
+
+1. **Index Manifest Parsing**: When a manifest digest is encountered, the system fetches and parses the original OCI index manifest
+2. **Platform Matching**: The system finds the original manifest descriptor within the index
+3. **Nydus Alternative Search**: It searches for platform-compatible manifests that contain:
+   - `platform.os.features` containing `nydus.remoteimage.v1`, or
+   - `artifactType` set to `application/vnd.nydus.image.manifest.v1+json`
+4. **Validation**: The found manifest is validated to ensure it's a valid Nydus manifest with metadata layers
+5. **Caching**: Results are cached to avoid repeated API calls for the same digest
+
+Both `platform.os.features` and `artifactType` are looked at because index manifests built using `merge-platform` before acceleration-service: v0.2.19 or nydus: v2.3.5 will have `platform.os.features` configured while images built after will have `artifactType` configured.
+
+### Detection Priority
+
+When both Index Detection and Referrer Detection are available, Index Detection takes priority due to its superior supply chain security properties.
+
+## Configuration
+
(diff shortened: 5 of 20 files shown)


.github/workflows/k8s-e2e.yml
```diff
@@ -28,7 +31,8 @@ jobs:
           cache-dependency-path: "go.sum"
       - name: Test
         run: |
-          AUTH_TYPE='${{ inputs.auth-type }}'
+          export AUTH_TYPE='${{ inputs.auth-type }}'
+          export INDEX_DETECT='${{ inputs.index-detect }}'
```
Fricounet: `export` are needed so that the env variables get picked up in `kind.sh`. Without it, they won't. Which makes me think that the `auth-type: cri` actually did not work as expected previously :sweat_smile:

imeoer:
@Fricounet Thanks for submitting such a helpful PR, overall it's quite good!
The nydus v2.3.5 has been released https://github.com/dragonflyoss/nydus/releases/tag/v2.3.5.

Fricounet:
@imeoer after doing more tests with this feature, I actually noticed an issue with the way nydus manifests are built currently on certain clients without nydus support.
They work fine with Docker however when trying to pull a `merge-platform` image on a containerd host without nydus, the pull fails because it tries to unpack the nydus layer and fails:
```
ERRO[2025-09-04T18:09:42.009149244+02:00] PullImage "my-image" failed  error="rpc error: code = Unknown desc = failed to pull and unpack image \"my-image\": failed to extract layer (application/vnd.oci.image.layer.nydus.blob.v1 sha256:dc99afaaa40c900e5ddd66d1ce5318c0ab97749d93b38c08d5ddc1611cfa0889) to overlayfs as \"extract-272586056-zzz5 sha256:dc99afaaa40c900e5ddd66d1ce5318c0ab97749d93b38c08d5ddc1611cfa0889\": failed to get stream processor for application/vnd.oci.image.layer.nydus.blob.v1: no processor for media-type"
```


After digging a bit into why this doesn't happen with soci implementation, I discovered that it's because we need to change the `config.mediaType` field in the image manifest. If the value is `application/vnd.oci.image.config.v1+json` (like it is right now), containerd will try to unpack the image's layers. However, if the value is different (like `application/vnd.amazon.soci.index.v2+json` for soci), then containerd skips unpacking the layers:
```
DEBU[2025-09-04T18:10:08.913672513+02:00] encountered unknown type application/vnd.amazon.soci.index.v2+json; children may not be fetched
```

Do you think it would be fine to modify the config mediaType to let's say the same value as the `artifactType` (application/vnd.nydus.image.manifest.v1+json) in the nydus images?

Fricounet:
@imeoer thanks for the feedback, I'll work on implementing this then.
Sorry in advance for all the releases i'll ask of you because we need:
- the fix to be made in nydus-snpahotter ([PR](https://github.com/containerd/nydus-snapshotter/pull/674))
- propagate the change in acceleration-service
- propagate the change in nydusify :sweat_smile:

Fricounet:
@imeoer thanks! Here's the acceleration-service PR https://github.com/goharbor/acceleration-service/pull/351

Fricounet:
Thanks and here is the final nydusify PR https://github.com/dragonflyoss/nydus/pull/1756

imeoer:
@Fricounet nydus [v2.3.6](https://github.com/dragonflyoss/nydus/releases/tag/v2.3.6) released!

Fricounet:
Thanks for all the bumps @imeoer  :bow: 

I've now discovered a new tricky issue though:
in the case where 2 OCI images have some layers in common
```
sha256:a606e9d35cbf05763aeb87455d69f3675828408bd4be5cdb9082f8767ef6052a
  \_ sha256:3ecb0bc443f8b1368bc94602369349be5731ae19bb2214384e75d332eaf63be5 (image A)
  \_ sha256:cd6bd35c7a8a26144c13402f9746f81a494c8105789520764d53ce40117e8d89 (image B)
```

If image A has been `nydusify --merge-manifest` converted but not image B and if image A was pulled before image B.
When containerd attempts to pull image B, when its active layer is prepared, the nydus-snapshotter will iterate the parent snapshots and find the common layer which has a nydus index reference. This has 2 consequences:
- first the nydus-snapshotter will try to fetch the metadata blob from image A. Depending on the registry, if the credentials that were initially used for this ref had an expiration, the pull can fail because the creds aren't valid anymore. This could be solved by some caching or some creds renewal logic however I'm not sure we'd want to do it because of the next point
- if it succeeds, the nydus-snapshotter will answer the `Mount` call with a mount slice that tries to mount the common layer among the `lowerdir`. However, correct me if I'm wrong, the nydus bootstrap layer is not scoped to only the original OCI layer content but contains the full image, so the final mount will actually contain the entirety of image A and not only the common layer

I'm trying to figure out how we can separate the non-nydus image pull from the nydus one so that this situation doesn't happen but I don't have an easy solution for now:
- when it pulls image B, it seems that containerd doesn't even try to `Prepare` the common layer again, probably because it considers that the layer has already been handled in image A
- all of this seems to be caused by the fact the containerd associates the same snapshot Key to the layers in both images

The value for the snapshot key is [computed](https://github.com/containerd/containerd/blob/main/core/unpack/unpacker.go#L350) from the `rootfs.diff_ids` list in the config object, so I think one way to solve this would be to re-compute the config object for the OCI image in the nydus-merged manifest to a different value from the original. But it seems very hacky :/

imeoer:
Hi @Fricounet, 

> the bootstrap layer only contains a single EntryBootstrap which is merged from all the EntryBootstrap in the other layers

Yes it's correct understanding, the `EntryBootstrap` from each layer will be merged into a single bootstrap in an overlay-like way, which we can call it merged bootstrap.

> fetching the whole layer and unpack it to extract the Bootstrap (defeats the purpose of doing lazy loading)

We can fetch and extract just the layered-bootstrap from the remote blob, instead of fetching the entire blob.

> fetch only the EntryBootstrap using range calls but is it even possible as the blob is zstd compressed?

The blob data (`EntryBlob` data segment) uses chunked zstd compression, rather than whole zstd compression:

EntryBlob: `zstd(chunk) | ... | zstd(chunk)`, a file is split into one or more chunks, each maximum sized at 1MiB, chunks from multiple files are concatenated to a blob.

> Are you fundamentally against me trying to make this work?

It's okay for me, it's very useful to mount layer independently, but it requires separating the logic from the normal merged bootstrap process in the snapshotter.

Fricounet:
> Thanks for the explanation, LGTM! This approach might feel a bit hacky, but it brings greater benefit when combining OCI v1 and nydus into a single image reference.

Indeed, since the merge-platform is useless today, it might be better to have something that works albeit a bit ugly compared to nothing at all. I've opened https://github.com/goharbor/acceleration-service/pull/352

.github/workflows/k8s-e2e.yml
fidencio: Please, use `env:` instead.

Fricounet: thanks for the tips, done

imeoer (approved):

imeoer:
Thanks for the excellent work!

PR #672, opened 2025-09-05
feat(cache): add hardlink or fallback copy for blob.meta to cache

sf1999817:
## Overview
​​This PR adds support for, after extracting a bootstrap file, preferentially creating a hardlink (if possible) from the associated blob.meta file to the cache directory to save disk space and improve performance. If hardlinking is not possible , it automatically falls back to copying the file, ensuring the feature works reliably.​​

## Related Issues
No related issue

## Change Details
This PR introduces a new function copyBlobMetaFilesto locate and cache all *.blob.metafiles associated with the bootstrap file during the mount process. It prioritizes creating hardlinks (for efficiency and space saving) and falls back to copying the files if hardlinking is not possible (e.g., cross-filesystem). This improves blob.meta handling reliability and cache management.

## Test Results
hardlink结果

## Change Type
- [ ] Bug Fix
- [x] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist
- [ ] I have added appropriate comments to my code (if applicable).


pkg/filesystem/fs.go
```diff
@@ -391,6 +398,95 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
+		return errors.Wrapf(err, "read meta file %s", metaPath)
+	}
+
+	headerData, err := os.ReadFile(headerPath)
+	if err != nil {
+		return errors.Wrapf(err, "read header file %s", headerPath)
+	}
+
+	metaSize := len(metaData)
+	alignedSize := (metaSize + 4095) & ^4095
```
BraveY: it's better to add some comments about the logic. Otherwise, numbers like 4095 here would be hard for later readers to understand.

pkg/filesystem/fs.go
```diff
@@ -391,6 +398,96 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
+		}
+
+		if err := os.Remove(blobMetaHeaderPath); err != nil {
+			log.L.Warnf("Failed to remove source blob meta header file %s: %v", blobMetaHeaderPath, err)
+		}
+	}
+	return nil
+}
+
+func (fs *Filesystem) assembleBlobMetaFile(metaPath, headerPath, destPath string) error {
```
imeoer: Can we assemble the `.blob.meta.header` and `.blob.meta` in the snapshotter converter rather than at runtime?

pkg/filesystem/fs.go
```diff
@@ -391,6 +397,43 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
+	pattern := filepath.Join(bootstrapDir, "*.blob.meta")
+	matches, err := filepath.Glob(pattern)
+	if err != nil {
+		return errors.Wrap(err, "glob blob meta files")
+	}
+
+	for _, srcPath := range matches {
+		fileName := filepath.Base(srcPath)
+		dstPath := filepath.Join(cacheDir, fileName)
+		if _, err := os.Stat(dstPath); err == nil {
```
BraveY: This check seems not strict enough; the path exists, but it doesn't necessarily mean it has been linked. It might just be the same filename. I suggest not skipping it and directly enforcing the logic below, as the data is not large and won't incur significant overhead.

pkg/filesystem/fs.go
```diff
@@ -391,6 +397,43 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
+			}
+		} else {
+			log.L.Debugf("Created hardlink for blob meta: %s -> %s", srcPath, dstPath)
+		}
+	}
+
+	return nil
+}
+
+func (fs *Filesystem) copyFile(src, dst string) error {
```
BraveY, line 427:
```suggestion
func (fs *Filesystem) copyFile(src, dst string) error {
	source, err := os.Open(src)
	if err != nil {
		return err
	}
	defer source.Close()

	destination, err := os.Create(dst)
	if err != nil {
		return err
	}
	defer destination.Close()

	_, err = io.Copy(destination, source)
	return err
}
```
Use stream copy to save memory.

PR #700, opened 2025-12-09
snapshot: implement lazy parent recovery for proxy mode

### LINKED ISSUE #680: [Bug] create snapshot: missing parent "..." bucket: not found
### Problem Description

After having kata-containers using nydus snapshotter for around a week, we've that our node became unusable due to the following issue:
```
  Normal   Scheduled               6s    default-scheduler  Successfully assigned default/kata-quay-io-mongodb-mongodb-community-server-sha256-8b737338 to snapshotters-ci-1
  Warning  FailedCreatePodSandBox  6s    kubelet            Failed to create pod sandbox: rpc error: code = NotFound desc = failed to create containerd container: create snapshot: missing parent "k8s.io/2/sha256:961e93cda9dd918dbe26aca24cccd6c5db05176850d2c53476d881df5d0d4816" bucket: not found
```

This could be easily reproduced with containerd 1.7, 2.0, 2.1, and 2.2.

The nydus configuration used is:
```
version = 1

# Snapshotter's own home directory where it stores and creates necessary resources
root = "/var/lib/containerd-nydus"

# The snapshotter's GRPC server socket, containerd will connect to plugin on this socket
address = "/run/containerd-nydus/containerd-nydus-grpc.sock"

[daemon]
# Enable proxy mode
fs_driver = "proxy"

[snapshot]
# Insert Kata volume information to `Mount.Options`
enable_kata_volume = true
```

Containerd's configuration has:
```
[plugins."io.containerd.cri.v1.images"]
disable_snapshot_annotations = false

...

[proxy_plugins.nydus]
type = "snapshot"
address = "/run/nydus-snapshotter/containerd-nydus-grpc.sock"
```

@imeoer, I'd appreciate if you have any hint on why it happens and how we could potentially avoid such issue.

### Expected Behavior

We would never face the issue described above.

### Actual Behavior

The issue described above.

### How to reproduce

I have a CI running every hour doing the following:
1, Deploy nydus
2. Pull one specific image using overlayfs
3. Pull the same specific image using nydus-snapshotter
4. Uninstall nydus
5. Repeat all the steps

And after a week running, it started failing.

### Environment Details

- Nydus-snapshotter version: v0.15.2
- Nydus version: N/A
- Container runtime: containerd
- Operating System: Ubuntu 24.04
- Kernel version: 6.8.0-71-generic

### Additional Information

_No response_

fidencio:
## Overview

There are several scenarios where a desynchronization between containerd's database and the nydus-snapshotter's BoltDB can occur, some of those being:
- BoltDB file corruption
- Storage issues affecting only nydus data
- Backup/restore asymmetry
- Upgrade failures

Such desynchronizations then lead to a createSnapshot() failure with 'missing parent bucket: not found', as reported in #680.

## Related Issues

Fixes #680 

## Change Details

When createSnapshot fails with 'missing parent bucket: not found', the new recovery mechanism, let's:
1. Detect the error pattern
2. For proxy mode (Kata Containers guest pulling), create a placeholder committed snapshot in the local BoltDB
3. Retry the original operation

The placeholder snapshot is marked with labels:
- nydus.recovered=true
- nydus.recovered.at=<timestamp>

This allows the system to self-heal and continue operating rather than requiring manual intervention to clean up stale containerd references.

## Test Results

This is being battle tested right now as part of these 2 CIs:
* https://github.com/fidencio/kata-containers/actions/workflows/run-nydus-snapshotter-install-cycle-test.yaml
* https://github.com/fidencio/kata-containers/actions/workflows/run-nydus-snapshotter-stability-test.yaml

Although I'm not 100% sure it'll fix the issue, I'm opening the PR for an early review and inviting @seamounts to also give it a try.

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -539,7 +539,12 @@ func (fs *Filesystem) Umount(_ context.Context, snapshotID string) error {
 			return errors.Wrapf(err, "remove snapshot %s", snapshotID)
 		}
 	case config.FsDriverNodev, config.FsDriverProxy:
-		// Nothing to do
+		// For proxy mode, we still need to clean up the DB entry to prevent
+		// "already exists" errors on subsequent Mount() calls.
+		if err := fsManager.RemoveRafsInstance(snapshotID); err != nil {
+			// Log but don't fail - the entry might not exist
+			log.L.Debugf("remove instance %s from DB: %v", snapshotID, err)
+		}
 	default:
 		return errors.Errorf("unknown filesystem driver %s for snapshot %s", fsDriver, snapshotID)
 	}


fidencio:
cc @imeoer, please, let me know if this makes sense, or if you'd have another suggestion on how to approach the potential desynchronization issue.

snapshot/snapshot.go
```diff
@@ -284,6 +291,22 @@ func NewSnapshotter(ctx context.Context, cfg *config.SnapshotterConfig) (snapsho
...
+	// Initialize containerd client for lazy parent recovery.
+	// This allows the snapshotter to query containerd for snapshot
+	// metadata when recovering from desynchronization scenarios where
+	// local BoltDB is missing entries.
+	var containerdClient *client.Client
+	containerdAddr := cfg.ContainerdAddress
+	if containerdAddr == "" {
+		containerdAddr = defaultContainerdAddress
+	}
+	containerdClient, err = client.New(containerdAddr)
```
bergwolf: What if containerd restarts? We should not expect `containerdClient` to work all the time.

imeoer:
@fidencio Thanks for the PR! I'll review it this week. :)

misc/snapshotter/config-proxy.toml
```diff
@@ -6,6 +6,19 @@ address = "/run/containerd-nydus/containerd-nydus-grpc.sock"
...
+# This is particularly important for proxy mode (Kata Containers) to recover
+# from scenarios where nydus's database gets wiped but containerd still has
+# snapshot references.
+# Different Kubernetes distributions use different socket paths:
+#   - Standard/vanilla k8s: /run/containerd/containerd.sock (default)
+#   - MicroK8s:             /var/snap/microk8s/common/run/containerd.sock
+#   - k3s:                  /run/k3s/containerd/containerd.sock
+#   - k0s:                  /run/k0s/containerd.sock
+#   - RKE2:                 /run/k3s/containerd/containerd.sock
+containerd_address = "/run/containerd/containerd.sock"
```
imeoer: Thanks so much for this PR! It looks like the fix is focused specifically on the proxy mode scenario. The change avoids inconsistency between the snapshotter's records and containerd's by creating a parent snapshot record on the snapshotter side.

But since we don't actually use the containerdClient, I'd suggest removing the `containerd_address` configuration and `containerdClient` for now. Also, because the usual call flow goes from containerd to the remote snapshotter, adding reverse calls might feel unnatural. If we later find a real need to access containerd for snapshot records, we can add it back then. But for this change, let's keep things simple and leave that part out.

imeoer (commented):
others LGTM!

misc/snapshotter/config-proxy.toml
fidencio: Ouch, you're right. The first implementation for this I was directly calling containerd (and yes, I know it's unnatueral), but then I realised:
* we call containerd -> containerd calls nydus :-/

Then I went ahead and made this specific for proxy mode and forgot to remove the depency.

It's update now.

Would it be possible to have a release of the project so we could bump to it on Kata / CoCo side?

Zephyrcf (approved):
LGTM for proxy-mode lazy recovery!

misc/snapshotter/config-proxy.toml
imeoer: Thanks! I'll create the new release once the PR's CI pass.

imeoer:
@fidencio Seems to be related to the ghcr service: failed to resolve reference \"ghcr.io/dragonflyoss/image-service/tomcat:nydus-nightly-v5\": failed to do request: Head \"[https://ghcr.io/v2/dragonflyoss/image-service/tomcat/manifests/nydus-nightly-v5\](https://ghcr.io/v2/dragonflyoss/image-service/tomcat/manifests/nydus-nightly-v5/)": dial tcp 140.82.116.33:443: i/o timeout

imeoer:
@fidencio Released tag: https://github.com/containerd/nydus-snapshotter/releases/tag/v0.15.10

PR #708, opened 2026-02-01
snapshot: fix proxy mode mounting for snapshots without nydus-proxy mode label

fidencio:
## Overview

When containerd unpacks an image for a remote snapshotter like nydus in proxy mode, it passes the CRIImageRef label via snapshots.WithLabels during the Prepare/Unpack flow. This label contains the image reference needed by nydus to construct the kata virtual volume metadata.

The mountProxy() function was being called without passing the available labels, creating a rafs.Rafs{} with an empty Annotations map. This caused nydus to fall back to using "dummy-image-reference" as the source, breaking guest-side image pulling. The fix passes the labels through to mountProxy() so CRIImageRef is available when constructing the kata virtual volume.

Additionally, snapshots may have CRIImageRef but lack the nydus-proxy-mode label. This happens when an image was unpacked before containerd properly set all labels, or when labels were updated via containerd's fixSnapshotLabels after the initial unpack. In these cases the code would incorrectly call mountNative() instead of mountProxy(), causing mount failures because proxy mode directories don't contain actual content (content is pulled inside the guest). The fix adds a fallback check before mountNative: if we're in proxy driver mode and the snapshot or its parent chain has CRIImageRef, use mountProxy instead.

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -439,13 +439,45 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
 
 	if treatAsProxyDriver(info.Labels) {
 		log.L.Warnf("[Mounts] treat as proxy mode for the prepared snapshot by other snapshotter possibly: id = %s, labels = %v", id, info.Labels)
-		return o.mountProxy(ctx, *snap)
+		mountLabels := o.getLabelsWithCRIImageRef(ctx, info.Labels, info.Parent)
+		if _, ok := mountLabels[label.CRIImageRef]; !ok {
+			log.L.Warnf("[Mounts] No CRIImageRef found in snapshot %s or its parents - guest pull may fail", key)
+		}
+		return o.mountProxy(ctx, mountLabels, *snap)
 	}
 
 	if needRemoteMounts {
 		return o.mountRemote(ctx, info.Labels, *snap, metaSnapshotID, key)
 	}
 
+	// Handle the case where the snapshotter is in proxy mode and the snapshot has
+	// CRIImageRef label but not nydus-proxy-mode label. This can happen when:
+	// 1. The image was unpacked before the containerd fix that adds proper labels
+	// 2. Containerd detected that existing snapshots had incorrect labels (e.g., a
+	//    digest instead of a pullable image reference) and updated them to contain
+	//    the correct CRIImageRef value needed for guest-side image pulling
+	// In this case, we should still use mountProxy instead of mountNative.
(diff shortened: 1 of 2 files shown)


fidencio:
cc @imeoer

snapshot/snapshot.go
```diff
@@ -439,13 +439,101 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
...
 		log.L.Warnf("[Mounts] treat as proxy mode for the prepared snapshot by other snapshotter possibly: id = %s, labels = %v", id, info.Labels)
-		return o.mountProxy(ctx, *snap)
+		// If the current snapshot doesn't have CRIImageRef, try to get it from parent snapshots.
+		// This handles the case where the container layer is created after the image layers,
+		// and the image layers have the labels but the container layer doesn't.
+		mountLabels := info.Labels
+		if mountLabels == nil {
+			mountLabels = make(map[string]string)
+		}
+		if _, ok := mountLabels[label.CRIImageRef]; !ok && info.Parent != "" {
```
BraveY: The code for walking the parent chain to find CRIImageRef is duplicated in three places (lines ~456, ~500, and ~528).
Could we refactor this into a helper function?
Others LGTM. Thanks for the PR.

fidencio: @BraveY, updated, thanks!

snapshot/snapshot.go
```diff
@@ -439,13 +439,43 @@ func (o *snapshotter) Mounts(ctx context.Context, key string) ([]mount.Mount, er
...
 	}
 
 	if needRemoteMounts {
 		return o.mountRemote(ctx, info.Labels, *snap, metaSnapshotID, key)
 	}
 
+	// Handle the case where the snapshotter is in proxy mode and the snapshot has
+	// CRIImageRef label but not nydus-proxy-mode label. This can happen when:
+	// 1. The image was unpacked before the containerd fix that adds proper labels
+	// 2. The labels were updated after the initial unpack via containerd's fixSnapshotLabels
```
imeoer: Hi @fidencio , thanks for the PR, can we explain what's `fixSnapshotLabels` here? others LGTM!

fidencio: Sure, I will do the addition later Today. Thanks @imeoer!

fidencio: Fixed, and moved the PR to "Draft" till the work on containerd side gets merged.

PR #711, opened 2026-02-02
[snapshot] Reimplement cache GC

### LINKED ISSUE #687: [Bug] Cache not GCed during snaphsot removal while using referrer detection (and future index detection)
### Problem Description

While working on #669, I discovered a bug during cache GC which is also shared by the referrer detection feature due to how similar both feature work.
They both lead to a mismatch between the containerd layers sha and the actual shas of the blobs being pulled.
In both cases, containerd will think it is preparing the layers of an OCI image (and the shas in containerd internal store will reflect this) but the nydus snaphsotter will actually skip downloading the OCI layers in favor of getting only the nydus metadata blob. And when nydusd is started, it will pull the nydus blobs (with they own sha) and store them under their sha.

[This is fine until containerd asks to remove the OCI layer snapshot, this is where the GC is supposed to act through RemoveBlobCache](https://github.com/containerd/nydus-snapshotter/blob/main/pkg/cache/manager.go#L99) but the blodID that's passed do not correspond to any blobID stored on disk since the blob sha will match the ones from the nydus image.

As a result blobs end up being leaked and are never cleaned up.

### Expected Behavior

Cache files shouldn't leak and should be GCed when they become unused.

### Actual Behavior

To illustrate, I've create I've created an index referrer image (but this works the same with a registry referrer image) from the busybox image:
```
$ crane manifest localhost:5000/busybox:latest-nydus
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:362d1b3ef32293cf36ac559a27d47945ffc3728a7ba9c60e7a66d4d447023655",
      "size": 1197,
      "annotations": {
        "containerd.io/snapshot/nydus-source-digest": "sha256:182014572d8981d8323fe9944876f63b39694e16ce08ae6296e97686c52b150c",
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:c8731312c921189b524cacfcba67084c05953b8f98502b4643d8fad81583372b",
      "size": 1387,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "artifactType": "application/vnd.nydus.image.manifest.v1+json"
    },
  ]
}
```

The OCI image:
```
$ crane manifest localhost:5000/busybox:latest-nydus@sha256:362d1b3ef32293cf36ac559a27d47945ffc3728a7ba9c60e7a66d4d447023655
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:99a438a05de5e50dcf214a9ddd61a580e769b5eea0181d14d3af7edd740dbc4e",
    "size": 726,
    "annotations": {
      "containerd.io/snapshot/nydus-source-digest": "sha256:0ed463b26daee791b094dc3fff25edb3e79f153d37d274e5c2936923c38dac2b"
    }
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1",
      "size": 32,
      "annotations": {
        "containerd.io/snapshot/nydus-empty-layer": "true"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494",
      "size": 2213082
    }
  ],
  "annotations": {
    "containerd.io/snapshot/nydus-source-digest": "sha256:182014572d8981d8323fe9944876f63b39694e16ce08ae6296e97686c52b150c",
  }
}
```

The Nydus image:
```
$ crane manifest localhost:5000/busybox:latest-nydus@sha256:c8731312c921189b524cacfcba67084c05953b8f98502b4643d8fad81583372b
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.nydus.image.config.v1+json",
    "digest": "sha256:bd7db24b9e0bf3ec3f50ad3a3fdfa888a47fb6405f122516509d18f391ed0500",
    "size": 521
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.nydus.blob.v1",
      "digest": "sha256:0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f",
      "size": 2243614,
      "annotations": {
        "containerd.io/snapshot/nydus-blob": "true"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:b74de3635e3f8a263aeb8e8040aaf22dfc0ea45a996747e33b45d04764381af0",
      "size": 4952,
      "annotations": {
        "containerd.io/snapshot/nydus-bootstrap": "true",
        "containerd.io/snapshot/nydus-fs-version": "6"
      }
    }
  ],
  "annotations": {
    "containerd.io/snapshot/nydus-builder-version": "v2.3.6",
    "containerd.io/snapshot/nydus-fs-version": "6",
    "containerd.io/snapshot/nydus-source-digest": "sha256:182014572d8981d8323fe9944876f63b39694e16ce08ae6296e97686c52b150c",
    "containerd.io/snapshot/nydus-source-reference": "localhost:5000/busybox:latest"
  }
}
```
Then I pulled the image.

Containerd shows the layers belonging to the OCI image (`5f70` corresponds to the empty layer unpacked sha):
```
$ ctr snapshots tree
 sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef
  \_ sha256:ab67c33135a9345ac64e63344332d566e7511901be1ee03428734a68deead80a
    \_ 7a33d73b32b903fbd36ed858bfe4a4f68a6fedac85e1827f6ea01bd49459bfab
``` 

but the nydus cache shows the sha from the Nydus image layers 
```
ls io.containerd.snapshotter.v1.nydus/cache
.rw-r----- 4,5M root  3 nov.  10:58 0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.blob.data
.rw-r----- 4,1k root  3 nov.  10:58 0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.blob.data.chunk_map
.rw-r----- 8,2k root  3 nov.  10:58 0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.blob.meta
```

then I removed the nydus image `crictl rmi -a`

Nydus logs shows it's attempting to delete the cache for the OCI layers (which fails because there's no file for this sha):
```
DEBU[2025-11-03T11:40:26.072134679+01:00] [Remove] snapshot with key k8s.io/38/7a33d73b32b903fbd36ed858bfe4a4f68a6fedac85e1827f6ea01bd49459bfab
INFO[2025-11-03T11:40:26.072215859+01:00] [Remove] snapshot with key k8s.io/38/7a33d73b32b903fbd36ed858bfe4a4f68a6fedac85e1827f6ea01bd49459bfab snapshot id 10
DEBU[2025-11-03T11:40:26.076682211+01:00] [Remove] snapshot with key sha256:ab67c33135a9345ac64e63344332d566e7511901be1ee03428734a68deead80a

INFO[2025-11-03T11:40:26.076720551+01:00] [Remove] snapshot with key sha256:ab67c33135a9345ac64e63344332d566e7511901be1ee03428734a68deead80a snapshot id 9
INFO[2025-11-03T11:40:26.076813907+01:00] remove cache sha256:80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494
DEBU[2025-11-03T11:40:26.076874350+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494.chunk_map doest not exist.
DEBU[2025-11-03T11:40:26.076890718+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494.blob.data.chunk_map doest not exist.
DEBU[2025-11-03T11:40:26.076902706+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494.blob.meta doest not exist.
DEBU[2025-11-03T11:40:26.076914722+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494 doest not exist.
DEBU[2025-11-03T11:40:26.076925227+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494.blob.data doest not exist.
DEBU[2025-11-03T11:40:26.076935506+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494.image.disk doest not exist.
DEBU[2025-11-03T11:40:26.076951661+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/80bfbb8a41a2b27d93763e96f5bdccb8ca289387946e406e6f24053f6a8e8494.layer.disk doest not exist.
DEBU[2025-11-03T11:40:26.082903830+01:00] [Remove] snapshot with key sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef

INFO[2025-11-03T11:40:26.082941505+01:00] [Remove] snapshot with key sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef snapshot id 8
INFO[2025-11-03T11:40:26.083035272+01:00] remove cache sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1
DEBU[2025-11-03T11:40:26.083090548+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.chunk_map doest not exist.
DEBU[2025-11-03T11:40:26.083113085+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.blob.data.chunk_map doest not exist.
DEBU[2025-11-03T11:40:26.083128591+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.blob.meta doest not exist.
DEBU[2025-11-03T11:40:26.083143034+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 doest not exist.
DEBU[2025-11-03T11:40:26.083156983+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.blob.data doest not exist.
DEBU[2025-11-03T11:40:26.083170659+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.image.disk doest not exist.
DEBU[2025-11-03T11:40:26.083184679+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1.layer.disk doest not exist.
DEBU[2025-11-03T11:40:26.087702455+01:00] [Cleanup] snapshots
```

And the nydus layers cache are still present on disk:
```
ls io.containerd.snapshotter.v1.nydus/cache
.rw-r----- 4,5M root  3 nov.  10:58 0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.blob.data
.rw-r----- 4,1k root  3 nov.  10:58 0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.blob.data.chunk_map
.rw-r----- 8,2k root  3 nov.  10:58 0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.blob.meta
```

### How to reproduce

1.
2.
3.

### Environment Details

- Nydus-snapshotter version: latest
- Nydus version: latest
- Container runtime:
- Operating System:
- Kernel version:

### Additional Information

I can see two ways of fixing this issue:
1. reinstate some kind of regular GC for unused nydus blobs. I know this was removed in https://github.com/containerd/nydus-snapshotter/pull/233 in favor of relying on containerd Cleanup mechanism but this approach doesn't work when using index/referrer detectors because those feature actively rely on pulling a different image than what containerd expects by design
   a. the cache logic doesn't have to be very complicated as it's fairly easy to find the currently used nydus files using the cache metrics exposed by the sockets: `for daemon in $(ls); do nydusctl --sock ./${daemon}/api.sock metrics cache --raw | jq -r .underlying_files[]; done | sort -u` and we can compare this list with the list of files present in the cache directory and proceed to the removal of the files not actively used by any nydus daemon. My main concern regarding having a regular GC running is about how sensitive is Nydusd if one of its blob referenced in `underlying_files` were to be deleted while it's running. Is nydusd able to recover? Or is it gonna break everything?
3. find a way to link the OCI layers to the nydus layers using some kind of annotation but I don't really know how we can do it because there's no real link between the original OCI layers and the final nydus ones. The layer count in the nydus image doesn't necessarily match the ones from the OCI image

What do you think?

Fricounet:
## Overview
_Please briefly describe the changes your pull request makes._

This PR is introducing back a GC mechanism for nydusd cache files. The cache was initially removed in https://github.com/containerd/nydus-snapshotter/pull/233 but is actually vital for the snapshotter to work correctly:

1. There is a structural issue whith cache removal when it occurs during containerd's Remove call. There's an inherent mismatch between containerd layers representation and nydusd's. Containerd stores layers under a computed key called the ChainID where each ID depends on its parents. This help containerd ensure that 2 snapshots sharing the same key will always have the same parent. On the other hand, nydusd stores each layer on disk under its layer digest directly (the same one you can see in the image manifest). This allows nydusd to reuse same blobs when used in different images.
However, this creates a rift between both systems because if we have 2 images which have the same layers but in different order (layerA+layerB for image1 and layerB+layerA for image2), then those will be completely separate images for containerd because their chainID won't match but they will be backed by the same blobs on disk.
As a result, if containerd has to remove image1, it won't consider its layers as in-use by an other image  and will Remove them. Thus, if the snapshotter also removes the cache, it will remove the same blob files that are in use by image2. The issue might not appear directly du to image2 nydusd keeping a fd open on the blob files. However, a reboot will lead image2 to have missing files.

```mermaid
sequenceDiagram
      participant Reg as Registry<br/>┌─ Image A (sha256:abc123)<br/>│  └─ Layer1 → Layer2<br/>└─ Image B (sha256:def456)<br/>   └─ Layer2 → Layer3
      participant C as Containerd
      participant NS as Nydus Snapshotter
      participant Nydusd as nydusd
      participant Cache as Cache Directory
      participant Disk as Containerd State

      Note over Reg,Disk: Pull Image A (Layer1 → Layer2)
      C->>NS: Prepare Layer1
      NS-->>C: Remote mount
      C->>Disk: Track ChainID(Layer1)
      C->>NS: Prepare Layer2
      NS-->>C: Remote mount
      C->>Disk: Track ChainID(Layer1, Layer2)
      C->>NS: Mount Image A
      NS->>Nydusd: Start nydusd for Image A
      Nydusd->>Reg: Download blob sha256:layer1-digest
      Nydusd->>Cache: Store blob sha256:layer1-digest
      Nydusd->>Reg: Download blob sha256:layer2-digest
      Nydusd->>Cache: Store blob sha256:layer2-digest
      Note over Cache,Disk: Cache: [layer1-digest, layer2-digest]<br/>Containerd: [ChainID_A_L1, ChainID_A_L2]

      Note over Reg,Disk: Pull Image B (Layer2 → Layer3)
      C->>NS: Prepare Layer2
      NS-->>C: Remote mount
      C->>Disk: Track ChainID(Layer2)
      C->>NS: Prepare Layer3
      NS-->>C: Remote mount
      C->>Disk: Track ChainID(Layer2, Layer3)
      C->>NS: Mount Image B
      NS->>Nydusd: Start nydusd for Image B
      Nydusd->>Cache: Check blob sha256:layer2-digest
      Cache-->>Nydusd: ✓ Already exists!
      Nydusd->>Reg: Download blob sha256:layer3-digest
      Nydusd->>Cache: Store blob sha256:layer3-digest
      Note over Cache,Disk: Cache: [layer1-digest, layer2-digest, layer3-digest]<br/>Containerd: [ChainID_A_L1, ChainID_A_L2, ChainID_B_L2, ChainID_B_L3]<br/>⚠️ layer2-digest used by both images!

      Note over Reg,Disk: Remove Image A - PROBLEM
      C->>NS: Remove ChainID_A_L2
      NS->>Cache: Delete blob sha256:layer2-digest
      Cache-->>NS: ✓ Deleted
      NS-->>C: Removed
      C->>Disk: Remove tracking ChainID_A_L2
      Note over Cache,Disk: ⚠️ BUG!<br/>Cache: [layer1-digest, layer3-digest]<br/>Image B still needs layer2-digest!<br/>Image B will fail on nydusd restart!
```

2. as mentioned in #687, the same ID mismatch will happen by design for index referrer images although in that case, its consequence is that cache blobs end up leaking on disk.

```mermaid
  sequenceDiagram
      participant Reg as Registry<br/>┌─ Nydus Image with index-detect sha256:abc123<br/>├─ OCI Layer sha256:80bfbb8a<br/>└─ Nydus Layer sha256:0713669e
      participant C as Containerd
      participant NS as Nydus Snapshotter
      participant Nydusd as nydusd
      participant Cache as Cache Directory
      participant Disk as Containerd State

      Note over Reg,Disk: Image Pull with Index Detection
      C->>NS: Prepare OCI layer sha256:80bfbb8a...
      NS->>Reg: Query index<br/>for sha256:80bfbb8a...
      Reg-->>NS: Index manifest sha256:abc123...
      NS->>NS: Found Nydus alternative!<br/>sha256:0713669e...
      NS-->>C: Remote mount
      C->>Disk: Track OCI layer sha256:80bfbb8a...
      C->>NS: Mount layer sha256:80bfbb8a...
      NS->>Nydusd: Start nydusd
      Nydusd->>Reg: Download Nydus blob<br/>sha256:0713669e...
      Reg-->>Nydusd: Nydus blob data
      Nydusd->>Cache: Store blob<br/>sha256:0713669e...
      Cache-->>Nydusd: Blob stored
      Nydusd-->>NS: Nydusd ready
      NS-->>C: Mount ready
      Note over Cache,Disk: ⚠️ State mismatch:<br/>Cache has sha256:0713669e (Nydus)<br/>Containerd tracks sha256:80bfbb8a (OCI)

      Note over Reg,Disk: Container Removed - Cleanup Attempt
      C->>NS: Remove layer sha256:80bfbb8a...
      NS->>Cache: Delete blob sha256:80bfbb8a...
      Cache-->>NS: ✗ File not found!<br/>Only sha256:0713669e exists
      NS-->>C: Removed
      C->>Disk: Remove tracking sha256:80bfbb8a...
      Note over Cache,Disk: ✗ Cache leak!<br/>Cache has sha256:0713669e (orphaned)<br/>No tracking for this blob
```

Note that for 1. this PR is addressing a situation which has a low change of occurring because it needs quite a bit of coincidences to be visible:

- 2 nydus images which have some layers which are the same but are not in the same order (so the chainID will be different) OR 1 regular nydus image and 1 nydus OCI-compatible image which have at least 1 layer in common (their chainID will necessarily be different due to the design of OCI-compatible images)
- one of the regular nydus image is deleted from the host
- enough time has passed so that the registry credentials from the other image have now expired
- nydusd crashes OR the kernel crashes, causing a new boot. The remaining image will fail to fetch its missing files on restart

To address these issues we reintroduce cache GC using a simple process done in 2 steps:

1. when nydusd starts and the snapshotter is waiting for it to be ready, once nydusd is running, the snapshotter will query its cache endpoint to get all the underlying files used by the RAFS instance. This information is then stored in the rafs in-memory cache and persisted in the nydus boltdb. This operation takes ~5ms due to needing to make an api call + modify the db. This will ensure that a snapshotter restart doesn't lose this information.
2. when the snapshotter Cleanup method is called, we iterate over all the running RAFS to list all the currently in-use files. Followed by a listing of all the files present in the cache directory. The files present in the cache but not in-use will be removed. This operation took <1ms in my tests but will grow with the amount of files in the cache directory. Though, the Cleanup method is here exactly for the use case of running potentially long cleanup operations so it's ok if it takes a bit longer

Addionally, we stop removing the cache blobs in the snapshotter `Remove` call and let the Cleanup method handle this. This, way it is impossible for the situation 1 to occur anymore because we make sure to only remove cache blobs which are unused.

## Related Issues
_Please link to the relevant issue. For example: `Fix #123` or `Related #456`._

Fix #687 

## Change Details

The Remove+Cleanup process illustrated above now becomes like this:

```mermaid
sequenceDiagram
      participant Reg as Registry<br/>┌─ Image A (sha256:abc123)<br/>│  └─ Layer1 → Layer2<br/>└─ Image B (sha256:def456)<br/>   └─ Layer2 → Layer3
      participant C as Containerd
      participant NS as Nydus Snapshotter
      participant Nydusd as nydusd
      participant Cache as Cache Directory
      participant Disk as Containerd State

      Note over Reg,Disk: Remove Image A - SOLUTION
      C->>NS: Remove ChainID_A_L2
      NS->>NS: Skip cache deletion<br/>Defer to Cleanup GC
      NS-->>C: Removed
      C->>Disk: Remove tracking ChainID_A_L2
      Note over Cache,Disk: ✓ Safe state:<br/>Cache: [layer1-digest, layer2-digest, layer3-digest]<br/>Image B still works!

      Note over Reg,Disk: Later: Snapshotter Cleanup (GC)
      C->>NS: Cleanup
      NS->>Nydusd: Query files in-use for active RAFS instances
      Nydusd-->>NS: In-use blobs:<br/>[sha256:layer2-digest, sha256:layer3-digest]
      NS->>Cache: List all cache blobs
      Cache-->>NS: All blobs:<br/>[sha256:layer1-digest, sha256:layer2-digest,<br/>sha256:layer3-digest]
      NS->>NS: Compare: cache vs in-use<br/>Find orphans: [layer1-digest]
      NS->>Cache: Delete orphaned blobs [layer1-digest]
      Cache-->>NS: ✓ Removed
      Note over Cache,Disk: ✓ Correct cleanup:<br/>Cache: [layer2-digest, layer3-digest]<br/>Only truly unused blobs removed
```

## Test Results

I created an index-referrer image and got the following logs when it got cleaned up:
```
DEBU[2025-12-05T17:43:42.736790365+01:00] [Cleanup] snapshots
INFO[2025-12-05T17:43:42.736866966+01:00] [Cleanup] orphan directories [/containerd-local/io.containerd.snapshotter.v1.nydus/snapshots/89 /containerd-local/io.containerd.snapshotter.v1.nydus/snapshots/90]
DEBU[2025-12-05T17:43:42.736876932+01:00] no RAFS filesystem instance associated with snapshot 89
INFO[2025-12-05T17:43:42.741153287+01:00] Destroy nydusd daemon d4pgj4pjjoqfk878ge2g. Host mountpoint /containerd-local/io.containerd.snapshotter.v1.nydus/snapshots/90/mnt
INFO[2025-12-05T17:43:42.757568156+01:00] Deleting resources [/containerd-local/io.containerd.snapshotter.v1.nydus/config/d4pgj4pjjoqfk878ge2g /containerd-local/io.containerd.snapshotter.v1.nydus/logs/d4pgj4pjjoqfk878ge2g /containerd-local/io.containerd.snapshotter.v1.nydus/socket/d4pgj4pjjoqfk878ge2g]
DEBU[2025-12-05T17:43:42.757710416+01:00] [cleanupUnusedCacheBlobs] found 0 used blob IDs
DEBU[2025-12-05T17:43:42.757734174+01:00] [cleanupUnusedCacheBlobs] found 1 stored blob IDs
INFO[2025-12-05T17:43:42.757741106+01:00] [Cleanup] unused cache blobs [0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f]
INFO[2025-12-05T17:43:42.757750627+01:00] remove cache sha256:0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f
DEBU[2025-12-05T17:43:42.757772881+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.chunk_map doest not exist.
DEBU[2025-12-05T17:43:42.757818030+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f doest not exist.
DEBU[2025-12-05T17:43:42.757956083+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.image.disk doest not exist.
DEBU[2025-12-05T17:43:42.757965258+01:00] file /containerd-local/io.containerd.snapshotter.v1.nydus/cache/0713669e321a795f8175192c6866928021e6a9ce832559d3315e35581dab340f.layer.disk doest not exist.
```

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -321,7 +321,7 @@ func (o *snapshotter) Cleanup(ctx context.Context) error {
 		}
 	}
 
-	cleanup, err = o.cleanupUnusedCacheBlobs(ctx)
+	cleanup, err = o.getUnusedCacheBlobs(ctx)
 	if err != nil {
 		return err
 	}
@@ -1253,7 +1253,7 @@ func (o *snapshotter) cleanupSnapshotDirectory(ctx context.Context, dir string) 
 	return nil
 }
 
-func (o *snapshotter) cleanupUnusedCacheBlobs(ctx context.Context) ([]string, error) {
+func (o *snapshotter) getUnusedCacheBlobs(ctx context.Context) ([]string, error) {
 	// Get a write transaction to ensure no other write transaction can be entered
 	// while the cleanup is scanning.
 	_, t, err := o.ms.TransactionContext(ctx, true)
@@ -1289,7 +1289,7 @@ func (o *snapshotter) cleanupUnusedCacheBlobs(ctx context.Context) ([]string, er
 		return nil, errors.Wrap(err, "walk managers to collect used blobs")
 	}
 
-	log.L.Debugf("[cleanupUnusedCacheBlobs] found %d used blob IDs", len(usedBlobIDs))
+	log.L.Debugf("[getUnusedCacheBlobs] found %d used blob IDs", len(usedBlobIDs))
 
 	// Update metrics for blobs in use
 	data.CacheBlobsInUse.Set(float64(len(usedBlobIDs)))
@@ -1314,13 +1314,13 @@ func (o *snapshotter) cleanupUnusedCacheBlobs(ctx context.Context) ([]string, er
 
 		blobID := cache.ExtractBlobIDFromFilename(entry.Name())
 		if blobID == "" {
-			log.L.Debugf("[cleanupUnusedCacheBlobs] skipping file with unknown format: %s", entry.Name())
+			log.L.Debugf("[getUnusedCacheBlobs] skipping file with unknown format: %s", entry.Name())
 			continue
 		}
 
 		storedBlobIDs[blobID] = nil
 	}
-	log.L.Debugf("[cleanupUnusedCacheBlobs] found %d stored blob IDs", len(storedBlobIDs))
+	log.L.Debugf("[getUnusedCacheBlobs] found %d stored blob IDs", len(storedBlobIDs))
 
 	var unusedBlobIDs []string
 	for blobID := range storedBlobIDs {


Fricounet:
@imeoer this PR is now ready for review, I've made a temporary fix to address the failing tests due to #712 while we address the issue long term

snapshot/snapshot.go
```diff
@@ -319,6 +320,23 @@ func (o *snapshotter) Cleanup(ctx context.Context) error {
...
 		}
 	}
+
+	cleanup, err = o.cleanupUnusedCacheBlobs(ctx)
+	if err != nil {
+		return err
+	}
+	log.L.Infof("[Cleanup] unused cache blobs %v", cleanup)
+
+	for _, blob := range cleanup {
```
BraveY, line 330: Deleting too many blobs may cost a lot time. I think we should add timeout handling here. How about use a select to listen ctx.Done/Ctx.err event in for loop?

snapshot/snapshot.go
```diff
@@ -319,6 +320,23 @@ func (o *snapshotter) Cleanup(ctx context.Context) error {
...
+		// Add sha256: prefix to make it a valid blob digest
+		if err := o.fs.RemoveCache("sha256:" + blob); err != nil {
```
BraveY, line 335: Maybe we could add small delay to avoid IO spike. Because we may delete a large number of files in a short period of time, causing excessive system load.

snapshot/snapshot.go
BraveY, line 305: Cleanup relies entirely on containerd calling `Cleanup()`. If containerd's GC is delayed or skipped, cache files accumulate. In high-churn environments, disk can fill up between cleanups. 
Maybe we should add a disk usage threshold, to trigger  GC early if cache exceeds this.

BraveY (changes requested):

BraveY:
@Fricounet  Thanks for the great PR. I leaved some comments about performance and risk.

snapshot/snapshot.go
Fricounet: @BraveY are you aware of specific situations where you've seen containerd's GC being skipped or delayed significantly?
From my observations, containerd will always call Cleanup once it has finished removing all the layers for an image and this happens at most a few seconds after the Remove calls.

I'm not a big fan of adding a disk usage threshold because:
1. it's a bit of an expensive operation to compute the disk usage
2. the kubelet already does this job and will remove images if it needs to reclaim space. I'd rather not have 2 competing thresholds

If you really think the snapshotter should proactively Cleanup behind itself, even if containerd doesn't call the Cleanup method, how about we create a simple goroutine when we initialize the snapshotter that calls Cleanup on a schedule to be determined by the administrator? But I haven't seen any case of containerd failing to do its job despite running this in a few thousand nodes so I'm not sure it's really needed at all :sweat_smile:

snapshot/snapshot.go
Fricounet: @BraveY Isn't it a bit counter productive? 
Deleting the files isn't gonna be a very expensive operation since it's only deleting the files metadata, not the actual bytes on the disk. Maybe the metadata operations can become an issue if we start deleting thousands of file at once but I don't see that happening in practice because images are only composed of a few dozens layers maximum and containerd will call a cleanup for every image it removes.

Looking at the new `nydus.snapshotter_cache_blobs_deleted.count` I've created in the PR, in my infrastructure, the maximum amount of blobs I've seen it trying to delete at once is 52, which I think is quite reasonable.

If anything, adding a sleep will massively lengthen the Cleanup operation if there are a lot of blobs to delete, potentially  worsening situations where we are under disk pressure and need to GC things quickly per your other comment https://github.com/containerd/nydus-snapshotter/pull/711#discussion_r2785700681

What do you think?

snapshot/snapshot.go
Fricounet: Good point, fixed it in [e952996](https://github.com/containerd/nydus-snapshotter/pull/711/commits/e952996ffd404118037ecfce8dd8cea99ef3e021)

Fricounet: Also, for information, from observations of the snapshotter CLEANUP operation durations with the prometheus metrics, I haven't seen a significant bump in the average duration for the snapshotter Cleanup since enabling this feature internally

snapshot/snapshot.go
BraveY: In our inner version, we add some delay by other people. But i have not observed the case that removing too many files caused some issues. So, it LGTM to keep your code original.

snapshot/snapshot.go
BraveY: I haven't encountered this either, just a bit worried that Containerd's GC might be lagging.
What you said makes sense, I think we can deal with it without adding goroutine first, and when there are really related problems in the future, we could fix it in a targeted manner.

Fricounet: sounds good thanks!

snapshot/snapshot.go
```diff
@@ -1241,6 +1253,85 @@ func (o *snapshotter) cleanupSnapshotDirectory(ctx context.Context, dir string)
 	return nil
 }
 
+func (o *snapshotter) cleanupUnusedCacheBlobs(ctx context.Context) ([]string, error) {
```
imeoer: May be rename the func to `getUnusedCacheBlobs`, others LGTM, thanks!

Fricounet: :+1: done in fc330cc

PR #734, opened 2026-04-07
fix: handle pod concurrency issues with daemon readiness check and bootstrap validation

akazi-bamfunds-oss:
## Overview
We see a concurrency issue under these conditions:
- Greater than 1 pod schedules onto a node at exactly the same time
- Both attempt to use the same image
- The image has not been pulled on the node before

This throws the following errors (or similar errors):
```
 [2026-04-09 15:41:15.200253 +00:00] ERROR [/api/src/error.rs:23] Error:
         "invalid Rafs v6 metadata size: bootstrap size 523264 is not aligned"
         at rafs/src/metadata/layout/v6.rs:171
         note: enable `RUST_BACKTRACE=1` env to display a backtrace
...
...
 [2026-04-09 15:41:15.327559 +00:00] ERROR [/api/src/error.rs:23] Error:
         "failed to get blob size, Request(ErrorWithMsg(\"\"))"
         at storage/src/meta/toc.rs:426
         note: enable `RUST_BACKTRACE=1` env to display a backtrace
 [2026-04-09 15:41:15.329073 +00:00] DEBUG [/src/metadata/mod.rs:782] failed to load inlined RAFS meta, I/O error (os error 5)
 [2026-04-09 15:41:15.329096 +00:00] ERROR [/service/src/fusedev.rs:753] service mount error: RAFS failed to handle request, Failed to fill superBlock: Invalid argument (os error 22)`
 ERROR [/api/src/error.rs:23] Error:
         Rafs(FillSuperBlock(Os { code: 22, kind: InvalidInput, message: "Invalid argument" }))
         at service/src/fusedev.rs:754
         note: enable `RUST_BACKTRACE=1` env to display a backtrace
 [2026-04-09 15:41:15.329132 +00:00] ERROR [src/bin/nydusd/main.rs:535] Failed in starting daemon:
 Error: Custom { kind: Other, error: "" }
```

This results in a zombie process, the daemon does not start successfully. Both pods fail with this error, as well as any subsequent pods which try to use the same image.
```
Error: failed to create containerd container: wait until daemon is RUNNING: get daemon state: daemon socket /var/lib/containerd/io.containerd.snapshotter.v1.nydus/socket/d7bsh6oob1lavl67ajtg/api.sock: not found
```

The root cause seems to be the snapshotter attempting to read the `image.boot`/`blob.meta` files before they are fully initialized, but only when multiple pods schedule at the same time.

This leaves behind a broken/failed snapshot which subsequent pods attempt to use since the rafs instance is persisted to cache regardless of if it is healthy or not.

## Related Issues
N/A

## Change Details
This PR provides two ways of preventing this issue:
- Prevents persistence of rafs instance to cache until the daemon has reached a `running` state
- Adds validation for bootstrap and `blob.meta` files during Mount() to ensure they are not read before the files are fully written/ready

The file validation checks the following:
- Bootstrap and `blob.meta` files exist and are readable
- RAFS version can be detected from the header
- Bootstrap size is aligned to block size
- State of both files are identical across two consecutive checks to ensure they are not still changing

## Test Results
Tested in large enterprise-scale AWS EKS cluster and promoted to production with thousands of pods scheduled daily. This issue no longer causes pod failures and instead logs the following and waits for readiness.
```
containerd-nydus-grpc[3490]: time="2026-04-14T15:41:13.357600693Z" level=warning msg="bootstrap is not ready, retrying... attempt: 1, error: bootstrap is not ready, current state: bootstrap=39370752;blobmeta=none"
```

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/pkg/filesystem/bootstrap.go
+++ b/pkg/filesystem/bootstrap.go
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2026. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package filesystem
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/containerd/log"
+	"github.com/containerd/nydus-snapshotter/pkg/layout"
+)
+
+const (
+	defaultBootstrapReadyAttempts = 20
+	defaultBootstrapReadyInterval = 50 * time.Millisecond
+	v6BlockBitsOffset             = layout.RafsV6SuperBlockOffset + 12
+	v6BlockBits512                = 9
+	v6BlockBits4096               = 12
+)
+
+func waitForReadyBootstrap(path string) error {
+	return waitForReadyBootstrapWithRetry(path, defaultBootstrapReadyAttempts, defaultBootstrapReadyInterval)
+}
+
+func waitForReadyBootstrapWithRetry(path string, attempts int, interval time.Duration) error {
+	var (
+		lastSize int64 = -1
+		lastErr  error
+	)
+
+	for idx := 0; idx < attempts; idx++ {
+		size, err := validateBootstrap(path)
+		if err == nil {
+			if size == lastSize {
+				return nil
+			}
+			lastSize = size
+			lastErr = fmt.Errorf("bootstrap size %d is still changing", size)
+		} else {
+			log.L.WithError(err).Warnf("bootstrap %s validation attempt %d/%d failed", path, idx+1, attempts)
+			lastErr = err
+		}
+
+		if idx+1 < attempts {
+			time.Sleep(interval)
+		}
+	}
+
+	if lastErr == nil {
+		lastErr = fmt.Errorf("bootstrap is not ready")
+	}
+
+	return fmt.Errorf("bootstrap %s is not ready: %w", path, lastErr)
+}
+
+func validateBootstrap(path string) (int64, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return 0, err
+	}
+	if !info.Mode().IsRegular() {
+		return 0, fmt.Errorf("bootstrap is not a regular file")
+	}
+
+	file, err := os.Open(path)
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+
+	headSize := int(info.Size())
+	if headSize > layout.MaxSuperBlockSize {
+		headSize = layout.MaxSuperBlockSize
+	}
+	if headSize < 8 {
+		return 0, fmt.Errorf("bootstrap is too small: %d", info.Size())
+	}
+
+	head := make([]byte, headSize)
+	if _, err := io.ReadFull(file, head); err != nil {
+		return 0, err
+	}
+
+	version, err := layout.DetectFsVersion(head)
+	if err != nil {
+		return 0, err
+	}
+
+	if version == layout.RafsV6 {
+		blockSize, err := detectV6BlockSize(head)
+		if err != nil {
+			return 0, err
+		}
+		if info.Size()%int64(blockSize) != 0 {
+			return 0, fmt.Errorf("v6 bootstrap size %d is not aligned to %d", info.Size(), blockSize)
+		}
+	}
+
+	return info.Size(), nil
+}
+
+func detectV6BlockSize(head []byte) (int, error) {
+	if len(head) <= int(v6BlockBitsOffset) {
+		return 0, fmt.Errorf("v6 bootstrap header is too small")
+	}
+
+	switch head[v6BlockBitsOffset] {
+	case v6BlockBits512:
+		return 512, nil
+	case v6BlockBits4096:
+		return 4096, nil
+	default:
+		return 0, fmt.Errorf("unknown v6 block bits %d", head[v6BlockBitsOffset])
+	}
+}
--- a/pkg/filesystem/bootstrap.go
+++ b/pkg/filesystem/bootstrap.go
@@ -53,13 +53,13 @@ func waitForReadyBootstrapWithRetry(path string, attempts int, interval time.Dur
 
 		return nil
 	},
-	retry.Attempts(uint(attempts)),
-	retry.Delay(interval),
-	retry.DelayType(retry.FixedDelay),
-	retry.LastErrorOnly(true),
-	retry.OnRetry(func(n uint, err error) {
-		log.L.Warnf("bootstrap is not ready, retrying... attempt: %d, error: %v", n+1, err)
-	}),
+		retry.Attempts(uint(attempts)),
+		retry.Delay(interval),
+		retry.DelayType(retry.FixedDelay),
+		retry.LastErrorOnly(true),
+		retry.OnRetry(func(n uint, err error) {
+			log.L.Warnf("bootstrap is not ready, retrying... attempt: %d, error: %v", n+1, err)
+		}),
 	)
 
 	if err != nil {
--- a/pkg/filesystem/bootstrap.go
+++ b/pkg/filesystem/bootstrap.go
@@ -10,18 +10,24 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/containerd/log"
 	"github.com/containerd/nydus-snapshotter/pkg/layout"
+	"github.com/containerd/nydus-snapshotter/pkg/utils/retry"
 )
 
 const (
 	defaultBootstrapReadyAttempts = 20
 	defaultBootstrapReadyInterval = 50 * time.Millisecond
-	v6BlockBitsOffset             = layout.RafsV6SuperBlockOffset + 12
-	v6BlockBits512                = 9
-	v6BlockBits4096               = 12
+
+	v6BlockBitsOffset = layout.RafsV6SuperBlockOffset + 12
+	v6BlockBits512    = 9
+	v6BlockBits4096   = 12
 )
 
 func waitForReadyBootstrap(path string) error {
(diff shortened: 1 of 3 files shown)


pkg/filesystem/fs.go
```diff
@@ -249,13 +250,39 @@ func (fs *Filesystem) TryStopSharedDaemon() {
...
-			return nil
+	const (
+		attempts = 20
+		interval = 100 * time.Millisecond
+	)
+
+	var rafs *racache.Rafs
+	ready := false
+
+	for idx := 0; idx < attempts; idx++ {
```
Fricounet: can't you use the `retry` lib in `pkg/utils/retry`? This would be cleaner than this for loop

Fricounet: also, should we consider a backoff?

pkg/filesystem/fs.go
```diff
@@ -323,8 +350,16 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
 	if rafs != nil {
-		// Instance already exists, how could this happen? Can containerd handle this case?
-		return nil
+		switch rafs.State {
+		case racache.RafsStateReady:
+			// Instance already exists, how could this happen? Can containerd handle this case?
+			return nil
+		case racache.RafsStateFailed:
+			// previous mount attempt failed, clean up and try again
+			racache.RafsGlobalCache.Remove(snapshotID)
```
Fricounet: should we return after that?

pkg/filesystem/bootstrap.go
```diff
@@ -0,0 +1,121 @@
...
+	return waitForReadyBootstrapWithRetry(path, defaultBootstrapReadyAttempts, defaultBootstrapReadyInterval)
+}
+
+func waitForReadyBootstrapWithRetry(path string, attempts int, interval time.Duration) error {
+	var (
+		lastSize int64 = -1
+		lastErr  error
+	)
+
+	for idx := 0; idx < attempts; idx++ {
```
Fricounet: same, please use a retry lib for this

Fricounet (commented):
Thanks for the PR, can you clarify in the description what is the issue you are trying to solve exactly? And if you made sure this indeed solved the issue?

akazi-bamfunds-oss:
@Fricounet apologies - this was supposed to be a draft PR. I will resubmit with proper description when ready!

pkg/filesystem/fs.go
akazi-bamfunds-oss: Switched to `retry` lib - thanks for the suggestion!

pkg/filesystem/fs.go
akazi-bamfunds-oss: Removed the rafs state field as not required for minimal fix

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: Switched to retry lib

akazi-bamfunds-oss:
@Fricounet / @imeoer ready for review when you guys have time!

pkg/filesystem/bootstrap.go
```diff
@@ -0,0 +1,200 @@
...
+		lastState *string
+	)
+
+	err := retry.Do(func() error {
+		state, err := validateBootstrapAndBlobMeta(path)
+		if err != nil {
+			return err
+		}
+
+		// Require two consecutive "ready" states to reading unready bootstrap
```
Fricounet: why do you need to have 2 "ready" to cosider the bootstrap read?

pkg/filesystem/bootstrap.go
```diff
@@ -0,0 +1,200 @@
...
+
+		return nil
+	},
+		retry.Attempts(uint(attempts)),
+		retry.Delay(interval),
+		retry.DelayType(retry.FixedDelay),
+		retry.LastErrorOnly(true),
+		retry.OnRetry(func(n uint, err error) {
+			log.L.Warnf("bootstrap is not ready, retrying... attempt: %d, error: %v", n+1, err)
+		}),
```
Fricounet, line 89: this might result in a large amount of logs no?
Maybe use `Debug` instead?

pkg/filesystem/bootstrap.go
```diff
@@ -0,0 +1,200 @@
...
+	if !info.Mode().IsRegular() {
+		return 0, fmt.Errorf("bootstrap is not a regular file")
+	}
+
+	size := info.Size()
+	headSize := int(size)
+	if headSize > layout.MaxSuperBlockSize {
+		headSize = layout.MaxSuperBlockSize
+	}
+	if headSize < 8 {
```
Fricounet, line 136: why 8? Consider using an explicit const to clarify?

pkg/filesystem/fs.go
```diff
@@ -371,6 +371,9 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
 		if err != nil {
 			return errors.Wrapf(err, "find bootstrap file snapshot %s", snapshotID)
 		}
+		if err := waitForReadyBootstrap(bootstrap); err != nil {
+			return errors.Wrapf(err, "wait for bootstrap file snapshot %s", snapshotID)
+		}
```
Fricounet, line 376: if this fails, do we cleanup properly? Did you test if the next call will recover correctly?

Fricounet (commented):
a few things to clarify but mostly LGTM

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: The 2 consecutive ready states ensure the bootstrap file is not still being written. The main symptom is that the file is read before it has actually finished writing

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: In my experience, this is usually only 2-3 log lines for a new snapshotter and only really logs when pods are scheduled concurrently. But happy to move to a debug log instead.

Fricounet: ok it should be fine if it's only this much, we can always revisit later if needed

pkg/filesystem/bootstrap.go
Fricounet: ah ok I see! I hadn't properly caught that in this case `state` is the string you built with `bootstrap=%d;%s`.
Would it make sense to use a struct instead of a string? Because rereading through `validateBlobMetaFiles` I find it a bit hard to follow what the `blobMetaState` string is supposed to look like. having a struct and comparing the old/new one could be a bit clearer I think. What do you think?

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: @imeoer wrote this part but I believe it's the minimum bytes to be able to call DetectFsVersion(). It's just an early failure check by the looks of it - happy to move this to a const

pkg/filesystem/fs.go
akazi-bamfunds-oss: Since we're validating bootstrap before the daemon is created or anything is added to db cache, there is nothing to clean up here. The deferred cleanup only needs to remove the in-memory cache entry which umount() handles

pkg/filesystem/bootstrap.go
Fricounet: Thanks for the explanation, I'll let @imeoer chime in to get his opinion on this

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: Thanks for the suggestion! I agree, have made changes to switch to structs - should be more readable now

Fricounet (approved):
LGTM! Thanks a lot for this very good PR!
I'll let a bit more time for @imeoer to chime in if he wants to and we'll merge this next week I think.

BraveY (approved):
LGTM! Thanks

pkg/filesystem/bootstrap.go
imeoer: Thanks, we'd better reuse the codes in `snapshot/mount_option.go` to avoid reimplementing the metadata check logic.

pkg/filesystem/bootstrap.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, nil
+	}
+
+	files := make([]blobMetaState, 0, len(matches))
+	for _, p := range matches {
+		file, err := os.Open(p)
+		if err != nil {
+			return nil, err
+		}
+
```
imeoer, line 184: We can introduce an internal `validateBlobMetaFile` func to simplify the code by using `defer file.Close()`.

pkg/filesystem/bootstrap.go
```diff
@@ -0,0 +1,226 @@
...
+			file.Close()
+			return nil, fmt.Errorf("blob.meta %s is not a regular file", p)
+		}
+		if info.Size() <= 0 {
+			file.Close()
+			return nil, fmt.Errorf("blob.meta %s is empty", p)
+		}
+
+		var buf [1]byte
+		if _, err := io.ReadFull(file, buf[:]); err != nil {
```
imeoer: Is the purpose of reading one byte here to verify the download integrity of the `.blob.meta` file? In production environment, could both `.blob.meta` and `bootstrap` files become incomplete under concurrent access?

akazi-bamfunds-oss: Yes, I observed when testing, even if we validate the bootstrap file, the `blob.meta` file can also become corrupted.

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: Thank you for the suggestion - I have implemented it now

pkg/filesystem/bootstrap.go
akazi-bamfunds-oss: mount_option.go doesn't cover the file validation + size checks + blob.meta validation unfortunately so it doesn't cover the current usecases

akazi-bamfunds-oss:
@imeoer / @Fricounet Happy to merge this in? 😄 Unsure if the failing tests are relevant - they seem unrelated to my changes

Fricounet (approved):
LGTM, sorry for the delay in merging this!

PR #738, opened 2026-04-15
[snapshot] Add IDMapping support for user namespace remapping

wangrzneu:
Add IDMapping support for user namespace remapping.


snapshot/utils_test.go
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) 2025. Nydus Developers. All rights reserved.
```
Fricounet:
```suggestion
 * Copyright (c) 2026. Nydus Developers. All rights reserved.
```

snapshot/snapshot.go
```diff
@@ -1022,19 +1064,6 @@ func (o *snapshotter) mergeTarfs(ctx context.Context, s storage.Snapshot, pID st
...
-		{
-			Type:   "bind",
-			Source: source,
-			Options: []string{
-				roFlag,
-				"rbind",
-			},
-		},
-	}
-}
```
Fricounet, line 1036: why did you remove this?

snapshot/snapshot.go
```diff
@@ -858,14 +873,41 @@ func (o *snapshotter) createSnapshotWithRecovery(ctx context.Context, kind snaps
...
+			return nil, storage.Snapshot{}, errors.New("incompatible types after stat call: *syscall.Stat_t expected")
+		}
+		if mappedUID == -1 {
+			mappedUID = int(stat.Uid)
+		}
+		if mappedGID == -1 {
+			mappedGID = int(stat.Gid)
+		}
+	}
+	if mappedUID != -1 && mappedGID != -1 {
```
Fricounet: is this condition necessary?

Fricounet (commented):
Thanks for the PR, I have a few questions?

snapshot/utils_test.go
wangrzneu: I fixed it.

snapshot/snapshot.go
wangrzneu: bindMount was already unused on main — it had zero call sites. With the IDMapping changes, bind mounts are now constructed inline in mountNative to support appending idmapOptions. Keeping it would cause a lint failure (unused function), so the removal is intentional.

snapshot/snapshot.go
wangrzneu: Yes, this condition is necessary. When len(s.ParentIDs) == 0 (no parent snapshot) and no IDMapping labels are present, both mappedUID and mappedGID remain at their initial value of -1. Without this guard, Lchown would be called with -1 on non-idmapped snapshots, which is incorrect. The condition ensures we only chown when a valid mapped UID/GID has been determined — either from labels or inherited from the parent.

snapshot/snapshot.go
Fricounet: Is that an LLM generated response? 
> bindMount was already unused on main

This is false? You removed the call sites [yourself](https://github.com/containerd/nydus-snapshotter/pull/738/changes#diff-5c6bcd0ff5f76d3d6e311122b52e890cff76233ee29e88a138f074bcc74518e1L1164) in the PR. If you prefer my question is why did you remove the call sites?

snapshot/snapshot.go
Fricounet: My concern is that [previously](https://github.com/containerd/nydus-snapshotter/pull/738/changes#diff-5c6bcd0ff5f76d3d6e311122b52e890cff76233ee29e88a138f074bcc74518e1L868) we would run chown in all cases. Now, you are telling me that it might not be run in some cases. That's a behavioral change.
I think we should keep a behavior where lchown is called in all cases.

Fricounet: or if that's factually impossible for mappedUID and mappedGID to both remain at -1, then I think it should be clearly stated as a comment in the code

snapshot/snapshot.go
wangrzneu: Sorry about the incorrect previous response. You're right — bindMount had two call sites in mountNative that I removed in this PR. The reason was that the original bindMount(source, roFlag) signature couldn't pass idmapOptions. I've now restored bindMount with a variadic signature bindMount(source, roFlag string, extraOpts ...string) so it can carry the idmap mount options without inlining. See commit 74a8cac.

snapshot/snapshot.go
wangrzneu: I've refactored the chown logic into two exclusive branches so the original behavior is completely preserved

snapshot/utils.go
```diff
@@ -7,17 +7,39 @@
...
 	"syscall"
 
 	"github.com/containerd/continuity/fs"
+	"github.com/pkg/errors"
 )
 
+func lchown(target string, st os.FileInfo) error {
+	stat := st.Sys().(*syscall.Stat_t)
+	return os.Lchown(target, int(stat.Uid), int(stat.Gid))
+}
```
Fricounet, line 21: can you move this back to the original position so it doesn't create a diff?

snapshot/utils.go
```diff
@@ -7,17 +7,39 @@
...
+	)
+	if _, err := fmt.Sscanf(mapping, "%d:%d:%d", &ctrID, &hostID, &length); err != nil {
+		return -1, errors.Wrapf(err, "failed to parse ID mapping %q", mapping)
+	}
+	if ctrID < 0 || hostID < 0 || length <= 0 {
+		return -1, errors.Errorf("invalid mapping %q: IDs must be non-negative and size must be positive", mapping)
+	}
+	if ctrID != 0 {
+		return -1, errors.Errorf("only container ID 0 is supported in ID mapping, got %d", ctrID)
+	}
```
Fricounet, line 43: not familiar with this feature, why only 0 is supported?

snapshot/snapshot.go
```diff
@@ -858,13 +872,34 @@ func (o *snapshotter) createSnapshotWithRecovery(ctx context.Context, kind snaps
...
+			mappedUID = uid
+		}
+		if hasGID {
+			gid, err := parseIDMappingHostID(gidMapping)
+			if err != nil {
+				return nil, storage.Snapshot{}, errors.Wrapf(err, "invalid GID mapping %q", gidMapping)
+			}
+			mappedGID = gid
+		}
+		if err := os.Lchown(filepath.Join(td, "fs"), mappedUID, mappedGID); err != nil {
```
Fricounet, line 895: if `hasUID` or `hasGID` evaluates to `false` (the snapshot is missing one of gid or uid label), then one of `mappedUID` or `mappedGID` will remain equal to -1. I think the chown will fail in this case no?

imeoer (commented):
Thanks for the idmapping support! Overall LGTM.

markasoftware-tc:
How does this work, since the underlying nydus rafs/etc filesystem does not have idmap support? Is containerd internally falling back to some alternate method when you tell it to uidmap/gidmap the mount but the underlying fs doesn't support it?

PR #719, opened 2026-03-17
[Feat] Add helm charts for nydus-snapshotter for convenient K8s deploytment 

Zephyrcf:
## Overview
_Please briefly describe the changes your pull request makes._

## Related Issues
_Please link to the relevant issue. For example: `Fix #123` or `Related #456`._
https://github.com/containerd/nydus-snapshotter/issues/717
## Change Details
_Please describe your changes in detail:_

## Test Results
_If you have any relevant screenshots or videos that can help illustrate your changes, please add them here._

## Change Type
_Please select the type of change your pull request relates to:_
- [ ] Bug Fix
- [x] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist
_Before submitting a pull request, please ensure you have completed the following:_
- [ ] I have run a code style check and addressed any warnings/errors.
- [ ] I have added appropriate comments to my code (if applicable).
- [ ] I have updated the documentation (if applicable).
- [ ] I have written appropriate unit tests.

PR #749, opened 2026-04-30
ci: bump actions/checkout from v4 to v6 (#724)

mvanhorn:
Refs #724.

## What changed

Bumped \`actions/checkout\` from \`@v4\` to \`@v6\` across all 13 occurrences in the five workflow files:

| Workflow | Jobs touched |
|---|---|
| ci.yml | 6 |
| e2e.yml | 2 |
| k8s-e2e.yml | 1 |
| optimizer.yml | 1 |
| release.yml | 3 |

## Why this matters

actions/checkout@v4 runs on Node.js 20, which GitHub flagged as deprecated in the warning quoted in #724:

> Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026.

actions/checkout@v6 runs on Node.js 24 and is the current stable major. Bumping it now lets the CI surface real failures (if any) before the forced cutover instead of hitting them at the deadline.

## Out of scope

The issue also lists \`azure/setup-helm@v4\`, but that action is not currently used by any workflow in main. It's coming in via the open Helm chart PR (#719). When that PR lands, the same v4→v5 bump can be applied there or as a focused follow-up — keeping it out of this PR avoids cross-branch entanglement.

This PR also leaves the v4-pinned \`actions/cache\`, \`actions/upload-artifact\`, and \`actions/download-artifact\` alone. Those are still on Node.js 20 and will need their own bump when stable v5 majors land, but they're a different upstream-driven cadence than checkout's release schedule.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Check out code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
       - name: Setup Golang
         uses: actions/setup-go@v5
         with:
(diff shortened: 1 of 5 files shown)


imeoer (approved):
LGTM, thanks!

PR #751, opened 2026-05-11
[Bug] converter: skip content store lookup for blob descs when backend is set

Tan0:
## Overview

When a storage backend (e.g. S3) is configured together with `--chunk-dict`, `MergeLayers` unconditionally iterates `blobDigests` and calls `cs.Info()` to build `blobDescs` from the local content store.

When a backend is active, blobs are pushed directly to the backend and are **not** retained in the local content store. If `--chunk-dict` is also in use, the dict blobs (from the base image) only exist in the backend, causing `cs.Info()` to return:

```
failed to get chunk info: content digest sha256:xxxx not found
```

This makes it impossible to use `--chunk-dict` and `--backend-type s3` together, even though both features work correctly in isolation.

## Root Cause

`pkg/converter/convert_unix.go` — `MergeLayers()`, the `blobDigests` loop:

```go
for idx, blobDigest := range blobDigests {
    blobInfo, err := cs.Info(ctx, blobDigest)  // always queries local store
    ...
}
```

This loop runs regardless of whether `opt.Backend != nil`. When a backend is set, `convertManifest` only uses `blobDescs` when `opt.Backend == nil`. However, building `blobDescs` is still needed in the backend path for annotation/metadata purposes, and `cs.Info()` fails when chunk-dict blobs only exist in the remote backend rather than the local content store.

## Fix

1. Add `Size(blobDigest digest.Digest) (int64, error)` to the `Backend` interface to query blob size from remote storage.
2. Implement `Size()` for all three backends: **OSS** (`GetObjectMeta`), **S3** (`HeadObject`), **LocalFS** (`os.Stat`).
3. In `MergeLayers`, when `opt.Backend != nil`, use `opt.Backend.Size()` to get blob size instead of `cs.Info()`. When `opt.Backend == nil`, continue using `cs.Info()` as before.

This way `blobDescs` is always built correctly regardless of whether blobs live in the local content store or a remote backend.

## Change Details

- `pkg/backend/backend.go`: add `Size()` method to `Backend` interface
- `pkg/converter/types.go`: add `Size()` method to converter `Backend` interface (mirror)
- `pkg/backend/oss.go`: implement `OSSBackend.Size()` via `GetObjectMeta`
- `pkg/backend/s3.go`: implement `S3Backend.Size()` via `HeadObject`
- `pkg/backend/localfs.go`: implement `LocalFSBackend.Size()` via `os.Stat`
- `pkg/converter/convert_unix.go`: use `Backend.Size()` when backend is set, `cs.Info()` otherwise

## Test Results

Verified with nydusify v2.4.1 + S3 backend + chunk-dict (5 representative base images covering ~300 production images):

| | Without chunk-dict (baseline) | With chunk-dict |
|---|---|---|
| S3 storage | 94.8 GiB | 19.2 GiB |
| Blob count | 1808 | 281 |
| Dedup ratio | — | **79.7%** |

Before this fix: conversion fails immediately with `content digest not found`.
After this fix: conversion succeeds, deduplication rate 75.7%–79.7%.

## Change Type

- [x] Bug Fix

## Self-Checklist

### DIFF
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -1297,15 +1297,15 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 		var err error
 		if opt.Backend != nil {
 			blobSize, err = opt.Backend.Size(blobDigest)
+			if err != nil {
+				return nil, nil, errors.Wrap(err, "get blob size from backend")
+			}
 		} else {
 			blobInfo, err := cs.Info(ctx, blobDigest)
 			if err != nil {
 				return nil, nil, errors.Wrap(err, "get info from content store")
 			}
 			blobSize = blobInfo.Size
-		}
-		if err != nil {
-			return nil, nil, errors.Wrap(err, "get blob size")
 		}
 		blobDesc := ocispec.Descriptor{
 			Digest:    blobDigest,
--- a/pkg/converter/convert_unix.go
+++ b/pkg/converter/convert_unix.go
@@ -1285,44 +1285,46 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
 	originalBlobDigests := <-originalBlobDigestChan
 	blobDescs := []ocispec.Descriptor{}
 
-	// When a storage backend is configured, blobs are pushed directly to the
-	// backend and are not retained in the local content store. In convertManifest,
-	// blobDescs is only consumed when opt.Backend == nil (manifest.Layers is set
-	// to []ocispec.Descriptor{*bootstrapDesc} when a backend is active). Skip
-	// building blobDescs to avoid spurious "content digest not found" errors,
-	// which occur when chunk-dict blobs exist only in the backend and not locally.
-	if opt.Backend == nil {
-		var blobDigests []digest.Digest
-		if opt.OCIRef {
-			blobDigests = nydusBlobDigests
-		} else {
-			blobDigests = mergeManifestBlobDigests(nydusBlobDigests, originalBlobDigests)
-		}
+	var blobDigests []digest.Digest
+	if opt.OCIRef {
+		blobDigests = nydusBlobDigests
+	} else {
+		blobDigests = mergeManifestBlobDigests(nydusBlobDigests, originalBlobDigests)
+	}
 
-		for idx, blobDigest := range blobDigests {
+	for idx, blobDigest := range blobDigests {
+		var blobSize int64
+		var err error
+		if opt.Backend != nil {
+			blobSize, err = opt.Backend.Size(blobDigest)
+		} else {
 			blobInfo, err := cs.Info(ctx, blobDigest)
 			if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/converter/convert_unix.go
```diff
@@ -1285,36 +1285,44 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
...
+			blobDigests = nydusBlobDigests
+		} else {
+			blobDigests = mergeManifestBlobDigests(nydusBlobDigests, originalBlobDigests)
 		}
 
-		if opt.Encrypt != nil {
-			blobDesc.Annotations[LayerAnnotationNydusEncryptedBlob] = "true"
-		}
+		for idx, blobDigest := range blobDigests {
+			blobInfo, err := cs.Info(ctx, blobDigest)
```
imeoer, line 1304: Thanks for the PR! Can we get the size info from backend here (if `opt.Backend != nil`)?
I’m concerned that blobDescs might be used externally by Merge.

Needs to add a `Size(blobDigest digest.Digest) (int64, error)` implementation in `type Backend interface`, oss example:

```golang
func (b *OSSBackend) Size(blobDigest digest.Digest) (int64, error) {
	blobID := blobDigest.Hex()
	blobObjectKey := b.objectPrefix + blobID
	headers, err := b.bucket.GetObjectMeta(blobObjectKey)
	if err != nil {
		return 0, errors.Wrap(err, "get object size")
	}
	sizeStr := headers.Get("Content-Length")
	size, err := strconv.ParseInt(sizeStr, 10, 0)
	if err != nil {
		return 0, errors.Wrap(err, "parse content-length header")
	}
	return size, nil
}
```

Tan0:
Good suggestion. Implemented `Backend.Size()` for all three backends (S3/OSS/LocalFS) and updated `MergeLayers` to use it when a backend is configured. See the latest two commits for details.

BraveY (approved):
LGTM.

pkg/converter/convert_unix.go
```diff
@@ -1293,13 +1293,23 @@ func MergeLayers(ctx context.Context, cs content.Store, descs []ocispec.Descript
...
 	for idx, blobDigest := range blobDigests {
-		blobInfo, err := cs.Info(ctx, blobDigest)
+		var blobSize int64
+		var err error
+		if opt.Backend != nil {
+			blobSize, err = opt.Backend.Size(blobDigest)
+		} else {
+			blobInfo, err := cs.Info(ctx, blobDigest)
+			if err != nil {
+				return nil, nil, errors.Wrap(err, "get info from content store")
```
imeoer, line 1306: The error message need be changed for backend, others LGTM. :)

imeoer (approved):

pkg/converter/convert_unix.go
Tan0: Fixed. Separated the backend error handling so the backend path now returns `get blob size from backend` in 5858f9c.

imeoer (approved):
Thanks!

PR #753, opened 2026-05-19
[daemonconfig] Write config file atomically

Fricounet:
## Overview
`DumpConfigFile` writes the daemon JSON config via `os.WriteFile`, which truncates the target before writing the new contents. If the snapshotter is killed mid-write, or if a concurrent writer races on the same path, the file can be left empty or partial, breaking subsequent loads by the snapshotter or by nydusd.

This PR switches the write to a tmp-file + `rename(2)` pattern so the swap is atomic: readers always observe either the previous or the new contents.

## Related Issues
N/A

## Change Details
- Write the JSON to a tmp file created via `os.CreateTemp` in the same directory as the target.
- `os.Rename` over the target. Same-filesystem rename is atomic on Linux.
- On error, the tmp file is removed by a deferred cleanup.
- `os.CreateTemp` creates the tmp file with mode `0600`, matching the previous permissions.

## Test Results
- `make build` passes.
- `go test ./config/daemonconfig/... ./pkg/daemon/...` passes.
- `go vet` passes.

## Change Type
- [x] Bug Fix

## Self-Checklist

### DIFF
--- a/config/daemonconfig/daemonconfig.go
+++ b/config/daemonconfig/daemonconfig.go
@@ -127,11 +127,6 @@ type DeviceConfig struct {
 // For nydusd as FUSE daemon. Serialize Daemon info and persist to a json file
 // We don't have to persist configuration file for fscache since its configuration
 // is passed through HTTP API.
-//
-// The write is atomic: contents are written to a tmp file in the same directory
-// and renamed over the target. A concurrent reader, or a reader after the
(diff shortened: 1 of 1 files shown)


imeoer (approved):
LGTM, maybe wrap it like func `atomicWriteFile`.

Fricounet:
good point thanks @imeoer fixed in a65f101

imeoer (approved):
LGTM!

PR #755, opened 2026-06-08
fix: build release binaries statically to avoid glibc relocation errors

### LINKED ISSUE #748: [Bug] all 0.15.x images are broken due to incorrect glibc
### Problem Description

I use nydus-snapshotter as a pod/container in my k8s environment. As per Docker philosophy (and in general common sense), any binaries in an image should be runnable. But not in case of nydus-snapshotter, where this is completely wrong 

Moreover, the binary itself is correct and works on the host machine.

```
$ uname -a; lsb_release -a; docker cp nydus-snapshotter-issue:/opt/nydus-artifacts ./todelete; ./todelete/usr/local/bin/containerd-nydus-grpc --version; ldd --version
Linux avi-gn-fsk42 6.1.0-32-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.129-1 (2025-03-06) x86_64 GNU/Linux
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm
Successfully copied 127MB to /home/user/todelete
Version:     v0.15.15
Revision:    c2f1c15f47198d62261a2fdd23d8b1635f3f5375
Go version:  go1.25.9
Build time:  2026-04-17T14:53:18
ldd (Debian GLIBC 2.36-9+deb12u13) 2.36
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

### Expected Behavior

The binary should be runnable/work in container environment

### Actual Behavior

```
~$ docker run --name nydus-snapshotter-issue -it --rm ghcr.io/containerd/nydus-snapshotter:v0.15.15   sh -c '/bin/bash'
db828be2ab00:~# ldd /opt/nydus-artifacts/usr/local/bin/containerd-nydus-grpc
	/lib64/ld-linux-x86-64.so.2 (0x7f7bd1c7e000)
	libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f7bd1c7e000)
Error relocating /opt/nydus-artifacts/usr/local/bin/containerd-nydus-grpc: __vfprintf_chk: symbol not found
Error relocating /opt/nydus-artifacts/usr/local/bin/containerd-nydus-grpc: __fprintf_chk: symbol not found
```
```
# ldd --version
musl libc (x86_64)
Version 1.2.3
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
```

### How to reproduce

1. docker run --name nydus-snapshotter-issue -it --rm ghcr.io/containerd/nydus-snapshotter:v0.15.15   sh -c '/opt/nydus-artifacts/usr/local/bin/containerd-nydus-grpc --version'

### Environment Details

- Nydus-snapshotter version: any 0.15.x
- Nydus version:
- Container runtime: Docker, containerd - doesn't matter
- Operating System: Debian 12 (doesn't matter)
- Kernel version:

### Additional Information

As I understand it, according to the project's current design approach, the image is merely a delivery mechanism for the binary, and glibc inside the container is only required to run the shell script that copies the binary to the host. That said, we shouldn't overlook the option to run natively inside containers without extracting to the host.

mvanhorn:
## Overview

Builds the per-arch release binaries with `CGO_ENABLED=0` so the published 0.15.x images, which run on a musl/alpine base, no longer fail with glibc `__vfprintf_chk` relocation errors.

## Related Issues

Fixes #748

## Change Details

The release workflow built the per-arch `make package` targets with cgo enabled, producing glibc-dynamically-linked binaries. Run on the musl/alpine image base, the dynamic loader fails on glibc-specific relocations such as `__vfprintf_chk`. Setting `CGO_ENABLED=0` for the `make package` step produces statically-linked, libc-independent binaries (the same approach the existing `static-package` matrix arm already uses), so the images run on alpine. @imeoer diagnosed this root cause and invited the fix.

## Test Results

This is a release-workflow change exercised by the release pipeline. The fix mirrors the existing `static-package` arm's static-linking approach, so the produced binaries no longer depend on glibc at runtime.

## Change Type

- [x] Bug Fix

## Self-Checklist

- [x] I have run a code style check and addressed any warnings/errors (single-line workflow edit)
- [x] I have added appropriate comments to my code (n/a, one-line build-flag change)
- [x] I have updated the documentation (n/a, no user-facing docs affected)
- [x] I have written appropriate unit tests (n/a, a CI build-flag change is exercised by the release pipeline, not unit tests)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,7 +71,7 @@ jobs:
           if [ "${{ matrix.build-arch }}" == "static" ]; then
             make static-package
           else
-            make package GOOS=${{ matrix.build-os }} GOARCH=${{ matrix.build-arch }}
+            CGO_ENABLED=0 make package GOOS=${{ matrix.build-os }} GOARCH=${{ matrix.build-arch }}
           fi
       - name: upload artifacts
         uses: actions/upload-artifact@v4


imeoer (approved):
LGTM, thanks!

mvanhorn:
Much appreciated, @imeoer. Static release binaries sidestep the glibc relocation headaches on older hosts.

Issue #740, opened 2026-04-17
[Bug] cleanup branch of misc/snapshotter/snapshotter.sh is not so good


wangqu123456:
### Problem Description

After the deploy process is completed, the sleep infinity action will create a child process to block the main process and keep the container running. If a cleanup operation is performed later, which actually involves restarting the containerd service, kubelet will disconnect from the container, causing the PreStop service to fail every time, and the forced kill signal sent by kubelet will not be received at this time.

### Expected Behavior

The preStop process needs to execute successfully and restore the containerd configuration file. This will make nydus-snapshotter more robust and ensure a stable restart and recovery process.

### Actual Behavior

The preStop process failed due to residual containerd configuration files and incomplete cleanup of nydus files, preventing the proper uninstallation process.

### How to reproduce

1. just deploy in k8s cluster，and then delete nydus-snapshotter，you can see files remain  on the node, and error logs in kubelet logs

### Environment Details

- Nydus-snapshotter version: latest
- Nydus version:
- Container runtime: containerd
- Operating System:
- Kernel version:

### Additional Information

I recommend against directly restarting containerd during the preStop process, as kubelet is much more sensitive to container connectivity at this stage than during creation. Instead, it's better to use a signal mechanism. Use `trap` in the deploy main process to catch the Term signal, and then restore the containerd configuration and restart it within that function.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

imeoer:
Thanks for the report, could you submit a PR to address this? :)

PR #757, opened 2026-06-11
feature nydus-snapshotter startup script robustness 

wangqu123456:
## Overview
recover some problems when using nydus-snapshotter in k8s scenario.

## Related Issues
https://github.com/containerd/nydus-snapshotter/issues/740

## Change Details
1. use trap to recover config.toml backup file
2. use format log function
3. feature configurations of config.toml in containerd 2.* version 3.
4. add fetch sandbox image step.

## Test Results

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [x] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/misc/snapshotter/snapshotter.sh
+++ b/misc/snapshotter/snapshotter.sh
@@ -40,15 +40,24 @@ ENABLE_SYSTEMD_SERVICE="${ENABLE_SYSTEMD_SERVICE:-false}"
 
 COMMANDLINE=""
 
+# Log function with timestamp
+log() {
+    local level="${1:-INFO}"
+    local msg="$2"
+    local timestamp
+    timestamp=$(date '+%Y-%m-%d %H:%M:%S')
+    echo "[$timestamp] [$level] $msg"
+}
+
 # If we fail for any reason a message will be displayed
 die() {
     msg="$*"
-    echo "ERROR: $msg" >&2
+    log "ERROR" "$msg" >&2
     exit 1
 }
 
 print_usage() {
-    echo "Usage: $0 [deploy/cleanup]"
+    log "INFO" "Usage: $0 [deploy/cleanup]"
 }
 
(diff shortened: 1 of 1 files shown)


imeoer:
This needs a fair bit of work before it can go in.

Real bugs:

- `deploy_pid=$(ps aux | grep "sleep infinity" | grep -v grep | awk '{print $1}')` — with `ps aux`, `$1` is the user, not the pid. The `kill -TERM` right after can never work.
- `trap cleanup_on_term TERM INT EXIT` fires on *every* exit, including `die`'s `exit 1` and the `cleanup` path's `exit 0`. So a failed deploy now restores the containerd config and restarts containerd on the host as a side effect.
- `cleanup_snapshotter` dropped the `kill -9 $pid` for the non-systemd case. The binaries get removed but `containerd-nydus-grpc` keeps running.
- `get_container_runtime` no longer calls `die` on a bad node name, it silently falls back to `containerd`. That turns a clear failure into a wrong assumption. The new `CONTAINER_RUNTIME_OVERRIDE` env var is undocumented as well.
- `systemctl restart --no-block` plus a fixed `sleep 5`: `is-active` can report the old instance as active before the restart has even begun.

Please also drop the Chinese comments (`# 等待服务启动`, `# 重启containerd服务会导致容器断联...`, `# 尝试从环境变量或预设值获取`, `# 尝试 kubectl`) — the repo is English only.

Last thing: this is really five changes in one — log formatting, containerd v3 config support, the trap rework, pause image prefetch, and the cleanup rewrite — plus a lot of whitespace churn. That makes it very hard to review as a unit. Could you split it up? The v3 config support alone deserves its own PR. The commit message also needs a squash and a conventional-commit subject.

PR #758, opened 2026-06-12
feat: support stream_prefetch in fs_prefetch config

81981266:
## What this PR does

Adds the `stream_prefetch` field to the FUSE daemon config's `FSPrefetch`
(`fs_prefetch`) struct.

## Why

nydusd's `FsPrefetchControl` config accepts a `stream_prefetch` field
(introduced in nydus v2.4.3). The snapshotter unmarshals the user-supplied
daemon config into `FSPrefetch` and re-serializes it before handing it to
nydusd. Since `FSPrefetch` had no field for `stream_prefetch`, the key was
silently dropped during the unmarshal/marshal round-trip, so users who set
`stream_prefetch` in their config saw it have no effect.

## Changes

- Add `StreamPrefetch bool` (json: `stream_prefetch`, `omitempty`) to `FSPrefetch`.
- Cover the field in `TestLoadConfig`.

`omitempty` keeps the output unchanged for configs that don't set the field,
so this is backward compatible with older nydusd.

## How to verify

Set `"stream_prefetch": true` under `fs_prefetch` in the nydusd config and
confirm it now propagates to the config nydusd actually receives.

### DIFF
--- a/config/daemonconfig/daemonconfig_test.go
+++ b/config/daemonconfig/daemonconfig_test.go
@@ -50,7 +50,8 @@ func TestLoadConfig(t *testing.T) {
   "fs_prefetch": {
     "enable": true,
     "threads_count": 10,
-    "merging_size": 131072
+    "merging_size": 131072,
+    "stream_prefetch": true
   }
 }`)
 	var cfg FuseDaemonConfig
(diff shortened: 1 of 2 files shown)


Fricounet (approved):
LGTM

PR #767, opened 2026-07-03
[Feat] Support user-namespace UID/GID remap-ids for fusedev

### LINKED ISSUE #750: [Feature] Support idmapping and advertise `remap-ids` to better support user namespaces
### Feature Description

It would be good if nydus-snapshotter advertises the `remap-ids` capability so that it can work properly with user namespaces in kubernetes.

This will require changes both here and in nydusd.

### Problem and Use Case

When using user namespaces (ie, `hostUsers: false` in pod spec) in kubernetes, it has to use idmapping in order to get the file uids and gids to look correct inside the container. k8s chooses effectively random (or at least arbitrary) host UIDs to map the container's uids to. These arbitrary UIDs will usually not match whatever is in the container image, so a reverse layer of idmapping is needed to map back to the original uids. This requires filesystem and snapshotter support.

Right now, if you use nydus snapshotter in a k8s container with user namespaces (ie, `hostUsers: false` in the pod spec), k8s will request containerd to perform idmapping. Containerd, upon seeing that the snapshotter is not advertising `remap-ids` support, will do a slow workaround (which I haven't fully investigated) where it roughly creates a copy of the relevant image and chowns it to the correct permissions. This is very slow and defeats the purpose of using nydus.

### Related issues

_No response_

Zephyrcf:
## Overview
When a Kubernetes Pod has hostUsers: false, containerd creates a user namespace with a unique UID/GID range, sets containerd.io/snapshot/uidmapping and containerd.io/snapshot/gidmapping labels on the
  Active snapshot, and — if the snapshotter declares capabilities = ["remap-ids"] — skips its own recursive chown, expecting the snapshotter to handle ownership remapping.

  This PR makes the nydus snapshotter honor those labels, giving each pod its own nydusd FUSE mount with a distinct id_mapping tuple so:
  - container stat / shows 0:0
  - host stat on the FUSE mount shows the mapped external UID
  - overlay upperdir/workdir are chown'd to the external UID, allowing write
  - two pods sharing the same image get isolated lower-layer mounts and cleanup independently

## Related Issues
Close https://github.com/containerd/nydus-snapshotter/issues/750

## Change Details

## Test Results

## Change Type
- [ ] Bug Fix
- [x] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist
- [ ] I have run a code style check and addressed any warnings/errors.
- [ ] I have added appropriate comments to my code (if applicable).

### DIFF
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -12,6 +12,7 @@ package filesystem
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path"
@@ -248,14 +249,19 @@ func (fs *Filesystem) TryStopSharedDaemon() {
 
 // WaitUntilReady wait until daemon ready by snapshotID, it will wait until nydus domain socket established
 // and the status of nydusd daemon must be ready
-func (fs *Filesystem) WaitUntilReady(snapshotID string) error {
-	rafs := racache.RafsGlobalCache.Get(snapshotID)
+func (fs *Filesystem) WaitUntilReady(snapshotID string, labels map[string]string) error {
+	rafsKey, err := label.RafsInstanceIDFromLabels(snapshotID, labels)
+	if err != nil {
+		return errors.Wrapf(err, "parse id mapping for snapshot %s", snapshotID)
+	}
+
+	rafs := racache.RafsGlobalCache.Get(rafsKey)
 	if rafs == nil {
 		// If NoneDaemon mode, there's no need to wait for daemon ready
 		if config.GetDaemonMode() == config.DaemonModeNone {
 			return nil
 		}
-		return errors.Wrapf(errdefs.ErrNotFound, "no instance %s", snapshotID)
+		return errors.Wrapf(errdefs.ErrNotFound, "no instance %s", rafsKey)
 	}
 
 	if rafs.GetFsDriver() == config.FsDriverFscache || rafs.GetFsDriver() == config.FsDriverFusedev {
@@ -321,19 +327,43 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
 	mu.Lock()
 	defer mu.Unlock()
 
-	rafs := racache.RafsGlobalCache.Get(snapshotID)
-	if rafs != nil {
-		// Instance already exists, how could this happen? Can containerd handle this case?
-		return nil
+	// Parse id-mapping labels early so each pod with a distinct UID range
+	// gets its own RAFS instance (per-pod rafsKey includes the mapping suffix).
+	// A shared daemon has a single id_mapping config, so different pods
+	// cannot reuse the same FUSE mount.
+	idMapping, err := label.ParseIDMapping(labels)
+	if err != nil {
+		return errors.Wrapf(err, "parse id mapping for snapshot %s", snapshotID)
 	}
+
+	// Per-pod rafs cache key: two pods sharing the same image but with
+	// different user-namespace mappings get separate rafs instances.
+	rafsKey := label.RafsInstanceID(snapshotID, idMapping)
 
 	fsDriver := config.GetFsDriver()
 	if label.IsTarfsDataLayer(labels) {
 		fsDriver = config.FsDriverBlockdev
 	}
+
+	if idMapping != nil {
+		if fsDriver != config.FsDriverFusedev {
+			return errors.Errorf("id mapping (remap-ids) is only supported by the fusedev driver, got %q", fsDriver)
+		}
+	}
 	isSharedFusedev := fsDriver == config.FsDriverFusedev && config.GetDaemonMode() == config.DaemonModeShared
 	useSharedDaemon := fsDriver == config.FsDriverFscache || isSharedFusedev
 
+	rafs := racache.RafsGlobalCache.Get(rafsKey)
+	if rafs != nil && idMapping == nil {
+		// Instance already exists and no remap-ids needed — safe to reuse.
+		return nil
+	}
+	if rafs != nil && idMapping != nil {
+		// Per-pod compound rafsKey already includes the UID mapping in
+		// its suffix, so a cache hit means this exact pod already has
+		// a mount — duplicate Mount() call, safe to reuse.
+		return nil
+	}
 	var imageID string
 	imageID, ok := labels[snpkg.TargetRefLabel]
 	if !ok {
@@ -346,17 +376,19 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
 		}
 	}
 
-	rafs, err = racache.NewRafs(snapshotID, imageID, fsDriver)
+	rafs, err = racache.NewRafs(rafsKey, imageID, fsDriver)
 	if err != nil {
 		return errors.Wrapf(err, "create rafs instance %s", snapshotID)
+	}
+	if idMapping != nil {
+		rafs.SourceSnapshotID = snapshotID
 	}
 
 	defer func() {
 		if err != nil {
-			if umountErr := fs.Umount(ctx, snapshotID); umountErr != nil {
+			if umountErr := fs.umountRafsInstances([]*racache.Rafs{rafs}); umountErr != nil {
 				log.L.WithError(umountErr).Warnf("failed to umount snapshot %s during cleanup", snapshotID)
 			}
-			racache.RafsGlobalCache.Remove(snapshotID)
 		}
 	}()
 
@@ -393,6 +425,12 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
 			if err != nil {
 				return err
 			}
+			if idMapping != nil {
+				mp = fmt.Sprintf("%s-%d", mp, idMapping.External)
+				if err := os.MkdirAll(mp, 0755); err != nil {
+					return errors.Wrapf(err, "create idmap mountpoint dir %s", mp)
+				}
+			}
 			d, err = fs.createDaemon(fsManager, config.DaemonModeDedicated, mp, 0)
 			// if daemon already exists for snapshotID, just return
 			if err != nil && !errdefs.IsAlreadyExists(err) {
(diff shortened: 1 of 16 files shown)


pkg/label/label_test.go
```diff
@@ -0,0 +1,221 @@
+/*
+ * Copyright (c) 2025. Nydus Developers. All rights reserved.
```
Fricounet:
```suggestion
 * Copyright (c) 2026. Nydus Developers. All rights reserved.
```

snapshot/snapshot.go
```diff
@@ -656,6 +656,15 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
...
 	}
 
+	var removeRafsSnapshotID string
+	if info.Kind == snapshots.KindActive || info.Kind == snapshots.KindView {
+		if idMapping, parseErr := label.ParseIDMapping(info.Labels); parseErr == nil && idMapping != nil {
+			if metaID, _, metaErr := o.findMetaLayer(ctx, key); metaErr == nil {
+				removeRafsSnapshotID = metaID
+			}
+		}
+	}
```
Fricounet, line 666: can you clarify why we need this? I'm not sure I understand why we need to identify the meta layer when attempting to remove the active snapshot

snapshot/snapshot.go
```diff
@@ -1144,6 +1180,19 @@ func (o *snapshotter) mountRemote(ctx context.Context, labels map[string]string,
 	overlayOptions = append(overlayOptions, lowerDirOption)
 	log.G(ctx).Infof("remote mount options %v", overlayOptions)
 
+	// Chown overlay upperdir/workdir to the mapped host UID (from
+	// remap-ids labels) so the container can write to its rootfs.
+	if s.Kind == snapshots.KindActive {
+		if idMapping, _ := label.ParseIDMapping(labels); idMapping != nil {
+			externalID := int(idMapping.External)
+			for _, d := range []string{o.upperPath(s.ID), o.workPath(s.ID)} {
+				if err := os.Lchown(d, externalID, externalID); err != nil {
```
Fricounet, line 1189: so we are running chown on the whole image filesystem right? I wonder, did you benchmark how costly it's going to be on large images?

pkg/filesystem/fs.go
```diff
@@ -346,17 +376,19 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
 	}
 
-	rafs, err = racache.NewRafs(snapshotID, imageID, fsDriver)
+	rafs, err = racache.NewRafs(rafsKey, imageID, fsDriver)
 	if err != nil {
 		return errors.Wrapf(err, "create rafs instance %s", snapshotID)
 	}
+	if idMapping != nil {
+		rafs.SourceSnapshotID = snapshotID
+	}
```
Fricounet, line 385: would it hurt to configure rafs.SourceSnapshotID regardless of id mapping?

pkg/filesystem/fs.go
```diff
@@ -547,63 +590,139 @@ func (fs *Filesystem) copyFile(src, dst string) error {
...
 		log.L.Debugf("no RAFS filesystem instance associated with snapshot %s", snapshotID)
 		return nil
 	}
 
-	fsDriver := rafs.GetFsDriver()
-	if fsDriver == config.FsDriverNodev {
+	return fs.umountRafsInstances(rafsList)
+}
+
+func (fs *Filesystem) UmountByLabels(_ context.Context, snapshotID string, labels map[string]string) error {
```
Fricounet, line 602: drop the context if unused?

pkg/filesystem/fs.go
```diff
@@ -547,63 +590,139 @@ func (fs *Filesystem) copyFile(src, dst string) error {
...
 		return nil
 	}
-	fsManager, err := fs.getManager(fsDriver)
-	if err != nil {
-		return errors.Wrapf(err, "get manager for filesystem instance %s", rafs.DaemonID)
+
+	return fs.umountRafsInstances([]*racache.Rafs{rafs})
+}
+
+func (fs *Filesystem) collectRafsInstances(snapshotID string) []*racache.Rafs {
```
Fricounet, line 621: add a little doc comment on this function to explain what it collects exactly?

pkg/filesystem/fs.go
```diff
@@ -547,63 +590,139 @@ func (fs *Filesystem) copyFile(src, dst string) error {
...
+	var rafsList []*racache.Rafs
+	if r := racache.RafsGlobalCache.Get(snapshotID); r != nil {
+		rafsList = append(rafsList, r)
+	}
+	for _, r := range racache.RafsGlobalCache.List() {
+		if r.SnapshotID != snapshotID && r.GetSourceSnapshotID() == snapshotID {
+			rafsList = append(rafsList, r)
+		}
```
Fricounet, line 631: do you need the first `RafsGlobalCache.Get` call as you are going to have it in the `RafsGlobalCache.List()` as well?

you could transform 
```
if r.SnapshotID != snapshotID && r.GetSourceSnapshotID() == snapshotID {
```
into 
```
if r.SnapshotID != snapshotID || r.GetSourceSnapshotID() == snapshotID {
```

pkg/filesystem/fs.go
```diff
@@ -321,19 +327,43 @@ func (fs *Filesystem) Mount(ctx context.Context, snapshotID string, labels map[s
...
 	defer mu.Unlock()
 
-	rafs := racache.RafsGlobalCache.Get(snapshotID)
-	if rafs != nil {
-		// Instance already exists, how could this happen? Can containerd handle this case?
-		return nil
+	// Parse id-mapping labels early so we can decide daemon strategy.
+	// Each pod with hostUsers: false gets a unique UID range; shared daemon
+	// mode would reuse a single FUSE mount with a single id_mapping that
+	// only matches the first pod.
```
Fricounet: I probably missed it, but I don't understand where you use this to decide the daemon strategy?
And regarding shared daemon, can't you just chown the subdirectory that corresponds to the give snapshot id? `/<snapshotID>` so that multiple ID mappings can cohexist in the shared daemon?

Fricounet (commented):
Mostly LGTM!
A few comments/questions

snapshot/snapshot.go
Zephyrcf: In idmap scenarios, instance IDs have a suffix: snapshotID-userns-ExternalID.
When deleting an active/view snapshot:
- Non-idmap: RAFS instances (lower layer mount points) are shared and managed for unloading by other mechanisms; they don't need to be handled here.
- idmap: RAFS instances are independent (this should be the case in idmap scenarios). When deleting a snapshot, it's necessary to find the corresponding RAFS instance ID in the meta-layer and unload it; otherwise, FUSE mounts will remain.

snapshot/snapshot.go
Zephyrcf: Here, only two directories are chown, and not recursively: upperdir/workdir (corresponding to the fs and work folders). This is only to ensure that the container has permission to write files. The files will eventually be written to the fs folder. The image-lower layer is mapped by the fuse server at the bottom of nydusd, so there is no need to worry about the cost.

pkg/filesystem/fs.go
Zephyrcf: This will not affect SourceSnapshotID; for example, image.boot and blob.meta files will remain in SourceSnapshotdir for use by idmap snapshots.

pkg/filesystem/fs.go
Zephyrcf: To maintain consistency with Umount, timeout/deletion control may be implemented for umountRafsInstances in the future.

pkg/filesystem/fs.go
Zephyrcf: Two-step approach is intentional: Get finds the shared instance, the loop finds idmap derivatives. I can add some comments.

pkg/filesystem/fs.go
Zephyrcf: The comments are outdated. The first implementation did not yet support shared mode idmaps; it switched modes based on idMapping. Now, shared mode per-pod instances are supported, and idMapping no longer determines the mode; it's only used to build the per-pod rafsKey. I will update the comments.

Regarding `chown` subdirectories: No. The shared daemon has only one set of id_mapping, which applies to all files within the daemon service, not per-subdirectory. The `chown <root>/mnt/<snapshotID>` subdirectory entry itself does not change the daemon's internal UID/GID translation of file content, so multiple pods with different mappings will still conflict; only the first container will successfully set it. This is why I committed https://github.com/cloud-hypervisor/fuse-backend-rs/pull/227.

pkg/label/label_test.go
Zephyrcf: done.

pkg/filesystem/fs.go
Zephyrcf: done.

pkg/filesystem/fs.go
Fricounet: sorry I'm not sure I understand your answer.
Why not configure `rafs.SourceSnapshotID = snapshotID` even if idMapping is nil?
so that the field is always set (and in the case where there's no idMapping, `rafs.SourceSnapshotID` is the same as `rafs.SnapshotID`

Zephyrcf: When `SourceSnapshotID` is empty, the getter `GetSourceSnapshotID()` falls back to `SnapshotID` (see `rafs.go:155-161`), so all users will work correctly in both cases—always setting it is completely redundant.

More importantly, an empty `SourceSnapshotID` has semantic significance: it indicates that "this RAFS instance is not a derived instance of another snapshot." For ID mappings, each container's lower layer is a derived instance, requiring a lookup of the bootstrap/metadata file from the source snapshot. For non-ID mappings, this relationship does not exist, so leaving this field empty clearly distinguishes between the two cases.

Therefore, always setting it provides no functional benefit, while maintaining its conditional setting preserves a useful semantic signal.

Fricounet: thanks for the clarification, sounds good :+1:

Fricounet (approved):
LGTM!

PR #768, opened 2026-07-04
fix(daemon): only DecRef on Umount when the rafs instance was cached

### LINKED ISSUE #763: [Bug] Duplicate Umount can decrement shared daemon refcount incorrectly
### Problem Description

`Filesystem.Umount()` may be called more than once for the same snapshot during cleanup, retry, or recovery.

Currently `Daemon.RemoveRafsInstance(snapshotID)` always decrements the daemon refcount, even if the RAFS instance is no longer present in the daemon RAFS cache.

This can make the daemon refcount inaccurate. In shared daemon mode, a wrong refcount may cause the shared nydusd to be destroyed unexpectedly.

### Expected Behavior

RemoveRafsInstance() should only decrement the daemon refcount when it actually removed an existing RAFS instance.

After a successful unmount, the global RAFS cache should also be cleared to avoid stale entries and duplicate cleanup.

Suggestion:
* Make RafsCache.Remove(snapshotID) return whether an entry was removed.
* Only call DecRef() when the RAFS instance existed.
* Remove the snapshot from RafsGlobalCache after successful Filesystem.Umount().

### Actual Behavior

`RemoveRafsInstance()` removes the RAFS instance and decrements refcount unconditionally:
```go
func (d *Daemon) RemoveRafsInstance(snapshotID string) {
    if r := d.RafsCache.Get(snapshotID); r != nil {
        collector.NewDaemonImageCollector(d.ID(), r.ImageID).Delete()
    }
    d.RafsCache.Remove(snapshotID)
    d.DecRef()
}
```
But RafsCache.Remove() does not tell whether anything was actually removed.

### How to reproduce

.

### Environment Details

.

### Additional Information

_No response_

arpitjain099:
## Overview

I noticed #763 while reading the open bugs. `RemoveRafsInstance` unconditionally calls `DecRef()`, even when the snapshot was not in the RafsCache. Since `AddRafsInstance` only bumps the refcount when it actually adds an instance, a duplicate Umount (a retry or recovery for an already-removed snapshot) decrements a second time, and in shared-daemon mode that can push the count too low and tear down a nydusd that other snapshots still need.

This moves the `Remove` and `DecRef()` calls into the existing `if r != nil` guard so we only drop the reference when there was one to drop. I also added a daemon test that removes the same snapshot twice and checks the refcount stays at 0 instead of going to -1.

## Related Issues

Fix #763

## Test Results

`go test ./pkg/daemon/...` and `go vet ./pkg/daemon/...` pass on linux/amd64, and the new test fails without the fix.

## Change Type

- [x] Bug Fix

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -159,11 +159,16 @@ func (d *Daemon) UpdateRafsInstance(r *rafs.Rafs) {
 }
 
 func (d *Daemon) RemoveRafsInstance(snapshotID string) {
+	// Only drop the reference when the instance was actually cached.
+	// AddRafsInstance only IncRef()s when it adds, so a duplicate Umount
+	// (retry / recovery) for an already-removed snapshot must not DecRef()
+	// again, otherwise a shared daemon's refcount goes too low and the
+	// nydusd can be torn down while other snapshots still use it.
 	if r := d.RafsCache.Get(snapshotID); r != nil {
 		collector.NewDaemonImageCollector(d.ID(), r.ImageID).Delete()
+		d.RafsCache.Remove(snapshotID)
+		d.DecRef()
 	}
-	d.RafsCache.Remove(snapshotID)
-	d.DecRef()
 }
 
 // Get and cache daemon current working state by querying nydusd:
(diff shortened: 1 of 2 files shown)


Fricounet (approved):
LGTM! Thanks for the fix!

Issue #771, opened 2026-07-14
[Bug] nydusd process storm / orphan accumulation under registry pressure


qinfustu:
### Problem Description

Under registry pressure (slow/overloaded registry), a single node can accumulate
**hundreds to thousands of `nydusd` processes**. They mount under the same
directory, each independently pulls from the registry (auth + blob + prefetch),
multiplying registry QPS and driving the registry further down — a positive
feedback loop. Stale mountpoints cannot be unmounted and API sockets cannot be
removed, so the processes never get reclaimed.

### Expected Behavior

- The number of `nydusd` processes stays bounded (roughly one per distinct
  nydus image, per `daemon_mode`), even when the registry is slow.
- Failed or stale `nydusd` processes are reclaimed, and their mountpoints and
  API sockets are cleaned up.

### Actual Behavior

- Under registry pressure the node accumulates hundreds/thousands of live
  `nydusd` processes, far exceeding the number of distinct nydus images.
- They mount under the same directory and each pulls independently, spiking
  registry QPS/connections.
- Residual mountpoints can't be unmounted (`EBUSY`/`ENOTCONN`) and API sockets
  are left behind under the snapshotter root.
- Restarting the snapshotter does not reclaim the orphaned processes.

### How to reproduce

1. Deploy nydus-snapshotter (`daemon_mode=multiple`, `recover_policy=restart`,
   defaults) with containerd 1.7.x/2.1.x on a node.
2. Prepare several nydus (RAFS) images pointing at a registry.
3. Make the registry slow/unstable — e.g. throttle bandwidth, inject latency
   (`tc`/proxy), or rate-limit it — so nydusd startup pulls exceed the ~10s/~2s
   startup timeouts.
4. Concurrently start many pods using those nydus images (or repeatedly
   create/delete them) to drive repeated `Prepare` calls.
5. Observe on the node:
   - `pgrep -a nydusd | wc -l` keeps climbing far beyond the number of distinct
     images (hundreds/thousands).
   - Registry QPS/connections spike as every process pulls independently.
   - `mount | grep nydus` shows residual mountpoints that can't be unmounted
     (`EBUSY`/`ENOTCONN`); leftover API sockets remain under the snapshotter root.
   - Killing/restarting the snapshotter does not reclaim the orphaned nydusd.

### Environment Details

- Nydus-snapshotter version: v0.15.0 
- Nydus version: v2.3.4
- Container runtime: 1.7.x / 2.1.x
[...]

PR #772, opened 2026-07-14
fix: make unmount handling robust for busy/disconnected mounts

qinfustu:
## Overview
Make `mount.Umount` (and `WaitUntilUnmounted`) robust against **busy** and
**disconnected** mountpoints, so stale `nydusd` FUSE mounts are always torn down
instead of being left behind.

## Related Issues
Related #771

## Change Details
Previously `Umount` had two problems that leave residual mountpoints under
registry pressure / after a nydusd crash:
1. **Bails out on `ENOTCONN`.** When the backing nydusd is dead, the FUSE mount
   becomes disconnected and `IsMountpoint` (via `os.Stat`) fails with `ENOTCONN`.
   The old code returned that error and never even attempted to unmount — i.e.
   the mountpoint that most needs cleaning up was skipped.
2. **Bare unmount gives up on `EBUSY`.** `syscall.Unmount(target, 0)` had no
   fallback, so a busy mountpoint (old process still holding it) was left mounted.
Changes:
- `Umount` now treats an `ENOTCONN`/`ESTALE` mountpoint as "needs unmount" and
  proceeds instead of returning early; a genuinely-not-mounted path now returns
  `nil` (previously returned a misleading `"not mounted"` error).
- New `unmountWithFallback`: plain unmount → `MNT_FORCE` (abort in-flight FUSE
  requests) → `MNT_DETACH` (lazy detach for busy mounts). `EINVAL` is treated as
  already unmounted.
- `WaitUntilUnmounted` no longer reports a disconnected mountpoint as an error
  (it would otherwise spin for ~1s and log a misleading failure during teardown).
- `MNT_FORCE`/`MNT_DETACH` are platform-specific, so they live in
  `mount_linux.go` / `mount_darwin.go` (darwin has no lazy detach and degrades to
  force; that build exists only for local dev — nydusd mounts are Linux-only at
  runtime).
- Added small test seams (`syscallUnmount`, `isMountpoint`) to enable unit tests.
The only production caller of `mount.Mounter.Umount` is `daemon.ClearVestige`, so
the blast radius is small.

### Behaviour changes on exported API

- `Mounter.Umount` returns `nil` for a path that is not a mountpoint, where it
  previously returned `errors.New("not mounted")`. Teardown is now idempotent.
- `Mounter.Umount` returns an error wrapping the new `ErrUnmountedUnclean` when the
  path was freed by `MNT_FORCE`/`MNT_DETACH` rather than a plain unmount. The path is
  free to be mounted over, but the old mount may still be live and referenced, so
  callers must not assume the backing resources are safe to reclaim.
- `WaitUntilUnmounted` returns an error wrapping both `ErrMountpointDisconnected` and
  the original `ENOTCONN`/`ESTALE` for a disconnected mountpoint instead of `nil`; it
  is still mounted, and the wait is aborted immediately because the backing server
  never comes back.

## Test Results
New unit tests in `pkg/utils/mount/mount_test.go`:
- `TestIsDisconnected` — `ENOTCONN`/`ESTALE` detection, including through
  `pkg/errors` wrapping.
- `TestUnmountWithFallback` — plain success, `EINVAL` short-circuit,
  `EBUSY → force`, `force → lazy detach`, and all-fail returns error.
- `TestUmount` — disconnected mount still unmounts; non-mountpoint returns nil
  without unmounting; other errors propagate.
- `TestWaitUntilUnmountedIgnoresDisconnectedMountpoint` — disconnected mount
  returns immediately without retrying.

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/pkg/utils/mount/mount.go
+++ b/pkg/utils/mount/mount.go
@@ -13,6 +13,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containerd/log"
 	"github.com/pkg/errors"
 
 	"github.com/containerd/nydus-snapshotter/pkg/errdefs"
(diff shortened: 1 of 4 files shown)


Zephyrcf (approved):
LGTM

imeoer:
The force/lazy fallback needs a call from the maintainers, because `unmountWithFallback` returns nil after `MNT_DETACH` succeeds. Lazy unmount only detaches from the namespace — the mount stays live until the last reference goes away, so `ClearVestige` moves on and cleans up backing resources while a container may still be using it. `MNT_FORCE` similarly aborts in-flight FUSE requests and the container sees I/O errors. If we want this behaviour, it would be better to signal "unmounted, but not cleanly" back to the caller instead of a plain nil.

`WaitUntilUnmounted` returning nil on ENOTCONN looks wrong to me. A disconnected mountpoint is still mounted, the server is just gone. We're telling the caller the unmount finished when nothing was unmounted.

Two smaller ones:
- `Mounter.Umount` now returns nil instead of `errors.New("not mounted")` for a non-mountpoint. That's a contract change on an exported method, please mention it in the description.
- `mount_darwin.go` maps `umountDetach` to `MNT_FORCE`, so the fallback fires `MNT_FORCE` twice with different intent. We don't run on darwin, so I'd just drop the file rather than carry the extra surface.

qinfustu:
Thanks for the review @imeoer, all four points addressed:

- `unmountWithFallback` no longer returns `nil` after `MNT_FORCE`/`MNT_DETACH`; it now
  returns an error wrapping a new `ErrUnmountedUnclean` sentinel. `ClearVestige` logs
  that case distinctly so it's clear the old mount may still be referenced.
- `WaitUntilUnmounted` returns an error (wrapping both `ErrMountpointDisconnected` and
  the original `ENOTCONN`/`ESTALE`) instead of `nil` for a disconnected mountpoint, and
  stops retrying immediately since the server won't come back.
- The `Umount` contract change (`nil` for a non-mountpoint instead of `"not mounted"`)
  is now called out in the description.
- Dropped `mount_darwin.go`; the flags live only in `mount_linux.go`.

Unit tests updated accordingly. PTAL, thanks!

PR #773, opened 2026-07-14
fix: startdaemon reclaim on startup failure

qinfustu:
## Overview
Terminate nydusd processes that fail startup verification instead of leaving
them running as orphans that keep pulling from the registry.

## Related Issues
Related #771

## Change Details
When `StartDaemon`'s background goroutine detects that a newly spawned nydusd
failed to come up (API socket not ready within ~10s, event subscription failed,
or daemon not RUNNING within ~2s), it previously just logged and returned,
leaving the process alive. Under registry pressure these timeouts are easily
hit, and upper-layer retries (`Prepare`, `doDaemonRestart`) spawn yet another
daemon — each round leaks one live, still-pulling orphan.
**`pkg/manager/daemon_adaptor.go`**
- On any startup-verification failure, call the new `terminateFailedDaemon`
  helper which does `SIGTERM` → 5s timeout → `SIGKILL` → `Wait` (reap).
  It operates on the captured `*os.Process` handle (not pid) so it is safe
  against pid reuse and concurrent teardown.
- Before killing, unsubscribe the daemon from the liveness monitor so our own
  kill is not misinterpreted as a death event that triggers another restart.
- Two guards prevent mis-kills:
  - **Failover daemons** (`d.Supervisor != nil`): a failover-managed daemon
    legitimately stays in INIT/READY while the takeover flow runs
    (`TakeOver` → `Start`). Killing it would wreck the takeover, so the helper
    skips it and leaves cleanup to the failover flow.
  - **Already-retained shared daemons**: if a shared daemon has already been
    retained via `TryRetainSharedDaemon` (e.g. spurious timeout race), killing
    it would drop the active shared daemon. A new `IsSharedDaemonRetained`
    callback lets the helper detect and skip this case.
**`pkg/manager/manager.go`**
- Added `IsSharedDaemonRetained` callback field to the `Manager` struct.
**`pkg/filesystem/fs.go`**
- `initSharedDaemon` now waits for `DaemonStateRunning` **before** calling
  `TryRetainSharedDaemon`, so a failed shared daemon is never retained with a
  dead process — and a truly-failed one is cleaned up by the helper above.
- Wired `IsSharedDaemonRetained` callbacks for both fscache and fusedev managers.

## Test Results
New unit tests in `pkg/manager/daemon_adaptor_test.go`:
- `TestTerminateFailedDaemon/nil_process_is_a_no-op`
- `TestTerminateFailedDaemon/SIGTERM_stops_a_well-behaved_process` — asserts
  the process is reaped and finishes well before the SIGKILL escalation timeout.
- `TestTerminateFailedDaemon/escalates_to_SIGKILL_when_SIGTERM_is_ignored` —
  spawns a process that traps SIGTERM, asserts it is still killed via SIGKILL.

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -109,6 +109,9 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
 	// TODO: We still need to consider shared daemon the time sequence of initializing daemon,
 	// start daemon commit its state to DB and retrieving its state.
 	if fscacheManager, ok := fs.enabledManagers[config.FsDriverFscache]; ok {
+		fscacheManager.IsSharedDaemonRetained = func(d *daemon.Daemon) bool {
+			return fs.fscacheSharedDaemon != nil && fs.fscacheSharedDaemon.ID() == d.ID()
+		}
 		if !hasFscacheSharedDaemon && fs.fscacheSharedDaemon == nil {
 			log.L.Infof("initializing shared nydus daemon for fscache")
 			if err := fs.initSharedDaemon(fscacheManager); err != nil {
@@ -119,6 +122,9 @@ func NewFileSystem(ctx context.Context, opt ...NewFSOpt) (*Filesystem, error) {
 		return nil, errors.Errorf("shared fscache daemon is present, but manager is missing")
 	}
 	if fusedevManager, ok := fs.enabledManagers[config.FsDriverFusedev]; ok {
+		fusedevManager.IsSharedDaemonRetained = func(d *daemon.Daemon) bool {
+			return fs.fusedevSharedDaemon != nil && fs.fusedevSharedDaemon.ID() == d.ID()
+		}
 		if config.IsFusedevSharedModeEnabled() && !hasFusedevSharedDaemon && fs.fusedevSharedDaemon == nil {
 			log.L.Infof("initializing shared nydus daemon for fusedev")
 			if err := fs.initSharedDaemon(fusedevManager); err != nil {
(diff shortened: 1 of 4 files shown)


Zephyrcf (approved):
LGTM

imeoer:
Three things:

- `defaultDaemonTerminationTimeout` is added here and in #774, same package, so the two can't both merge — I tried it locally and the build breaks on the duplicate declaration. Both PRs also grow their own SIGTERM→SIGKILL helper (`terminateFailedDaemon` vs `terminateDaemonProcess`) with opposite policies for failover daemons. Could this and #774 be folded into a single change?
- Under `recover_policy=failover`, `SupervisorSet` is always created, so every daemon has `d.Supervisor != nil` and `terminateFailedDaemon` returns immediately. #771 stays unfixed for those users. That limitation should at least be spelled out in the PR description.
- The new `WaitUntilState(Running)` in `initSharedDaemon` only retries 20 × 100ms, so 2s total, and the error propagates out of `NewFileSystem`. On a cold or busy node nydusd won't always reach RUNNING that fast and the snapshotter just refuses to start. Either give this wait its own longer budget or downgrade it to a warning.

Smaller ones: `fscacheManager.IsSharedDaemonRetained = ...` is a plain field write on an already-live `Manager` while `StartDaemon` goroutines read it — please pass it through `manager.Opt` instead. The failover comment is also duplicated verbatim in `StartDaemon` and `terminateFailedDaemon`. And the old `// FIXME: Should clean the daemon record in DB if the nydusd fails starting` was dropped, but the record still isn't cleaned up.

PR #774, opened 2026-07-15
fix: serialize daemon restart and termination

qinfustu:
## Overview
When a nydusd daemon death event is received, the manager now terminates the
old daemon process before starting a new one, and serializes recovery so that
only one recovery routine runs per daemon at a time. This prevents zombie/
orphan nydusd accumulation and avoids duplicate concurrent restarts.
## Related Issues
Related #771

## Change Details
Previously, on a daemon death event `doDaemonRestart` / `doDaemonFailover` only
called `d.Wait()` and then started a new nydusd. When the death event was a
false positive (the API socket closed while the process was hung), the old
process kept running while a new one was spawned, and duplicate death events
could trigger multiple concurrent recoveries. Over time this leaked orphan
nydusd processes that kept hitting the registry.

This PR makes recovery robust:

- **Terminate the old process before restart/failover** (`terminateDaemonProcess`):
  - Grabs the `os.Process` handle first (uses a pidfd on Linux) to pin the
    handle to the specific process before any identity check, eliminating a
    PID-reuse TOCTOU.
  - Verifies process identity via `/proc/<pid>/cmdline` (matching `--apisock`)
    and classifies it as alive / zombie / gone / reused.
  - For a live process: sends `SIGTERM`, waits with a timeout, then escalates
    to `SIGKILL`; for a zombie: reaps it via `Wait()` with a timeout; for
    gone/reused: does nothing.
  - Returns an error on failure so recovery aborts instead of leaving an
    unmonitored live process behind. Unsubscribe now happens only after
    successful termination.
- **Serialize recovery per daemon** (`recoveryInFlight sync.Map` on `Manager`):
  - `beginDaemonRecovery` uses `LoadOrStore` to dedup concurrent death events;
    the flag is cleared when the recovery goroutine finishes.
- **Guard against stale death events**: the liveness monitor now carries the
  subscribed process PID in `deathEvent`, and `handleDaemonDeathEvent` ignores
  events whose PID no longer matches the daemon's current PID (generation
  check), preventing a delayed event for an old process from killing its
  replacement.
- Removed the `// TODO: ratelimit` and reads `d.Pid()` under the daemon lock
  (`daemonProcessID`) to avoid a data race.

## Test Results
Added unit tests in `pkg/manager/daemon_event_test.go` and updated
`monitor_test.go`:

- `TestBeginDaemonRecoverySerializesConcurrentCalls` – dedup guard.
- `TestInspectDaemonProcess` – alive / zombie / gone / reused classification.
- `TestTerminateDaemonProcessEscalatesToKill` – SIGTERM → SIGKILL escalation.
- `TestTerminateDaemonProcessReapsZombie` – zombie reap path.
- `TestTerminateDaemonProcessReturnsSignalErrorWithoutWaiting` – abort on
  signal failure.
- `TestHandleDaemonDeathEvent_*` – removed-daemon skip, dedup skip, flag clear,
  stale-PID skip.
- `TestLivenessMonitor` – verifies the PID is propagated in the death event.

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist

### DIFF
--- a/pkg/manager/daemon_event.go
+++ b/pkg/manager/daemon_event.go
@@ -9,10 +9,12 @@ package manager
 
 import (
 	"fmt"
+	"os"
 	"path"
 	"regexp"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/containerd/log"
@@ -23,8 +25,16 @@ import (
 	"github.com/pkg/errors"
 )
 
+var (
+	readProcessCmdline      = os.ReadFile
+	processExitPollInterval = 100 * time.Millisecond
+)
+
+var defaultDaemonTerminationTimeout = 5 * time.Second
+
 func (m *Manager) SubscribeDaemonEvent(d *daemon.Daemon) error {
-	if err := m.monitor.Subscribe(d.ID(), d.GetAPISock(), m.LivenessNotifier); err != nil {
+	pid := daemonProcessID(d)
+	if err := m.monitor.Subscribe(d.ID(), pid, d.GetAPISock(), m.LivenessNotifier); err != nil {
 		log.L.Errorf("Nydusd %s probably not started", d.ID())
 		return errors.Wrapf(err, "subscribe daemon %s", d.ID())
 	}
@@ -41,13 +51,25 @@ func (m *Manager) UnsubscribeDaemonEvent(d *daemon.Daemon) error {
 }
 
 func (m *Manager) handleDaemonDeathEvent() {
-	// TODO: ratelimit for daemon recovery operations?
 	for ev := range m.LivenessNotifier {
 		log.L.Warnf("Daemon %s died! socket path %s", ev.daemonID, ev.path)
 
 		d := m.GetByDaemonID(ev.daemonID)
 		if d == nil {
 			log.L.Warnf("Daemon %s was not found, may have been removed", ev.daemonID)
+			continue
+		}
+		currentPID := daemonProcessID(d)
+		if ev.processID != currentPID {
+			log.L.Warnf(
+				"Ignore stale death event for daemon %s: event pid %d, current pid %d",
+				ev.daemonID, ev.processID, currentPID,
+			)
+			continue
+		}
+
+		if m.isRecoveryInFlight(ev.daemonID) {
+			log.L.Warnf("Recovery already in progress for daemon %s, skipping duplicate death event", ev.daemonID)
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


pkg/manager/daemon_event.go
```diff
@@ -60,22 +89,44 @@ func (m *Manager) handleDaemonDeathEvent() {
...
-			// RecoverPolicyNone or RecoverPolicyInvalid - do nothing
+			m.recoveryInFlight.Delete(ev.daemonID)
 		}
 	}
 }
 
+func (m *Manager) beginDaemonRecovery(daemonID string) bool {
+	_, alreadyRecovering := m.recoveryInFlight.LoadOrStore(daemonID, struct{}{})
+	return !alreadyRecovering
+}
```
Zephyrcf:
```suggestion
func (m *Manager) isRecoveryInFlight(daemonID string) bool {
    _, loaded := m.recoveryInFlight.LoadOrStore(daemonID, struct{}{})
    return loaded
}
```
This makes it more readable.

pkg/manager/daemon_event.go
```diff
@@ -60,22 +89,44 @@ func (m *Manager) handleDaemonDeathEvent() {
...
+	defer d.Unlock()
+	return d.Pid()
+}
+
+func (m *Manager) runDaemonRecovery(d *daemon.Daemon, recover func()) {
+	go func() {
+		defer m.recoveryInFlight.Delete(d.ID())
+		recover()
+	}()
+}
```
Zephyrcf:
```suggestion
```
There's no need for excessive wrapping. Simply place `defer m.recoveryInFlight.Delete(d.ID())` directly into `doDaemonRestart` and `doDaemonFailover`, and restore the original call path. This is more intuitive.

pkg/manager/daemon_event.go
```diff
@@ -23,8 +25,23 @@ import (
...
 
+var (
+	readProcessCmdline = os.ReadFile
+	findProcess        = os.FindProcess
+	signalProcess      = func(p *os.Process, sig syscall.Signal) error {
+		return p.Signal(sig)
+	}
+	waitDaemonProcess = func(d *daemon.Daemon) error {
+		return d.Wait()
+	}
```
Zephyrcf: These are each a single line of code, but wrapping them in production code adds an extra jump. If the goal is simply to mock by extracting `os.FindProcess`, `p.Signal`, os.ReadFile, and `d.Wait` into variables line by line, I believe this is unnecessary and significantly impacts code readability.

pkg/manager/daemon_event.go
```diff
@@ -139,6 +190,145 @@ func (m *Manager) doDaemonRestart(d *daemon.Daemon) {
...
+		done <- waitDaemonProcess(d)
+	}()
+
+	select {
+	case err := <-done:
+		return err
+	case <-time.After(timeout):
+		return errors.Errorf("wait timed out after %s", timeout)
+	}
+}
```
Zephyrcf:
```suggestion
```
A zombie process is a completely dead child process. `d.Wait()` only reclaims the PCB for it; the operation is instantaneous. Adding a 5-second timeout plus a new goroutine to an instantaneous operation is meaningless.

Zephyrcf (changes requested):
Thanks for this pr. Here are a few questions:

qinfustu:
> Thanks for this pr. Here are a few questions:

@Zephyrcf Thanks for the review. I’ve addressed all the comments and updated the code accordingly. Please take another look when convenient.

Zephyrcf:
@qinfustu plz squash the commits.

Zephyrcf (approved):
LGTM, thanks!

qinfustu:
> LGTM, thanks!

@Zephyrcf Thanks! Security scan is still failing with GO-2026-5970. How should we handle it?

Zephyrcf:
> > LGTM, thanks!
> 
> @Zephyrcf Thanks! Security scan is still failing with GO-2026-5970. How should we handle it?

I submitted https://github.com/containerd/nydus-snapshotter/pull/788 to solve this problem.

imeoer:
This collides with #773: both add `defaultDaemonTerminationTimeout` to package `manager`, and both add their own process-termination helper. I tried merging them locally and the second one fails to build. Worth merging the two PRs into one.

On `inspectDaemonProcess`: identifying the process by grepping `--apisock <path>` out of `/proc/<pid>/cmdline` only works because `BuildCommand` currently emits the flag and its value as two separate argv entries. If nydusd ever accepts `--apisock=<path>`, the check silently reports `daemonProcessReused` and we skip termination. #789 is solving the same "is this still our process" question using the start time from `/proc/<pid>/stat`, which doesn't depend on the CLI shape. Can we settle on one approach across both PRs?

Also still open: @Zephyrcf asked for a squash.

PR #776, opened 2026-07-17
fix(snapshot): only protect idmapped RAFS instances from orphan cleanup

iaroslav-reflection:
`getCleanupDirectories` protects live RAFS instance directories from orphan cleanup by building `liveRafsSnapshotDirs` from every entry in `RafsGlobalCache`. This was introduced in #767 to keep per-pod idmapped instances (`<snapshotID>-userns-<uid>`), which are absent from containerd's snapshot `IDMap`, from being reaped while live.

It also protects ordinary (non-idmapped) instances, whose directory name is exactly the containerd snapshot ID. Those are already retained by the existing `IDMap` (`ids`) check while live. Once their metadata is removed by image GC they should become cleanup-eligible, but the instance remains in `RafsGlobalCache` until it is unmounted — and cleanup is what performs the unmount. The instance therefore protects itself indefinitely: it is never unmounted, never leaves the cache, and its mount keeps the cached blobs referenced, so disk is never reclaimed.

This fix restricts the protection set to genuinely idmapped instances (the `-userns-` suffix). Ordinary instances remain governed solely by the `IDMap` check — retained while live, eligible for cleanup once GC'd. Idmapped instances are unmounted on `Remove` via the existing `UmountByLabels`, after which they leave the cache and become eligible on the next pass.

A regression test (`TestGetCleanupDirectoriesProtectsOnlyIDMappedInstances`) covers all three cases: a live ordinary instance stays protected via the `IDMap`, an orphaned ordinary instance becomes cleanup-eligible, and a live idmapped instance stays protected.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -1281,11 +1281,20 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context) ([]string, erro
 		return nil, err
 	}
 
-	// Per-pod RAFS instances (e.g. "48-userns-<uid>") live outside
-	// containerd's IDMap — keep them alive while still referenced.
+	// Per-pod idmapped RAFS instances (e.g. "48-userns-<uid>") live outside
+	// containerd's IDMap, so they must be protected from orphan cleanup while
+	// still referenced. They are the only instances with a SourceSnapshotID
+	// (set when remap-ids labels are present). Ordinary instances share their
+	// ID with containerd's IDMap: while live they are already retained by the
+	// `ids` check below, and once their snapshot metadata is removed they must
+	// stay eligible for cleanup so the mount is torn down and the cached blobs
+	// released. Protecting them here would keep the mount referenced
+	// indefinitely and leak disk.
 	liveRafsSnapshotDirs := make(map[string]struct{})
 	for _, instance := range rafs.RafsGlobalCache.List() {
-		liveRafsSnapshotDirs[filepath.Base(instance.GetSnapshotDir())] = struct{}{}
+		if instance.SourceSnapshotID != "" {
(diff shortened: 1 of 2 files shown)


iaroslav-reflection:
cc @Zephyrcf @Fricounet — this is a small follow-up to the orphan-cleanup path added in #767. Would appreciate your review when you have a moment.

snapshot/snapshot.go
```diff
@@ -1281,10 +1281,19 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context) ([]string, erro
...
+	// and once their snapshot metadata is removed they must stay eligible for
+	// cleanup so the mount is torn down and the cached blobs released.
+	// Protecting them here would keep the mount referenced indefinitely and
+	// leak disk.
 	liveRafsSnapshotDirs := make(map[string]struct{})
 	for _, instance := range rafs.RafsGlobalCache.List() {
+		if !label.IsIDMappedRafsInstanceID(instance.SnapshotID) {
+			continue
+		}
 		liveRafsSnapshotDirs[filepath.Base(instance.GetSnapshotDir())] = struct{}{}
```
Zephyrcf:
```suggestion
		if instance.SourceSnapshotID != "" {
			liveRafsSnapshotDirs[filepath.Base(instance.GetSnapshotDir())] = struct{}{}
		}
```

Zephyrcf (changes requested):

snapshot/snapshot.go
iaroslav-reflection: Thanks, applied!

Zephyrcf (approved):
LGTM

PR #766, opened 2026-06-22
fix(snapshot): clean up fscache snapshots asynchronously

Park-Jiyeonn:
## Overview
Run fscache snapshot directory cleanup after metadata removal has been committed. This keeps slow unmount, tarfs detach, and RemoveAll work out of Remove's critical path.

## Related Issues
Fixes #761 

## Change Type
_Please select the type of change your pull request relates to:_
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [x] Performance Improvement
- [ ] Other (please describe)

## Self-Checklist
_Before submitting a pull request, please ensure you have completed the following:_
- [x] I have run a code style check and addressed any warnings/errors.
- [x] I have added appropriate comments to my code (if applicable).
- [x] I have updated the documentation (if applicable).
- [ ] I have written appropriate unit tests.

PR #777, opened 2026-07-20
test(snapshot): Remove cleans up snapshot directories only after the metadata commit

iaroslav-reflection:
A failed removal leaves the snapshot directories on disk, also when the metadata removal succeeds but its commit fails. A successful removal removes them, and any orphan directory, after the commit.

The commit failure is forced with a zero `RLIMIT_FSIZE`, so bolt's first page write fails with `EFBIG`. The test fails on the tree before #766.

### DIFF
--- a/snapshot/snapshot_test.go
+++ b/snapshot/snapshot_test.go
@@ -464,3 +464,50 @@ func TestMountRemoteUsesPerPodLowerdirForIDMap(t *testing.T) {
 	assert.Contains(t, mounts[0].Options, fmt.Sprintf("lowerdir=%s", mountpoint))
 	assert.NotContains(t, mounts[0].Options, fmt.Sprintf("lowerdir=%s", filepath.Join(snapshotterRoot, "snapshots", "meta", "fs")))
 }
+
+func TestRemoveCleansUpDirectoriesOnlyAfterCommit(t *testing.T) {
+	originalInstances := rafs.RafsGlobalCache.List()
+	rafs.RafsGlobalCache.SetIntances(make(map[string]*rafs.Rafs))
+	t.Cleanup(func() {
+		rafs.RafsGlobalCache.SetIntances(originalInstances)
+	})
+
+	snapshotterRoot := t.TempDir()
+	ms, err := storage.NewMetaStore(filepath.Join(snapshotterRoot, "metadata.db"))
+	require.NoError(t, err)
+	s := &snapshotter{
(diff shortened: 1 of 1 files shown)


iaroslav-reflection:
@imeoer @Zephyrcf PTAL

snapshot/snapshot.go
```diff
@@ -638,80 +638,70 @@ func (o *snapshotter) Remove(ctx context.Context, key string) error {
...
+			// For example: remove snapshot with key sha256:c33c40022c8f333e7f199cd094bd56758bc479ceabf1e490bb75497bf47c2ebf
+			log.L.Infof("[Remove] snapshot with key %s snapshot id %s", key, id)
+		}
 
-	if o.syncRemove {
-		var removals []string
-		removals, err = o.getCleanupDirectories(ctx)
-		if err != nil {
-			return errors.Wrap(err, "get directories for removal")
+		if _, _, err := storage.Remove(ctx, key); err != nil {
```
Zephyrcf:
```suggestion
		if _, _, err = storage.Remove(ctx, key); err != nil {
```

Zephyrcf (approved):
LGTM, thanks!

snapshot/snapshot.go
iaroslav-reflection: Thanks, applied!

Zephyrcf (approved):
LGTM

iaroslav-reflection:
@imeoer approved by @Zephyrcf, good to merge?

iaroslav-reflection:
#766 fixed the issue, this just adds a regression test.

iaroslav-reflection:

iaroslav-reflection:
@imeoer approved by @Zephyrcf in July. Good to merge?

Fricounet (commented):
@iaroslav-reflection I'm not against adding the test. But I just wanted to make sure it does catch the regression fixed by #766 but I doesn't really.
Running the test on a pre-#766 commit succeeds just as well. 
Getting the same conditions as a bolt failure in a test isn't gonna be easy (if at all possible) but I just wanted to point it out.

iaroslav-reflection:
Thank you. Added the failed-commit case: the test sets RLIMIT_FSIZE to zero around Remove, so bolt's commit fails on its first page write. It fails on d18dba1, the tree before #766, and passes on main.

Fricounet (approved):
LGTM thanks!

PR #778, opened 2026-07-20
fix(manager): surface liveness subscribe failures instead of reporting success

iaroslav-reflection:
`Subscribe` armed the epoll watch inside a `rawConn.Control` callback that assigned failures to the named error return — which the enclosing `err = rawConn.Control(...)` then overwrote with Control's own nil result. A failed `SetNonblock`/`EpollCtl` still logged "Subscribe daemon ... liveness event" and returned nil.

```mermaid
flowchart LR
    A["err = rawConn.Control(closure)"] --> B["closure: EpollCtl fails → sets err"]
    B --> C["Control returns nil →<br/>err overwritten"]
    C --> D["caller believes daemon is monitored"]
    D --> E["💥 nydusd dies → no death event →<br/>no restart/failover → ENOTCONN,<br/>logs claim monitoring was active"]
```

The daemon was never added to the epoll interest list, so its death produces no event and restart/failover recovery never starts — with nothing in the logs explaining why. The dialed unix connection also leaked on this path.

Add a `controlFD` helper returning the first error from either Control or the callback, make `Subscribe` all-or-nothing (close the connection and register nothing on failure), and use the helper in `unsubscribe`, where an `EPOLL_CTL_DEL` failure previously left a stale fd entry in the interest set that could collide with a future subscription reusing the same fd number.

### DIFF
--- a/pkg/manager/monitor.go
+++ b/pkg/manager/monitor.go
@@ -115,14 +115,15 @@ func (m *livenessMonitor) Subscribe(id string, path string, notifier chan<- deat
 	}
 
 	if rawConn, err = uc.SyscallConn(); err != nil {
-		return
+		if closeErr := uc.Close(); closeErr != nil {
+			log.L.WithError(closeErr).Warnf("close liveness connection for daemon %s", id)
+		}
+		return err
 	}
 
-	err = rawConn.Control(func(fd FD) {
-		err = unix.SetNonblock(int(fd), true)
-		if err != nil {
-			log.L.Errorf("Failed to set file. daemon id %s path %s. %v", id, path, err)
-			return
+	if err = controlFD(rawConn, func(fd FD) error {
+		if err := unix.SetNonblock(int(fd), true); err != nil {
+			return errors.Wrapf(err, "set connection non-blocking, daemon id %s path %s", id, path)
 		}
 
 		event := unix.EpollEvent{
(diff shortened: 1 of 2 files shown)


iaroslav-reflection:
@imeoer @Zephyrcf PTAL

pkg/manager/monitor.go
```diff
@@ -115,39 +115,60 @@ func (m *livenessMonitor) Subscribe(id string, path string, notifier chan<- deat
 	}
 
 	if rawConn, err = uc.SyscallConn(); err != nil {
+		if closeErr := uc.Close(); closeErr != nil {
+			log.L.WithError(closeErr).Warnf("close liveness connection for daemon %s", id)
+		}
 		return
```
Zephyrcf:
```suggestion
		return err
```

Zephyrcf (approved):
LGTM, thanks!

pkg/manager/monitor.go
iaroslav-reflection: Thanks, applied!

iaroslav-reflection:
@imeoer approved by @Zephyrcf, good to merge? The e2e failure is the race flake fixed in #781.

iaroslav-reflection:

iaroslav-reflection:
@imeoer approved by @Zephyrcf. Good to merge?

Fricounet (approved):
LGTM

PR #779, opened 2026-07-20
fix(snapshot): make blob cache GC fail-safe for instances with unknown usage

### LINKED ISSUE #762: [Bug] WaitUntilReady can fail pod creation on transient blobcache metrics NoCounter
### Problem Description

`WaitUntilReady()` queries nydusd blobcache metrics to fill `rafs.UnderlyingFiles`.

In some cases, nydusd returns `BlobcacheMetrics(Metrics(Stats(NoCounter)))` for a very short time. Retrying the same operation later succeeds.

This makes pod creation fail even though the actual RAFS mount may already be usable. 

### Expected Behavior

A transient NoCounter from blobcache metrics should not immediately fail pod creation.

Suggestion:
* retry GetCacheMetrics() briefly when nydusd returns Stats(NoCounter);
* make cache cleanup safe when UnderlyingFiles is missing, usie snapshot metadata and `.blob.meta` files as fallback

### Actual Behavior

In `WaitUntilReady()`:
```go
cacheMetrics, err := d.GetCacheMetrics(sid)
if err != nil {
    return errors.Wrapf(err, "failed to get cache metric")
}

rafs.UnderlyingFiles = cacheMetrics.UnderlyingFiles
```
So a temporary metrics failure makes the snapshot not ready, even if the actual mount may already be usable.

### How to reproduce

.

### Environment Details

.

### Additional Information

This is different from a metrics key mismatch issue #712 .

Even when snapshotter uses the correct metrics key, nydusd may still temporarily return NoCounter before the counter is available.

We have a patch for this internally. I'd like to submit a PR.

iaroslav-reflection:
The cache GC computes its used-blob set solely from `UnderlyingFiles`, which is populated only when containerd calls `Mounts()` (`WaitUntilReady`). Every instance between `fs.Mount()` and its first `Mounts()` call — a window that always exists during pod startup — contributes nothing to the used set, so `getUnusedCacheBlobs` classifies the blobs its freshly started nydusd is already serving as unused, and `RemoveCache` unlinks blob data, chunk maps and blob.meta files under the live daemon. Unlinked-but-open files also keep consuming disk while being invisible to the GC's own accounting.

```mermaid
flowchart LR
    A["fs.Mount: nydusd started<br/>UnderlyingFiles = ∅"] --> B["GC fires in the window"]
    B --> C["instance contributes nothing<br/>to the used set"]
    C --> D["🗑️ blob, chunk_map, blob.meta<br/>unlinked under a live daemon"]
    D --> E["💥 cache thrash · EIO ·<br/>unlinked-but-open files hide<br/>disk usage from GC itself"]
```

Treat unknown as in-use: extract the used-set computation into `collectUsedCacheBlobIDs`, which reports whether the set is complete, and skip the GC round entirely when any blobcache-backed instance has no usage information yet. The gap self-heals once `Mounts()` populates the instance, so pausing one round is safe; deleting on incomplete knowledge is not. Before taking the write transaction, the GC re-queries cache metrics for instances with no usage info yet, checking the daemon socket exists first since `Daemon.GetClient` can poll for it for up to 10s. The round is only skipped if usage is still unknown after that.

Consequently `WaitUntilReady` no longer fails the mount when the one-shot cache metrics fetch errors: the daemon is already verified RUNNING, the metrics only feed GC accounting, and GC now degrades safely when the information is missing.

Fixes #762

### DIFF
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -308,7 +308,13 @@ func (fs *Filesystem) WaitUntilReady(snapshotID string, labels map[string]string
 
 		cacheMetrics, err := d.GetCacheMetrics(sid)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get cache metric")
+			// The daemon is already verified RUNNING above; failing the mount
+			// over a metrics hiccup fails pod creation for no reason (#762).
+			// The underlying-files list only feeds blob cache GC accounting,
+			// and GC treats instances with unknown usage as fully in use, so
+			// continuing without it is safe.
+			log.L.WithError(err).Warnf("failed to get cache metrics for rafs instance %s", rafs.SnapshotID)
+			return nil
 		}
 		log.L.Debugf("Found %d underlying files for rafs instance %s", len(cacheMetrics.UnderlyingFiles), rafs.SnapshotID)
 
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -287,42 +287,16 @@ func (fs *Filesystem) WaitUntilReady(snapshotID string, labels map[string]string
 			return err
 		}
 
-		// For shared daemons, we need to use the correct cache ID to query metrics.
-		// For fscache, the cache is registered with fscacheID (a digest), not the raw snapshot ID.
-		// For fusedev, the cache is registered with the snapshot ID.
-		sid := ""
-		if d.IsSharedDaemon() {
-			if rafs.GetFsDriver() == config.FsDriverFscache {
-				// For fscache, use the fscache ID from annotations if available
-				if fscacheID, ok := rafs.Annotations[racache.AnnoFsCacheID]; ok && fscacheID != "" {
-					sid = fscacheID
-				} else {
-					// Fallback: compute fscacheID if not in annotations yet
-					sid = erofs.FscacheID(rafs.SnapshotID)
-				}
-			} else {
-				// For fusedev, use the snapshot ID directly
-				sid = rafs.SnapshotID
-			}
-		}
-
-		cacheMetrics, err := d.GetCacheMetrics(sid)
-		if err != nil {
+		if err := fs.RefreshUnderlyingFiles(rafs); err != nil {
 			// The daemon is already verified RUNNING above; failing the mount
 			// over a metrics hiccup fails pod creation for no reason (#762).
 			// The underlying-files list only feeds blob cache GC accounting,
-			// and GC treats instances with unknown usage as fully in use, so
-			// continuing without it is safe.
+			// and the GC reconciles instances with unknown usage before every
+			// round, treating them as fully in use until then, so continuing
+			// without it is safe.
 			log.L.WithError(err).Warnf("failed to get cache metrics for rafs instance %s", rafs.SnapshotID)
 			return nil
 		}
-		log.L.Debugf("Found %d underlying files for rafs instance %s", len(cacheMetrics.UnderlyingFiles), rafs.SnapshotID)
-
-		// Lock the daemon's RafsCache when modifying rafs fields to prevent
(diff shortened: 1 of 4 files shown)


iaroslav-reflection:
@imeoer @Zephyrcf PTAL

iaroslav-reflection:
@Fricounet this touches the cache GC you added in #711 — PTAL

iaroslav-reflection:
The failing `run-e2e-for-cgroups` jobs are pre-existing data races, fix proposed in #781.

snapshot/snapshot.go
```diff
@@ -1381,27 +1411,28 @@ func (o *snapshotter) getUnusedCacheBlobs(ctx context.Context) ([]string, error)
...
 	}); err != nil {
 		return nil, errors.Wrap(err, "walk managers to collect used blobs")
 	}
 
+	usedBlobIDs, complete := collectUsedCacheBlobIDs(instances)
+	if !complete {
+		// Unknown usage must count as "in use": deleting blobs based on
+		// incomplete knowledge unlinks cache files under a live daemon. The
+		// gap self-heals once WaitUntilReady populates the instance on the
+		// next Mounts() call, so skip this round instead of guessing.
```
Fricounet: what about if there's an issue and we can't get the cache metrics during the next Mount call as well?
Because if that happens, we won't ever populate the .UnderlyingFiles list and `collectUsedCacheBlobIDs` will always return `complete=false`

pkg/filesystem/fs.go
```diff
@@ -308,7 +308,13 @@ func (fs *Filesystem) WaitUntilReady(snapshotID string, labels map[string]string
 
 		cacheMetrics, err := d.GetCacheMetrics(sid)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get cache metric")
+			// The daemon is already verified RUNNING above; failing the mount
+			// over a metrics hiccup fails pod creation for no reason (#762).
```
Fricounet, line 292: well, I'd argue that failing the pod creation because we can't be sure that the image GC can proceed correctly is a valid reason 😅 

Especially with something like kubernetes where a failed pod creation isn't a big deal since it's just going to be retried a few seconds later anyway. Isn't the case in your situation? Can you share some logs?
Also, I'm curious as to how often you are encountering this?

Fricounet (commented):
> The gap self-heals once Mounts() populates the instance

As this PR is currently, I don't think this statement is always correct, I think we should have a way to reconcile the underlyingFiles map out of band. And the `Cleanup()` function would be the ideal place no?

> Re-querying the daemon at GC time was rejected because Daemon.GetClient can block for up to 10s waiting on the API socket while getUnusedCacheBlobs holds a metastore write transaction.

what about doing the re-query outside (before) the write transaction? This way we attempt to reconcile the rafs who still show `len(underlyingFiles) == 0` each time, and we can abort the cleanup if we still have one case. WDYT?

snapshot/snapshot.go
iaroslav-reflection: With the re-query this only happens while a daemon is actually unreachable. And it can't get stuck forever: removing the snapshot drops the instance from the caches, which unblocks GC.

pkg/filesystem/fs.go
iaroslav-reflection: I kept this part because with the re-query in place, GC doesn't depend on this fetch anymore. So failing the mount here doesn't protect anything, it just costs a kubelet retry.

No logs to share, I found this reading the GC code rather than from an incident. The pod creation failure is what #762 reports.

iaroslav-reflection:
@Fricounet good idea, done. I put the re-query before `TransactionContext` and added a socket existence check first, otherwise `GetClient` can wait up to 10s for a dead daemon on every GC round.

Fricounet (approved):
LGTM thanks!

PR #780, opened 2026-07-20
fix(manager): quarantine damaged daemon records instead of aborting recovery

iaroslav-reflection:
A persisted fusedev daemon record whose `config.json` cannot be reloaded made `recoverDaemons` return an error, aborting the whole `WalkDaemons` pass and failing `NewSnapshotter`. Since daemon records are never pruned, one damaged record crash-loops the snapshotter on every restart until the boltdb is repaired by hand.

Such records are easy to acquire: `createDaemon` persists the record before the configuration is dumped to disk, so a crash or a full disk (the likeliest cause, since the blob cache shares the volume) in that window leaves a record without a config file; a power loss can truncate an existing config the same way.

```mermaid
flowchart LR
    A["record persisted,<br/>config.json missing"] --> B["restart → recovery walk"]
    B --> C["config reload fails →<br/>whole walk aborts"]
    C --> D["NewSnapshotter fails →<br/>snapshotter exits"]
    D --> B
    C --> E["💥 crash loop until manual<br/>boltdb surgery"]
```

Make recovery fault-isolated per record: skip records whose configuration cannot be reloaded — log and continue recovering everything else. The record stays in the store, so restoring the config file lets a later restart re-adopt the daemon.

Persisting the record only after its configuration is on disk is a worthwhile follow-up hardening; this change makes recovery robust to such damage regardless of how it was produced.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -311,6 +311,12 @@ func (m *Manager) cleanUpDaemonResources(d *daemon.Daemon) {
 
 func (m *Manager) recoverDaemons(ctx context.Context,
 	recoveringDaemons *map[string]*daemon.Daemon, liveDaemons *map[string]*daemon.Daemon) error {
+	// Records whose on-disk artifacts are damaged — e.g. a daemon that was
+	// persisted but whose config file never made it to disk. One damaged
+	// record must not abort the whole walk: returning an error here fails
+	// NewSnapshotter, and since the record is never pruned the snapshotter
+	// crash-loops on every restart until the database is repaired by hand.
+	var quarantined []*daemon.Daemon
 	if err := m.store.WalkDaemons(ctx, func(s *daemon.ConfigState) error {
 		if s.FsDriver != m.FsDriver {
 			return nil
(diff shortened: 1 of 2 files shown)


iaroslav-reflection:
@imeoer @Zephyrcf PTAL

iaroslav-reflection:
The failing `run-e2e-for-cgroups` jobs are pre-existing data races, fix proposed in #781.

pkg/manager/manager.go
```diff
@@ -334,8 +340,10 @@ func (m *Manager) recoverDaemons(ctx context.Context,
 		if d.States.FsDriver == config.FsDriverFusedev {
 			cfg, err := daemonconfig.NewDaemonConfig(d.States.FsDriver, d.ConfigFile(""))
 			if err != nil {
-				log.L.Errorf("Failed to reload daemon configuration %s, %s", d.ConfigFile(""), err)
-				return err
+				log.L.Errorf("Quarantining daemon %s: failed to reload configuration %s, %s", d.ID(), d.ConfigFile(""), err)
+				quarantined = append(quarantined, d)
+				//nolint:nilerr
+				return nil
```
Zephyrcf:
```suggestion
				//nolint:nilerr
				return nil
```
I think it's best to skip this part; the existence of these bad records won't actually cause any problems. Another scenario is accidentally deleting these configuration file directories, which means that the existing daemon will be deleted during the subsequent recovery operation. In reality, I only need to restore the corresponding configuration files. Therefore, skipping this part instead of deleting is a conservative and effective approach.

iaroslav-reflection: Agreed, changed to skip without deleting. The daemon is still dropped from the in-memory cache since it's registered before the config load.

Zephyrcf (approved):
LGTM, thanks!

iaroslav-reflection:
@imeoer approved by @Zephyrcf, good to merge?

iaroslav-reflection:

iaroslav-reflection:
@imeoer approved by @Zephyrcf. Good to merge?

Fricounet (approved):
LGTM, thanks!

PR #781, opened 2026-07-20
fix: two data races between background goroutines and their readers

iaroslav-reflection:
Both races are regularly flagged by the `run-e2e-for-cgroups-*` race-enabled jobs and intermittently fail CI for unrelated pull requests (e.g. [this run](https://github.com/containerd/nydus-snapshotter/actions/runs/29733098208/job/88321970728)).

**1. Metrics histogram** — `MetricHistogram.constHists` is rebuilt by the collection loop (`Clear`/`Save`) while the Prometheus registry concurrently ranges over it in `Collect` from HTTP scrape goroutines, with no synchronization. A scrape can observe a torn slice header or a half-rebuilt histogram set. Guarded with a mutex.

**2. RAFS UnderlyingFiles** — `WaitUntilReady` wrote `rafs.UnderlyingFiles` while holding only the daemon's `RafsCache` mutex, but the same `*Rafs` is also handed out by `RafsGlobalCache`, whose `List()` deep-copies every field of the instance under the *global* cache's own mutex (e.g. the snapshot cleanup path). Two different mutexes exclude nothing: the deep copy races with the write, and the torn result feeds cleanup decisions. The update now goes through `Rafs.UpdateUnderlyingFiles`, which takes both cache locks (global first; nothing nests the two locks in the opposite order).

```mermaid
flowchart LR
    W["writer: WaitUntilReady<br/>🔒 d.RafsCache only"] -- writes --> S["shared *Rafs.UnderlyingFiles"]
    R["reader: cleanup via<br/>RafsGlobalCache.List() deepcopy<br/>🔒 global mutex only"] -- reads --> S
    S --> X["⚠️ different mutexes →<br/>no exclusion → torn read"]
```

Both fixes come with tests that fail under `-race` against the previous code (3 and 7 detector reports respectively) and pass now.

### DIFF
--- a/pkg/filesystem/fs.go
+++ b/pkg/filesystem/fs.go
@@ -354,11 +354,11 @@ func (fs *Filesystem) RefreshUnderlyingFiles(instance *racache.Rafs) error {
 	}
 	log.L.Debugf("Found %d underlying files for rafs instance %s", len(cacheMetrics.UnderlyingFiles), instance.SnapshotID)
 
-	// Lock the daemon's RafsCache when modifying rafs fields to prevent
-	// race conditions with concurrent reads (e.g., during cache cleanup)
-	d.RafsCache.Lock()
-	instance.UnderlyingFiles = cacheMetrics.UnderlyingFiles
-	d.RafsCache.Unlock()
+	// This instance is handed out by two registries — RafsGlobalCache
+	// (snapshot cleanup) and the daemon's RafsCache (cache GC, metrics) —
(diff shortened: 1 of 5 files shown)


Zephyrcf (approved):
LGTM, thanks!

iaroslav-reflection:
@Fricounet this one fixes the `-race` e2e flakes, and the reconcile path from #779 now goes through `UpdateUnderlyingFiles` here. Approved by @Zephyrcf, good to merge?

Fricounet (approved):
LGTM thanks

PR #784, opened 2026-07-22
fix(manager): start failover daemon in upgrade mode

Park-Jiyeonn:
## Overview
Failover waits for the replacement nydusd to reach INIT before takeover, but the normal startup path does not pass --upgrade.

Use upgrade mode for failover startup while keeping normal startup unchanged, so the replacement remains in INIT until takeover completes.

## Change Type
- [x] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [ ] Performance Improvement
- [ ] Other (please describe)

### DIFF
--- a/pkg/manager/daemon_adaptor_test.go
+++ b/pkg/manager/daemon_adaptor_test.go
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2026. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package manager
+
+import (
+	"os"
+	"path/filepath"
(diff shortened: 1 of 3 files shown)


pkg/manager/daemon_adaptor_test.go
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) 2022. Nydus Developers. All rights reserved.
```
Zephyrcf:
```suggestion
 * Copyright (c) 2026. Nydus Developers. All rights reserved.
```

Zephyrcf (approved):
LGTM

PR #786, opened 2026-07-22
fix(supervisor): bound daemon state synchronization

Park-Jiyeonn:
## Overview
Keep daemon state transfer from blocking mount and umount paths indefinitely. Bound supervisor socket waits and SendFd requests, and skip a state sync when the supervisor is busy.

## Change Type
- [ ] Bug Fix
- [ ] Feature Addition
- [ ] Documentation Update
- [ ] Code Refactoring
- [x] Performance Improvement
- [ ] Other (please describe)

### DIFF
--- a/pkg/daemon/client.go
+++ b/pkg/daemon/client.go
@@ -59,6 +59,9 @@ const (
 	jsonContentType = "application/json"
 )
 
+// Bound best-effort daemon state sync so it does not stall mount/unmount paths for too long.
+var sendFdRequestTimeout = 5 * time.Second
+
 // Nydusd HTTP client to query nydusd runtime status, operate file system instances.
 // Control nydusd workflow like failover and upgrade.
 type NydusdClient interface {
@@ -102,10 +105,9 @@ func (c *nydusdClient) url(path string, query query) (url string) {
 
 // A simple http client request wrapper with capability to take
 // request body and handle or process http response if result is expected.
-func (c *nydusdClient) request(method string, url string,
+func (c *nydusdClient) requestWithContext(ctx context.Context, method string, url string,
 	body io.Reader, respHandler func(resp *http.Response) error) error {
-
-	req, err := http.NewRequest(method, url, body)
+	req, err := http.NewRequestWithContext(ctx, method, url, body)
 	if err != nil {
 		return errors.Wrapf(err, "construct request %s", url)
 	}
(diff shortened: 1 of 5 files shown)


Park-Jiyeonn:
@Zephyrcf PTAL, thx!

imeoer:
`SendStatesTimeout` now takes the semaphore with an unbounded context:

```go
if err := su.sem.Acquire(context.Background(), 1); err != nil {
```

It didn't touch the semaphore at all before, so this is a new blocking point on the failover/upgrade path — `doDaemonFailover`, `DoDaemonUpgrade` and `system.go` can now stall behind an in-flight `FetchDaemonStates`. And with `to == 0` (there's such a call in the existing tests) `Accept()` never returns and the semaphore is held forever. Please use a bounded context here.

The bigger question is the 200ms acquire timeout in `FetchDaemonStates`. `SendStates()` is deferred from `SharedMount`/`SharedUmount`, so under load we quietly skip persisting the mount set and the supervisor's copy goes stale. If failover kicks in right then, the new nydusd takes over an outdated state and loses instances. `TestSendStatesTimeoutBlocksFetchDaemonStates` effectively pins that behaviour down.

If we do accept the trade-off, could we at least count the skips in a metric and force one non-skippable sync before failover actually starts?

Park-Jiyeonn:
Thx, I agree with both points, I'll fix it.

This script I wrote is a temporary operational recovery tool that rebuilds fscache blob registrations from snapshotter-side config files
```bash
#!/usr/bin/env bash
set -euo pipefail

ROOT="${ROOT:-/home/docker/containerd/io.containerd.snapshotter.v1.nydus}"
SYSTEM_SOCK="${SYSTEM_SOCK:-/run/containerd-nydus/system.sock}"

curl -fsS --unix-socket "$SYSTEM_SOCK" http://localhost/api/v1/daemons |
  jq -r '
    .[] |
    .id as $daemon |
    .api_socket as $api |
    .instances | keys[] |
    [$daemon, $api, .] | @tsv
  ' |
  while IFS=$'\t' read -r daemon api snapshot; do
    config="$ROOT/config/$daemon/$snapshot/config.json"
    test -f "$config"

    response=$(curl -sS --unix-socket "$api" \
      -X PUT \
      -H 'Content-Type: application/json' \
      --data-binary "@$config" \
      http://localhost/api/v2/blobs)

    case "$response" in
      "") echo "rebound $snapshot" ;;
      *"already exists"*) echo "exists  $snapshot" ;;
      *) echo "failed  $snapshot: $response" >&2; exit 1 ;;
    esac
  done
```

PR #789, opened 2026-07-27
fix(daemon): verify process identity before signaling the persisted PID

iaroslav-reflection:
`Terminate` sends SIGTERM to the PID persisted in the daemon record without checking what is actually running there. After a snapshotter restart the record can be stale, and on busy hosts PIDs are recycled quickly — the signal can land on an unrelated process (another nydusd, a shim, a container process).

```mermaid
flowchart LR
    A["record: PID 1234"] --> B["nydusd 1234 dies,<br/>snapshotter restarts"]
    B --> C["kernel recycles 1234<br/>for an unrelated process"]
    C --> D["DestroyDaemon → Terminate"]
    D --> E["💥 SIGTERM to the wrong process"]
```

Fix: record the process start time (field 22 of `/proc/<pid>/stat`) next to the PID at spawn. PID + start time identify a process instance — a recycled PID cannot have the same start time. `Terminate` verifies identity before signaling and refuses on mismatch. This is the same guard runc uses against PID reuse when killing containers (`hasInit`/`signalInit`, comparing `init_process_start` from its state file).

Records persisted by older versions carry no start time and keep today's unverified behavior, so existing daemons are still torn down; every record written after this change is verified. The new `ConfigState` field is JSON-serialized, so old records load cleanly.


iaroslav-reflection:
@imeoer @Zephyrcf PTAL

imeoer:
`RecordProcess` doesn't reset `ProcessStartTime` when reading /proc fails:

```go
func (d *Daemon) RecordProcess(pid int) {
	d.States.ProcessID = pid
	startTime, err := tool.GetProcessStartTime(pid)
	if err != nil {
		log.L.WithError(err).Warnf(...)
		return
	}
	d.States.ProcessStartTime = startTime
}
```

`StartDaemon` runs again on the same `*Daemon` from `doDaemonRestart`/`doDaemonFailover`. If one of those reads fails (short-lived process, for example), the stale start time from the previous process sticks around. `isRecordedProcess()` then compares the new pid's start time against that stale value, always returns false, and `Terminate()` becomes a permanent no-op for that daemon — worse than before the fix. Setting `d.States.ProcessStartTime = 0` in the error branch should cover it.

One more thing: `Terminate()` isn't the only place using the raw pid. `CgroupMgr.AddProc(d.States.ProcessID)` in `StartDaemon` and `recoverDaemons` has real side effects if the pid was recycled, so it's worth guarding there too.

iaroslav-reflection:
Thanks! Both fixed. `RecordProcess` now zeroes the start time when the read fails, so a failed re-record falls back to the grandfather path instead of pinning the stale value — added a regression test for the restart/failover reuse case. For the second point I centralized the check into `VerifiedPid() (pid int, ok bool)` and made `Terminate` and both `AddProc` sites go through it, so future call sites that act on the process have one obvious safe accessor. In `recoverDaemons` a failed check only skips the cgroup add — the daemon still counts as live since its socket answered.

iaroslav-reflection:
@imeoer your points are addressed, PTAL when you have a moment.

PR #790, opened 2026-07-29
fix(auth): only warn when a stored credential fails to renew

iaroslav-reflection:
The renewal loop warns twice per image ref on every reconcile pass — at startup and on every renewal tick — whenever no renewable provider returns credentials. With `enable_cri_keychain` and no renewable providers configured this is the expected outcome (CRI credentials are captured at pull time and cannot be re-fetched), so the warnings read like broken registry auth but signal nothing.

This keeps the initial reconciliation synchronous (unchanged) and fixes the logging instead:

- the image-pull path still warns on a miss (a real problem);
- renewal warns only when a previously stored credential fails to renew (about to go stale);
- a renewal miss for a ref never in the store logs at debug.

Replaces the original defer approach per review discussion.

### DIFF
--- a/pkg/auth/keychain.go
+++ b/pkg/auth/keychain.go
@@ -110,13 +110,21 @@ func getRegistryKeyChainFromProviders(ref string, labels map[string]string, prov
 		// If not available, request credentials valid until the next renewal tick.
 		authReq.ValidUntil = time.Now().Add(renewalStore.renewInterval)
 	}
-	return fetchFromProviders(authReq, providers)
+	kc, err := fetchFromProviders(authReq, providers)
+	if err != nil {
+		// On the pull path a miss is a real problem: the image is being
+		// fetched right now and no provider could authenticate it.
+		logger.WithError(err).Warn("Could not get registry credentials.")
+	}
+	return kc
 }
 
 // fetchFromProviders walks providers in order and returns credentials from the
 // first one that succeeds. If the winning provider is renewable and the renewal
-// store is active, the credentials are cached for periodic renewal.
-func fetchFromProviders(req *AuthRequest, providers []AuthProvider) *PassKeyChain {
+// store is active, the credentials are cached for periodic renewal. When no
+// provider returns credentials, the joined provider errors are returned so the
+// caller can decide how loudly to report the miss.
+func fetchFromProviders(req *AuthRequest, providers []AuthProvider) (*PassKeyChain, error) {
 	logger := log.L.WithField("ref", req.Ref)
 
 	var errs []error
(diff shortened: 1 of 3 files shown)


iaroslav-reflection:
@Fricounet PTAL — this is the renewal path from #718

Fricounet:
Can you share some logs of this happening please? I would like to see how long do we really lose.

If I understand correctly, you are talking about a case where no renewable provider is registered and you only have the CRI provider right? In such a case I think we should just not start the renew loop because without any renewable provider configured there's just no chance of anything coming out of it.

Because in cases where we have renewable providers configured, I think we'd still want to make sure all registered images have their credentials available before starting. Otherwise we risk having containerd calling the snapshoter before it gets creds back, resulting in a nydus daemon created with empty creds, which would be very bad.

> and logs a no secret listener initialized warning wall
if that's a bother we can probably tune the log level or the logic to not do that

iaroslav-reflection:
Yes, CRI keychain only in my case.

Measured on a test node (fusedev dedicated mode, 8 rafs daemons, snapshotter restart). The initial reconcile walks all 8 refs in ~1ms, two warnings per ref:

```
14:23:53.7715 warning "Could not get registry credentials." error="get credentials from *auth.DockerProvider: auth config not complete for host: <registry>\nget credentials from *auth.KubeSecretProvider: no secret listener initialized" ref=<image 1>
14:23:53.7716 warning "credential renewal returned no credentials from any provider" ref=<image 1>
... 7 more pairs ...
14:23:53.7731 info "setup image proxy keychain"
```

So the latency claim in the description was wrong — the walk costs nothing, and I'm dropping the defer. The ~2min startup delay on that node was daemon recovery: 15s per dead daemon record burned in `WaitUntilSocketExisted` retrying on a pid that no longer exists. Filed with data in #792.

On the gate: there's currently no signal for "no renewable provider registered" — `renewableProviders()` always includes `DockerProvider` and `KubeSecretProvider`, and every `CanRenew()` returns true unconditionally. A startup-time "is it configured" check would also permanently disable renewal for docker configs mounted after startup.

Alternative: keep the initial reconcile synchronous where it is (agreed on the empty-creds risk), and only warn when a previously stored credential fails to renew; a first-time fetch that finds nothing logs at debug. The store is empty at startup and stays empty with CRI-only creds, so that removes both the startup wall and the recurring per-tick warnings, while real renewal failures still warn. I'd rework this PR to that if it works for you.

iaroslav-reflection:
Reworked: dropped the defer, the initial reconcile stays synchronous. Now this only changes logging — warn when a stored credential fails to renew, debug when the store never had the ref (the CRI-only case). @Fricounet @imeoer PTAL

iaroslav-reflection:
@Fricounet reworked as discussed above, does this work for you?

Fricounet (approved):
Sorry for the delay. Works for me, thank you!

Issue #791, opened 2026-07-29
startup: gRPC socket only appears after full recovery — restarts on nodes with accumulated state go dark for minutes


iaroslav-reflection:
On restart, `NewSnapshotter` completes all recovery work before `Serve()` creates the unix socket: per-daemon `GetState`, restarting dead daemons (including remounting their RAFS instances), the hot-upgrade check, and the initial credential reconciliation (#790 moves that one off the startup path). Recovery time scales with the daemons/mounts/snapshots accumulated on the node, and nothing prunes those, so on a long-lived node a restart can leave the socket missing for minutes — containerd gets `Unavailable` for every snapshot RPC and all new container creation on the node fails until recovery finishes. systemd reports the unit `active` the whole time (`Type=simple`), so nothing signals the window. A crash during recovery re-enters it every `RestartSec`.

Two design questions before proposing anything concrete:

1. Is serve-after-full-recovery an invariant you want to keep (correctness of snapshot ops against half-recovered state), or would you consider serving earlier and recovering daemons in the background?
2. If the ordering must stay: would you take (a) an opt-in readiness signal (`sd_notify` + `Type=notify`) so operators can gate on and observe the recovery window, and/or (b) parallelizing the per-daemon recovery walk so the window is bounded by the slowest daemon instead of the sum?

Happy to send PRs for whichever direction you prefer.

Fricounet:
Personally, I think 1. is important. Having the nydus-snapshotter or any of the nydus daemon enter into a correctness issue is always a nightmare to fix, and I find that the best solution is often to trash the node entirely. So I'd rather not risk more situations leading to this.

For 2. I think both a and b make sense if you want to implement them

Other opinions @imeoer @BraveY ?

iaroslav-reflection:
Measured the restart path on a test node (fusedev dedicated, 8 daemon records, processes dead at startup): socket absent for 120s, a pod created during the window took 116s to start and produced 11 error events (9× FailedCreatePodSandBox). The entire window is `recoverDaemons` — strictly serial, ~15s per dead daemon, and the 15s each is a retry bug in `WaitUntilSocketExisted`, fixed in #792. With that fix the same node serves in 0.5s and the pod starts clean.

So the pathological case is gone; 2(a) sd_notify still seems worth it for observability, and 2(b) matters mainly for many live daemons where GetState/remount time dominates. Will send 2(a) first unless you'd rather I start elsewhere.

BraveY:
[...]

PR #792, opened 2026-07-31
fix(daemon): stop waiting for the API socket when the daemon process is gone

iaroslav-reflection:
`WaitUntilSocketExisted` passes an abort callback to `retry.OnlyRetryIf` (return true = stop retrying). The callback returns `false` when `IsZombieProcess` errors — i.e. when the process is gone entirely — so waiting on a dead daemon runs all 100 attempts (~15s with jitter) instead of aborting on the first. A zombie aborts immediately; a fully gone process doesn't.

The impact is on startup: `recoverDaemons` → `GetState` → `GetClient` → `WaitUntilSocketExisted` runs serially for every recorded daemon before `Serve()`, so N dead daemon records (host reboot, daemons killed while the snapshotter was down) mean N×15s with no gRPC socket and containerd getting `Unavailable` for every snapshot RPC.

Measured on a test node with 8 dead daemon records, main @ 8a7cfda:

| | socket absent | pod created during the window |
|---|---|---|
| before | 120.3s | Running after 116s, 11 error events (9× FailedCreatePodSandBox) |
| after | 0.5s | Running after 5s, no error events |

Before — each daemon burns ~15s:
```
14:21:53.623 Recovering daemon ID d9l1sjgmf3lkti1guqog
14:22:08.736 Daemon d9l1sjgmf3lkti1guqog died somehow. Clean up its vestige!
14:22:08.736 Recovering daemon ID d9l1t90mf3lldd30og8g
14:22:23.859 Daemon d9l1t90mf3lldd30og8g died somehow. Clean up its vestige!
... (8 total)
```

After — all 8 handled within 1ms:
```
14:24:18.7935 Recovering daemon ID d9l1sjgmf3lkti1guqog
14:24:18.7936 Process 1836476 is gone, giving up waiting for socket ...
14:24:18.7936 Daemon d9l1sjgmf3lkti1guqog died somehow. Clean up its vestige!
... (8 total)
```

The regression test fails on main (runs the full 15s) and passes with the fix.

Related: #791 — this removes the pathological case of that startup window.

### DIFF
--- a/pkg/daemon/client.go
+++ b/pkg/daemon/client.go
@@ -191,7 +191,10 @@ func WaitUntilSocketExisted(sock string, pid int) error {
 		retry.OnlyRetryIf(func(error) bool {
 			zombie, err := tool.IsZombieProcess(pid)
 			if err != nil {
-				return false
+				// The process is gone, so its socket will never show up.
+				// Stop retry immediately.
+				log.L.WithError(err).Errorf("Process %d is gone, giving up waiting for socket %s", pid, sock)
+				return true
 			}
 			// Stop retry if nydus daemon process is already in Zombie state.
 			if zombie {
--- a/pkg/daemon/client_test.go
+++ b/pkg/daemon/client_test.go
@@ -12,8 +12,10 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"os/exec"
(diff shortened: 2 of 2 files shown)


imeoer (approved):
LGTM, thanks!

PR #794, opened 2026-08-10
feat(cmd): notify systemd readiness after the snapshotter starts serving

iaroslav-reflection:
Implements 2(a) from #791.

The snapshotter serves its gRPC socket only after recovering all daemon state, but with `Type=simple` systemd marks the unit `active` the moment the process starts — the recovery window is invisible and nothing can order on it.

This sends `READY=1` right before the gRPC server starts serving, and `STOPPING=1` on shutdown. Both are no-ops unless `NOTIFY_SOCKET` is set, so nothing changes for existing deployments; the shipped unit files gain a commented `Type=notify` line to opt in. With `Type=notify`, the unit stays `activating` during recovery, `After=` dependents wait for actual readiness, and a crash loop during recovery shows as a unit that never reaches `active` instead of flapping `active`.

`coreos/go-systemd` was already an indirect dependency; promoted to direct.

### DIFF
--- a/cmd/containerd-nydus-grpc/sdnotify.go
+++ b/cmd/containerd-nydus-grpc/sdnotify.go
@@ -25,10 +25,10 @@ func notifySystemd(state string) {
 	}
 }
 
-// notifyReady tells the service manager the snapshotter has finished recovery
-// and its gRPC socket is serving. With Type=notify, systemd keeps the unit in
-// "activating" until this point, so dependents can order on real readiness
-// instead of process start.
+// notifyReady tells the service manager that the snapshotter has finished
+// recovery and its gRPC socket is ready to accept connections. With
+// Type=notify, systemd keeps the unit "activating" until this point, allowing
+// dependents to wait for actual readiness instead of merely process startup.
 func notifyReady() {
 	notifySystemd(sd.SdNotifyReady)
 }
--- a/cmd/containerd-nydus-grpc/sdnotify.go
+++ b/cmd/containerd-nydus-grpc/sdnotify.go
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2026. Nydus Developers. All rights reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package main
+
+import (
+	"github.com/containerd/log"
+	sd "github.com/coreos/go-systemd/v22/daemon"
+)
+
+// notifySystemd reports a lifecycle state to the service manager (sd_notify).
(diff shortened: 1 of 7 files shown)


cmd/containerd-nydus-grpc/sdnotify.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	if sent {
+		log.L.Infof("notified service manager: %s", state)
+	}
+}
+
+// notifyReady tells the service manager the snapshotter has finished recovery
+// and its gRPC socket is serving. With Type=notify, systemd keeps the unit in
+// "activating" until this point, so dependents can order on real readiness
+// instead of process start.
```
Zephyrcf:
```suggestion
// notifyReady tells the service manager that the snapshotter has finished
// recovery and its gRPC socket is ready to accept connections. With
// Type=notify, systemd keeps the unit "activating" until this point, allowing
// dependents to wait for actual readiness instead of merely process startup.
```

Zephyrcf (commented):
Others LGTM

cmd/containerd-nydus-grpc/sdnotify.go
iaroslav-reflection: Thanks, applied!

iaroslav-reflection:
@Zephyrcf PTAL

iaroslav-reflection:
@Zephyrcf @imeoer PTAL.

misc/snapshotter/nydus-snapshotter.fscache.service
```diff
@@ -5,6 +5,10 @@ Before=containerd.service
 
 [Service]
 Type=simple
+# The snapshotter serves its gRPC socket only after recovering all daemon
+# state. Switch to Type=notify to keep the unit "activating" until then,
+# so dependents order on real readiness instead of process start.
+#Type=notify
```
Fricounet, line 14: could you add a small comment info on `TimeoutStartSec` as well? 
IIRC the default value is 90s so this might not be enough for cases that take multiple minutes, leading to systemd restarting the service, and the service never coming up as a result.

Fricounet (commented):
mostly LGTM, 1 doc nit.
Thanks for the clean implementation!

misc/snapshotter/nydus-snapshotter.fscache.service
iaroslav-reflection: Added to all three unit files.

Fricounet: thanks!

Fricounet (approved):
LGTM

PR #795, opened 2026-08-10
feat(manager): recover daemons concurrently

iaroslav-reflection:
Implements 2(b) from #791 (endorsed there by @imeoer).

`recoverDaemons` did all per-daemon work (supervisor create, config load, `GetState` over the API socket) serially inside the `WalkDaemons` bolt transaction, so the pre-serving window scales with the sum of per-daemon costs. This collects the records during the walk (the transaction no longer wraps the recovery work), then recovers with a bounded fan-out of 8. The per-daemon logic is unchanged; the result maps are mutex-guarded, the daemon cache and supervisor set already take their own locks.

One behavioral difference: an error in one daemon's recovery no longer stops the remaining daemons from being processed — recovery still fails overall with the first error.

Test: 6 live mock daemons responding after 400ms each plus 2 dead records recover in well under the 2.4s serial cost, with the maps populated correctly.

Note: #780 touches the same function; happy to rebase whichever lands second.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -329,11 +329,19 @@ func (m *Manager) cleanUpDaemonResources(d *daemon.Daemon) {
 	log.L.Infof("Deleting resources %v", resource)
 }
 
-// recoverConcurrency bounds how many daemons are recovered at once. Recovery
-// is IO-bound (an HTTP request over each daemon's API socket), so a modest
+// recoverConcurrency bounds how many daemons are probed at once. Probing is
+// IO-bound (an HTTP request over each daemon's API socket), so a modest
 // fan-out bounds the pre-serving window by the slowest daemons instead of the
 // sum, without stampeding hosts that accumulated many daemons.
 const recoverConcurrency = 8
+
+// probedDaemon is the outcome of the concurrent, side-effect-free probe of
+// one persisted daemon record.
+type probedDaemon struct {
+	d     *daemon.Daemon
+	state types.DaemonState
+	dead  bool
+}
 
 func (m *Manager) recoverDaemons(ctx context.Context,
 	recoveringDaemons *map[string]*daemon.Daemon, liveDaemons *map[string]*daemon.Daemon) error {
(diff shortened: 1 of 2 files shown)


pkg/manager/manager.go
```diff
@@ -328,83 +329,108 @@ func (m *Manager) cleanUpDaemonResources(d *daemon.Daemon) {
...
+	var eg errgroup.Group
+	eg.SetLimit(recoverConcurrency)
+
+	for _, s := range states {
+		eg.Go(func() error {
+			opt := make([]daemon.NewDaemonOpt, 0)
+			var d, _ = daemon.NewDaemon(opt...)
+			d.States = *s
+
+			m.daemonCache.Update(d)
```
Zephyrcf: @iaroslav-reflection I’m concerned about partial state being committed when one recovery fails. The cache and supervisor set are updated here before config loading and GetState; later, liveDaemons is updated before AddProc. Any of those later steps can return an error, while the earlier mutations are not rolled back. Since the errgroup allows the other recoveries to continue, eg.Wait() may report startup failure after several goroutines have already partially updated global state. Could we first perform the fallible probing/validation concurrently, then commit cache, supervisor, cgroup, and result-map changes only for successfully recovered daemons?

iaroslav-reflection: Done — split into probe and commit. The goroutines only build the daemon, reload config and GetState, writing into a results slice by index; nothing shared is touched. If any probe fails, recovery returns with nothing committed (added a test for that). Commit runs serially: cache, supervisor, then AddProc before the liveDaemons insert. The result-map mutex is gone too.

iaroslav-reflection:
@Zephyrcf PTAL

iaroslav-reflection:
@Zephyrcf @imeoer PTAL.

Fricounet (approved):
LGTM on my end. Leaving this open a bit longer if @Zephyrcf wants to chime in

Zephyrcf (approved):
LGTM!

imeoer:
Plz resolve the conflicts then we can merge it. :)

iaroslav-reflection:
resolved

imeoer (approved):

Fricounet (approved):

PR #797, opened 2026-09-10
snapshotter: wait for nydus socket before restarting containerd

### LINKED ISSUE #796: containerd CRI fails 'failed to find snapshotter nydus' when containerd is restarted before the nydus socket is ready
### Summary

During a DaemonSet **rolling update** (or any pod re-deploy), `snapshotter.sh deploy` restarts containerd **unconditionally** on every node (`wait_service_active` -> `systemctl restart containerd`). When containerd restarts **before the nydus snapshotter's gRPC socket is ready**, its CRI plugin fails to load and the node goes NotReady (`container runtime is down`).

The same failure reproduces on **both** containerd 1.7 and 2.x once `snapshotter = "nydus"` is actually in effect.

### Error observed (containerd journal)

```
level=warning msg="failed to load plugin"
  error="failed to create CRI service: failed to find snapshotter \"nydus\""
  id=io.containerd.grpc.v1.cri
level=warning msg="failed to load plugin"
  error="unable to load CRI image service plugin dependency: failed to find snapshotter \"nydus\""
  id=io.containerd.cri.v1.images
level=warning msg="failed to load plugin"
  error="unable to load CRI image service plugin dependency: failed to find snapshotter \"nydus\""
  id=io.containerd.grpc.v1.sandbox-controllers
```

containerd core still reports `containerd successfully booted`, but the CRI plugin is disabled, so the kubelet reports `container runtime is down` -> node `Ready=False`.

### Root cause

In `misc/snapshotter/snapshotter.sh`, `deploy_snapshotter()` ends with:

```bash
wait_service_active 30 5 ${CONTAINER_RUNTIME}   # does `systemctl restart containerd`
```

`wait_service_active()` restarts containerd the moment the nydus-snapshotter service is `is-active` — which can be true **before the snapshotter has bound its socket** (`/run/containerd-nydus/containerd-nydus-grpc.sock`). containerd then starts with `[proxy_plugins.nydus]` pointing at a not-yet-ready socket, the proxy plugin fails to register, and the CRI plugin (configured with `snapshotter = "nydus"`) hard-fails on lookup.

- **First install / node provision** usually survives because the socket happens to be bound before the restart.
- **Rolling updates / pod re-deploys** race and fail — and the old pod's `cleanup` also restarts containerd with the snapshotter stopped.
- Imported: the failure is **not containerd-version-specific**. It appears regardless of 1.x vs 2.x once `snapshotter = "nydus"` is effective (2.x only appeared "tolerant" while the config was written to a section it ignored).

### Suggested directions (not implemented)

1. `snapshotter.sh`: wait for the nydus **socket to be ready** (e.g. `while [ ! -S "${SNAPSHOTTER_GRPC_SOCKET}" ]; do sleep 1; done`) before `wait_service_active ${CONTAINER_RUNTIME}` — covering both deploy and cleanup restart sites.
2. Or, don't `restart` containerd when the config is already applied / only restart when it actually changed.
3. Or, make the CRI plugin tolerate a temporarily-missing snapshotter (retry/lazy) instead of hard-failing at init.

### Environment

- nydus-snapshotter v0.15.4 and v0.15.15 (both reproduce)
- containerd 1.7.34 and 2.3.3-2
- Deployment: DaemonSet (helm `nydus`), `snapshotter.sh deploy`
- ~10 nodes per rolling rollout **lost** (kubernetes `NodeStatusUnknown` / `KubeletNotReady`, container runtime down) before the rollout was stopped.

Happy to provide more logs or run any additional diagnostics.

logesh-motorq:
## Summary

Fixes #796. During nydus DaemonSet rolling updates / pod re-deploys, `deploy_snapshotter()`
restarts containerd the moment the nydus-snapshotter systemd service reports `is-active` —
which can be **before the nydus gRPC socket is bound**. containerd then starts with
`[proxy_plugins.nydus]` pointing at a not-yet-ready socket, the proxy plugin fails to register,
and the CRI plugin hard-fails with `failed to find snapshotter "nydus"`, taking the node down
(`container runtime is down` → NotReady).

Reproduces on containerd **1.7 and 2.x** once `snapshotter = "nydus"` is in effect. First install
usually survives (socket happens to be bound in time); rolling updates / re-deploys race and kill
the node (~10 nodes lost in a single representative rollout before it was stopped).

View of the failing containerd journal:

```
level=warning msg="failed to load plugin"
  error="failed to create CRI service: failed to find snapshotter \"nydus\""
  id=io.containerd.grpc.v1.cri
```

## Change

Add a bounded `wait_for_nydus_socket()` (socket exists, up to a 30s timeout) in
`deploy_snapshotter()` before restarting containerd, so the CRI plugin can resolve nydus at init.
Not added to the cleanup/`remove` path because that reverts the config to a non-nydus baseline
before restarting, so nydus isn't required there.

## Test plan
- [ ] Rolling update of a nydus DaemonSet no longer drops nodes (validated against containerd 2.3.3)
- [ ] `snapshotter.sh` still passes shellcheck/syntax

### DIFF
--- a/misc/snapshotter/snapshotter.sh
+++ b/misc/snapshotter/snapshotter.sh
@@ -183,9 +183,14 @@ function deploy_snapshotter() {
         sed -i "s|^ExecStart=.*$|ExecStart=$COMMANDLINE|" "${SNAPSHOTTER_SERVICE}"
         nsenter -t 1 -m systemctl daemon-reload
         nsenter -t 1 -m systemctl enable nydus-snapshotter.service
+        # A socket file left behind by a previous (crashed or cleaned-up)
+        # snapshotter process would satisfy the readiness check below against
+        # a dead listener, so drop it before (re)starting the service.
+        rm -f "${SNAPSHOTTER_GRPC_SOCKET}"
         wait_service_active 30 5 nydus-snapshotter
     else
         echo "running snapshotter as standalone process"
+        rm -f "${SNAPSHOTTER_GRPC_SOCKET}"
         ${COMMANDLINE} &
     fi
 
@@ -202,15 +207,19 @@ function deploy_snapshotter() {
 }
 
 # Wait until the nydus snapshotter gRPC socket exists (bound), up to a timeout.
+# Aborts deployment if the socket never appears: restarting containerd without
+# it takes the node down, while aborting keeps the currently running containerd
+# alive and surfaces the failure as a pod error.
 function wait_for_nydus_socket() {
-    local wait_time="${NYDUS_SOCKET_WAIT_TIMEOUT:-30}"
+    local timeout="${NYDUS_SOCKET_WAIT_TIMEOUT:-30}"
+    local wait_time="${timeout}"
     while [ "$wait_time" -gt 0 ] && [ ! -S "${SNAPSHOTTER_GRPC_SOCKET}" ]; do
         echo "waiting for nydus snapshotter socket ${SNAPSHOTTER_GRPC_SOCKET}"
         sleep 1
         wait_time=$((wait_time - 1))
     done
     if [ ! -S "${SNAPSHOTTER_GRPC_SOCKET}" ]; then
-        echo "WARNING: nydus snapshotter socket not ready after timeout"
+        die "nydus snapshotter socket ${SNAPSHOTTER_GRPC_SOCKET} not ready after ${timeout}s"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


Fricounet (commented):
Thanks for the PR! Mostly LGTM, can you also document the flag [there](https://github.com/containerd/nydus-snapshotter/blob/main/docs/setup_snapshotter_by_daemonset.md)? 

Although, this issue will probably be solved by https://github.com/containerd/nydus-snapshotter/pull/794, since systemd will wait before considering nydus ready.

